PR #6, opened 2019-06-13
chore: fix make targets and install operator-sdk by curl

dipak-pawar:
*Changes proposed in this PR*
- removes not-exist make target `/vendor`
- install operator-sdk by curl
- reduce number of caching layer by adding `LABEL` and `ENV` in single dir.
- exporting  `GO111MODULE=on`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,9 @@ include ./make/format.mk
 include ./make/lint.mk
 include ./make/test.mk
 
+GO111MODULE?=on
+export GO111MODULE
+
 .PHONY: build
 ## Build the operator
 build: ./out/operator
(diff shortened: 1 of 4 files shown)


dipak-pawar:
>I'm just not sure about the devtools@redhat.com email address for the Dockerfile maintainer.

sure, we need to have some valid email which belongs to us. we can change it later.

> we could also want to run the go mod vendor command to copy the dependencies in the ./vendor directory, so we can get a direct access to the sources from the IDE (which is something I like to have personally)

I don't want to run `go mod vendor` for every build. maybe we should add some make target which developer has to run when he wanted to see source code in his IDE

PR #20, opened 2019-06-20
enhance useraccont ctrl

nurali-techie:
- Add Watch for `UserAccount` as primary resource
- Add Watch for `User` and `Identity` as secondary resource
- Create/Ensure User, Identity, UserIdentityMapping are created for given UserAccount instance


pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -41,6 +47,18 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+		IsController: true,
+		OwnerType:    &toolchainv1alpha1.UserAccount{},
+	}
+	if err := c.Watch(&source.Kind{Type: &userv1.User{}}, enqueueRequestForOwner); err != nil {
+		return err
+	}
+	if err := c.Watch(&source.Kind{Type: &userv1.Identity{}}, enqueueRequestForOwner); err != nil {
+		return err
+	}
+
```
nurali-techie: adding `UserIdentityMapping` watch NOT possible.  Basically, `UserIdentityMapping` Get() operation is NOT available and adding watch calls it and failed with error - no kind for `UserIdentityMappingList`

MatousJobanek: The `UserIdentityMapping` Get() operation is available: https://docs.openshift.com/container-platform/3.11/rest_api/apis-user.openshift.io/v1.UserIdentityMapping.html but it's not possible to list/watch this kind of resources. But this is a detail - the result is the same: it's not possible to watch them. This means that it needs to be created at the same time when a User is created and then, in the next reconcile, check that the specific mapping exists.

alexeykazakov: Yeah..
`UserIdentityMapping` is a virtual resource. It doesn't exist in etcd.
The `create/delete` rest endpoint for this resource just updates the corresponding User and Identity: https://github.com/openshift/openshift-apiserver/blob/27f8efac5366bf8ec380644e326a2212fc0ff6ea///pkg/user/apiserver/registry/useridentitymapping/rest.go#L83
And `get` generates it's on fly out of the corresponding User and Identity if they have references to each other - https://github.com/openshift/openshift-apiserver/blob/27f8efac5366bf8ec380644e326a2212fc0ff6ea/////pkg/user/apiserver/registry/useridentitymapping/rest.go#L313:6

So, there is nothing to watch. We should watch User and Identity instead. And create them. Also in every reconcile if both User and Identity exist we should try to get UserIdentiyMapping (the api server will just check if User/Identity resources refer to each other). If it doesn't "exist" then "create" it. Please note that we should "create" it only if both User and Identity exist.
That "create" will update the corresponding User/Identity which will trigger another reconcile. So finally we will get an event when everything is ready.

nurali-techie: @MatousJobanek @alexeykazakov thanks for details.

.. but neither `Watch()` nor `Get()` possible for `UserIdentityMapping` (though there is GET REST API for the same)

Below Get call will give error:
**Code:** `r.client.Get(context.TODO(), types.NamespacedName{Name: name}, mapping)`
**Error:** no kind "UserIdentityMappingList" is registered for version "user.openshift.io/v1" in scheme "pkg/runtime/scheme.go:101"

Overall, I have already done what @alexeykazakov mentioned.  Plz check `encureMapping()` [link](https://github.com/codeready-toolchain/member-operator/blob/406473b5aed387786d343093ddbe3b27911d6224/pkg/controller/useraccount/useraccount_controller.go#L158).

Now, I am not getting what is missing and what should I change ??

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -64,18 +82,136 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
 			return reconcile.Result{}, nil
 		}
-		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	var created bool
+	var user *userv1.User
+	if user, created, err = r.ensureUser(userAcc); err != nil || created == true {
+		return reconcile.Result{}, err
```
alexeykazakov: Let's update the UserAccount status to `provisioning` if `ensureUser()` returns `created == true`.

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -64,18 +82,136 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+	}
+
+	var identity *userv1.Identity
+	if identity, created, err = r.ensureIdentity(userAcc); err != nil || created == true {
+		return reconcile.Result{}, err
+	}
+
+	if created, err = r.ensureMapping(userAcc, user, identity); err != nil || created == true {
 		return reconcile.Result{}, err
 	}
```
alexeykazakov: And update the status to `provisioned` here.

pkg/controller/useraccount/useraccount_controller.go
alexeykazakov: It looks good now. Just need to update the status.

pkg/controller/useraccount/useraccount_controller.go
nurali-techie: I guess, status=_provisioning_ will be done by host_cluster on MasterUserRecord which later sync to member_cluster with UserAccount.status=_provisioning_
Ref: [user provisioning seq flow](https://docs.google.com/document/d/19UzaXVqiOo9JqDYz3qUKguJwW-7zceA4UtTr358R_CA/edit#heading=h.bv9n6w3laifk)

pkg/controller/useraccount/useraccount_controller.go
nurali-techie: I guess, we can do that once the namespace provisioned for this user as part of https://jira.coreos.com/browse/CRT-74 .. fine ??

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -64,18 +82,136 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+	userAcc := &toolchainv1alpha1.UserAccount{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: config.GetOperatorNamespace(), Name: request.Name}, userAcc)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
 			return reconcile.Result{}, nil
 		}
-		// Error reading the object - requeue the request.
```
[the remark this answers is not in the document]
nurali-techie: I guess not necessary,  the same is covered by method level comment.

nurali-techie:
@alexeykazakov thanks for review.  I will make sure the relevant task related to ci stuff will be created in jira once synced with @xcoulon.  If possible, please approve this PR if there are no more feedback.

nurali-techie:
[test]

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+func TestEnsureUser(t *testing.T) {
+	logf.SetLogger(logf.ZapLogger(true))
+	username := "johnsmith"
+	userID := "34113f6a-2e0d-11e9-be9a-525400fb443d"
+	s := scheme.Scheme
+	apis.AddToScheme(s)
```
golangcibot: Error return value of `apis.AddToScheme` is not checked (from `errcheck`)

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -0,0 +1,143 @@
...
+		assert.Equal(t, userAcc.Name, newUser.Name)
+	})
+}
+
+func TestEnsureIdentity(t *testing.T) {
+	logf.SetLogger(logf.ZapLogger(true))
+	username := "johnsmith"
+	userID := "34113f6a-2e0d-11e9-be9a-525400fb443d"
+	s := scheme.Scheme
+	apis.AddToScheme(s)
```
golangcibot: Error return value of `apis.AddToScheme` is not checked (from `errcheck`)

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -64,18 +82,136 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+		return reconcile.Result{}, err
+	}
+
+	var created bool
+	var user *userv1.User
+	if user, created, err = r.ensureUser(userAcc); err != nil || created == true {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `created` (from `gosimple`)

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -64,18 +82,136 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+		return reconcile.Result{}, err
+	}
+
+	var identity *userv1.Identity
+	if identity, created, err = r.ensureIdentity(userAcc); err != nil || created == true {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `created` (from `gosimple`)

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -64,18 +82,136 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+		return reconcile.Result{}, err
+	}
+
+	if created, err = r.ensureMapping(userAcc, user, identity); err != nil || created == true {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `created` (from `gosimple`)

pkg/controller/useraccount/useraccount_controller.go
alexeykazakov: Host does not change status of MasterUserRecord directly.

T1. MasterUserRecord `state` is set to `approved`
T2. Host controller set its status to `provisioning` and creates a UserAccount in the member
T3. UserAccount controllers provisions the user and namespaces and updates the UserAccount status accordingly.
T4. When all done, the member controller updates the `resourceVersion` in the corresponding MasterUserAccount in host.
T5. Host controller checks the corresponding UserAccounts in member and updates the MasterUserRecord status.

So, each controller updates status of its primary resources only.

pkg/controller/useraccount/useraccount_controller.go
alexeykazakov: Well, it's still a completed provisioning. We just don't do the namespace provisioning yet in this PR. And it's a few lines of code.

pkg/controller/useraccount/useraccount_controller.go
MatousJobanek: > Let's update the UserAccount status to provisioning if ensureUser() returns created == true.

I think that this is too late - firstly, it should check if the User exists, if not, then it should set `status = provisioning` and in the next reconcile it should create all resources. Otherwise, it would create a `User` resource and then update the status which would trigger two reconciles

MatousJobanek: I think that T3 step is more complex - we should break it down to:

T3-a: the UserAccount controller sets status = provisioning
T3-b: the member controller updates the resourceVersion in the corresponding MasterUserAccount in host.
T3-c: Host controller checks the corresponding UserAccounts in member and updates the MasterUserRecord status.
T3-d (in parallel with T3-c): UserAccount controllers provisions the user and namespaces and updates the UserAccount status accordingly.

nurali-techie: > Otherwise, it would create a User resource and then update the status which would trigger two reconciles

@MatousJobanek yes, I have come across this issue and used to get two reconcile callback.  Thanks for solution.

Overall, I see that coding to get `reconcile()` callback in proper manner is becoming complex and not obvious.  Each time you change some stuff and you need to see if you are getting proper `reconcile()` callback or not.  I really would like to check if other_operator has follow some different approach here.

alexeykazakov: Hm.. I was assuming that status updates are ignored but they are not. Status updates do trigger reconciling.
But it looks like `GenerationChangedPredicate` from operator-sdk does the trick: https://github.com/operator-framework/operator-sdk/blob/121664e0093774c6d4fa4528fe7367d0848e6d74/pkg/predicate/predicate.go

I tested it and it looks like it filters out the events which caused by status updates only:
```
err = c.Watch(&source.Kind{Type: &toolchainv1alpha1.MasterUserRecord{}}, &handler.EnqueueRequestForObject{}, predicate.GenerationChangedPredicate{})
...
```
In this case we can update the status without warring about triggering a reconcile.
Take a look at Knative serving controller for example: https://github.com/knative/serving-operator/blob/master/pkg/controller/knativeserving/knativeserving_controller.go#L142-L153

nurali-techie: @alexeykazakov this helped and working :+1:  Thank you :-)

nurali-techie: @MatousJobanek not sure if we want to do this as part of this PR for now https://github.com/codeready-toolchain/member-operator/pull/20#discussion_r297146111 @alexeykazakov  ??

pkg/controller/useraccount/useraccount_controller.go
nurali-techie: done [d79d00f](https://github.com/codeready-toolchain/member-operator/pull/20/commits/d79d00f6cf3d7eed5890446ea9c32e75f7563fc6)

pkg/controller/useraccount/useraccount_controller.go
alexeykazakov: That’s fine to do the rest in separate PRs. Smaller PRs easier to review :)

nurali-techie:
> Do we have e2e tests jira created?

@alexeykazakov https://jira.coreos.com/browse/CRT-160

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -64,18 +84,158 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
 		}
-		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
 
-	return reconcile.Result{}, nil
+	var created bool
+	var user *userv1.User
+	if user, created, err = r.ensureUser(userAcc); err != nil {
+		r.updateStatus(userAcc, toolchainv1alpha1.StatusProvisioning, err.Error())
```
golangcibot: Error return value of `r.updateStatus` is not checked (from `errcheck`)

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -64,18 +84,158 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+		return reconcile.Result{}, err
+	} else if created {
+		return reconcile.Result{}, r.updateStatus(userAcc, toolchainv1alpha1.StatusProvisioning, "")
+	}
+
+	var identity *userv1.Identity
+	if identity, created, err = r.ensureIdentity(userAcc); err != nil {
+		r.updateStatus(userAcc, toolchainv1alpha1.StatusProvisioning, err.Error())
```
golangcibot: Error return value of `r.updateStatus` is not checked (from `errcheck`)

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -64,18 +84,158 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+		return reconcile.Result{}, err
+	} else if created {
+		return reconcile.Result{}, r.updateStatus(userAcc, toolchainv1alpha1.StatusProvisioning, "")
+	}
+
+	if created, err = r.ensureMapping(userAcc, user, identity); err != nil {
+		r.updateStatus(userAcc, toolchainv1alpha1.StatusProvisioning, err.Error())
```
golangcibot: Error return value of `r.updateStatus` is not checked (from `errcheck`)

pkg/controller/useraccount/useraccount_controller.go
nurali-techie: @MatousJobanek I guess there is already sub-task for https://github.com/codeready-toolchain/member-operator/pull/20#discussion_r297146111 .. task https://jira.coreos.com/browse/CRT-138

MatousJobanek: Yes I agree, synchronization should not be part of this PR

MatousJobanek: > ... But it looks like GenerationChangedPredicate from operator-sdk  ....

yeah, after reading this blog post https://blog.openshift.com/kubernetes-custom-resources-grow-up-in-v1-10/ I finally understand the purpose of the subresources and the usage.

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -64,18 +84,158 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
 		}
-		// Error reading the object - requeue the request.
 		return reconcile.Result{}, err
 	}
 
-	return reconcile.Result{}, nil
+	var created bool
+	var user *userv1.User
+	if user, created, err = r.ensureUser(userAcc); err != nil {
+		_ = r.updateStatus(userAcc, toolchainv1alpha1.StatusProvisioning, err.Error())
```
MatousJobanek: if `r.ensureUser` returns an error, does it really mean that the status should be set to `provisioning`?  Shouldn't that be either `failed` or `unknown`?

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -64,18 +84,158 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+		return reconcile.Result{}, err
+	} else if created {
+		return reconcile.Result{}, r.updateStatus(userAcc, toolchainv1alpha1.StatusProvisioning, "")
+	}
+
+	var identity *userv1.Identity
+	if identity, created, err = r.ensureIdentity(userAcc); err != nil {
+		_ = r.updateStatus(userAcc, toolchainv1alpha1.StatusProvisioning, err.Error())
```
MatousJobanek: same with the status here and also in case of mapping

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -0,0 +1,184 @@
...
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+	"github.com/codeready-toolchain/member-operator/pkg/apis"
+	"github.com/stretchr/testify/assert"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+func TestEnsureUser(t *testing.T) {
```
MatousJobanek: I would be really thankful if we could follow and use the `//given //when //then` parts approach - it really helps with understanding the test

pkg/controller/useraccount/useraccount_controller.go
MatousJobanek: btw, either we return the error and handle it correctly or, do not return the error, but definitely not ignore it - in this case, you could log it

pkg/controller/useraccount/useraccount_controller.go
nurali-techie: my understanding is, first step in provisioning is create_user. so we started provisioning and tried to create_user which failed it means that there is error in provisioning due to create_user.

pkg/controller/useraccount/useraccount_controller.go
nurali-techie: Inside updateStatus() err is logged if not nil.

The idea is, if caller already has one_error (like error in create_identity) then he will use that error and fail `reconcile()`.  If there are no_error and if updateStatus() failed then use that error and fail `reconcile()`.

pkg/controller/useraccount/useraccount_controller.go
MatousJobanek: it can also fail on getting the User resource (even if it exists)... My point was that when there is an error, then you cannot say for sure that the user is/not provisioned. The status should reflect that... On the other hand, the next reconcile should start pretty quickly, so the next round could fix that.

pkg/controller/useraccount/useraccount_controller.go
MatousJobanek: > Inside updateStatus() err is logged if not nil.

that's the point, I think that the error should be logged outside of this method. 
Anyway, every golinter would scream on ignoring returned errors and I would call it also as a good practice - to handle returned errors...

MatousJobanek: btw, as for the logging (I didn't noticed it during the review) - we should use the logger with the additional values https://github.com/codeready-toolchain/member-operator/blob/master/pkg/controller/useraccount/useraccount_controller.go#L83 and not the "global variable". The reason is that the values give us a context which resource the error/info is related to

nurali-techie: > that's the point, I think that the error should be logged outside of this method.

@MatousJobanek anyway @alexeykazakov has created follow-up PR for this.  Here is my thought on this.

I have thought all possible way before making this change.  I understand that log statement should be outside but for that I need to check "err != nil" for each call made to `updateStatus()` and as updateStatus is called many time (now and usage will increase in future code) I see that the code will be having lot of this `err != nil` check all over and it will **compromise the readability of code**.  So even though linter will scream I was fine as gaining readability of code.

Also, this method is really internal/private to this controller so I was fine adding log.err inside method.  If this is public then obviously it should return the result and left on caller to log/not_log it.

pkg/controller/useraccount/useraccount_controller.go
nurali-techie: @MatousJobanek I am not getting you point.

What I can say is, when you start `reconcile()` and if userAcc is found then one can/should set status="provisioning".  Later if some error happened then the userAcc.Status.Error should be updated with it and above code does that.  I am not elaborating the reasoning much as we already have another PR for follow-up changes.

PR #26, opened 2019-06-27
e2e test

nurali-techie:
- add e2e test for UserAccount ctrl

PR #29, opened 2019-07-01
chore: add 'make test-e2e' 

xcoulon:
This PR adds initial make target to run end to end tests.

This make target runs e2e tests in `minishift` and `openshift-ci` environment.

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -42,9 +42,24 @@ rules:
 - apiGroups:
   - core.kubefed.k8s.io
   resources:
-  - KubeFedCluster
+  - kubefedclusters
   verbs:
   - "*"
+- apiGroups:
+  - core.kubefed.k8s.io
+  resources:
+  - kubefedclusters/status
+  verbs:
+  - update
+- apiGroups:
+  - user.openshift.io
+  resources:
+  - users
+  - identities
+  verbs:
+  - list
+  - watch
+  - create
 - apiGroups:
   - ""
   resources:
@@ -58,3 +73,9 @@ rules:
   - useraccounts
   verbs:
   - '*'
+- apiGroups:
+  - toolchain.dev.openshift.com
+  resources:
+  - useraccounts/status
+  verbs:
+  - update
--- a/make/test.mk
+++ b/make/test.mk
@@ -46,15 +46,15 @@ upload-codecov-report:
 ############################################################
 
 .PHONY: test-e2e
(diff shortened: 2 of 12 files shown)


make/test.mk
```diff
@@ -26,3 +38,41 @@ test-with-coverage:
...
 	# Upload coverage to codecov.io
 	bash <(curl -s https://codecov.io/bash) -f $(COV_DIR)/coverage.txt -t 51cc45ad-2e54-4e68-98cb-8ab15cf2b8df
+
+############################################################
+#
+# End-to-end Tests
+#
+############################################################
+
+.PHONY: test-e2e
```
xcoulon: Ideally this goal will be run with an extra job on OpenShift CI. I'll see if it's doable

make/test.mk
```diff
@@ -26,3 +38,41 @@ test-with-coverage:
...
+
+e2e-setup: ./vendor get-test-namespace e2e-cleanup docker-image
+	$(Q)oc new-project $(TEST_NAMESPACE) --display-name e2e-tests
+	$(Q)-oc apply -f ./deploy/service_account.yaml 
+	$(Q)-oc apply -f ./deploy/role.yaml 
+	$(Q)sed -e "s,REPLACE_NAMESPACE,$(TEST_NAMESPACE)," ./deploy/role_binding.yaml | oc apply -f -
+	$(foreach crd_file,$(wildcard deploy/crds/*.yaml), \
+		oc apply -f $(crd_file); \
+	)
+	$(eval OPERATOR_IMAGE_NAME := $(shell minishift openshift registry)/${TEST_NAMESPACE}/${GO_PACKAGE_REPO_NAME}:${GIT_COMMIT_ID_SHORT})
```
alexeykazakov: minishift? It's not going to work on openshift-ci, is it? Or I'm missing something?

xcoulon: ah good point! I'll keep that for the local tests, and I'll see if we can have something similar with the test cluster or if we need to push the image to quay.io

deploy/role.yaml
```diff
@@ -42,9 +42,24 @@ rules:
...
+  - update
+- apiGroups:
+  - user.openshift.io
+  resources:
+  - users
+  - identities
+  verbs:
+  - list
+  - watch
+  - create
```
alexeykazakov: We also need get and update. But Nurali is adding it in https://github.com/codeready-toolchain/member-operator/pull/26/

PR #30, opened 2019-07-03
Switch UserAccount to Status Conditions

alexeykazakov:
This PR introduces Status Condition to UserAccount Controller: **Ready**

Also there are new [additional printer columns](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#additional-printer-columns) in UserAccount CRD:
```
$ oc get useraccounts
NAME        NS LIMIT   TIER NAME   READY     REASON 
bob         default    basic       True      Provisioned   
alice       default    basic       False     Provisioning   
johnsmith   admin      basic       False     UnableToCreateIdentity   

$ oc get useraccounts -o wide
NAME        USER ID               NS LIMIT   TIER NAME   READY     REASON                    DISABLED
bob         d8a88d7-1d27c0d7e52   default    basic       True      Provisioned   
alice       f758563-2d27c0d7e53   default    basic       False     Provisioning   
johnsmith   1a03eca-2dd7fb21c40   admin      basic       False     UnableToCreateIdentity   
```
Where in the example above ^ **bob** successfully provisioned. **Alice** is still being provisioned and **JohnSmith** provision failed.

### DIFF
--- a/deploy/crds/toolchain_v1alpha1_useraccount.yaml
+++ b/deploy/crds/toolchain_v1alpha1_useraccount.yaml
@@ -30,6 +30,10 @@ spec:
           type: object
         spec:
           properties:
+            disabled:
+              description: If set to true then the corresponding user should not be
+                able to login "false" is assumed by default
+              type: boolean
             nsLimit:
               description: The namespace limit name
               type: string
@@ -73,12 +77,32 @@ spec:
           type: object
         status:
           properties:
-            error:
-              description: The error message in case of failed status
-              type: string
-            status:
-              description: 'Observed status. For example: provisioning|provisioned'
-              type: string
+            conditions:
+              description: 'Conditions is an array of current User Account conditions
+                Supported condition types: Provisioning, UserNotReady, IdentityNotReady,
+                UserIdentityMappingNotReady, NSTemplateSetNotReady and Ready +patchMergeKey=type
+                +patchStrategy=merge'
+              items:
+                properties:
+                  lastTransitionTime:
+                    description: Last time the condition transit from one status to
+                      another.
+                    format: date-time
+                    type: string
+                  message:
+                    description: Human readable message indicating details about last
+                      transition.
+                    type: string
+                  reason:
+                    description: (brief) reason for the condition's last transition.
+                    type: string
+                  status:
+                    description: Status of the condition, one of True, False, Unknown.
+                    type: string
+                required:
+                - status
+                type: object
+              type: array
           type: object
   version: v1alpha1
 status:
--- a/deploy/crds/toolchain_v1alpha1_useraccount.yaml
+++ b/deploy/crds/toolchain_v1alpha1_useraccount.yaml
@@ -15,6 +15,7 @@ spec:
     - name: User ID
       type: string
       JSONPath: .spec.userID
+      priority: 1
     - name: NS Limit
       type: string
       JSONPath: .spec.nsLimit
@@ -30,6 +31,7 @@ spec:
     - name: Disabled
       type: boolean
       JSONPath: .spec.disabled
+      priority: 1
   subresources:
     status: {}
   validation:
--- a/deploy/crds/toolchain_v1alpha1_useraccount.yaml
+++ b/deploy/crds/toolchain_v1alpha1_useraccount.yaml
@@ -21,15 +21,15 @@ spec:
     - name: Tier Name
       type: string
       JSONPath: .spec.nsTemplateSet.tierName
-    - name: Disabled
-      type: boolean
-      JSONPath: .spec.disabled
     - name: Ready
       type: string
       JSONPath: ".status.conditions[?(@.type==\"Ready\")].status"
     - name: Reason
       type: string
       JSONPath: ".status.conditions[?(@.type==\"Ready\")].reason"
+    - name: Disabled
+      type: boolean
+      JSONPath: .spec.disabled
   subresources:
     status: {}
   validation:
--- a/deploy/crds/toolchain_v1alpha1_useraccount.yaml
+++ b/deploy/crds/toolchain_v1alpha1_useraccount.yaml
@@ -11,6 +11,25 @@ spec:
     kind: UserAccount
     plural: useraccounts
   scope: Namespaced
+  additionalPrinterColumns:
+    - name: User ID
+      type: string
+      JSONPath: .spec.userID
+    - name: NS Limit
+      type: string
+      JSONPath: .spec.nsLimit
+    - name: Tier Name
+      type: string
+      JSONPath: .spec.nsTemplateSet.tierName
+    - name: Disabled
+      type: boolean
+      JSONPath: .spec.disabled
+    - name: Ready
+      type: string
+      JSONPath: ".status.conditions[?(@.type==\"Ready\")].status"
+    - name: Reason
+      type: string
+      JSONPath: ".status.conditions[?(@.type==\"Ready\")].reason"
   subresources:
     status: {}
   validation:
--- a/deploy/crds/toolchain_v1alpha1_useraccount.yaml
+++ b/deploy/crds/toolchain_v1alpha1_useraccount.yaml
@@ -99,7 +99,11 @@ spec:
                   status:
                     description: Status of the condition, one of True, False, Unknown.
                     type: string
+                  type:
+                    description: Type of condition
+                    type: string
                 required:
+                - type
                 - status
                 type: object
               type: array
--- a/pkg/controller/useraccount/useraccount_controller.go
+++ b/pkg/controller/useraccount/useraccount_controller.go
@@ -135,7 +135,7 @@ func (r *ReconcileUserAccount) ensureUser(logger logr.Logger, userAcc *toolchain
 			if err := r.client.Create(context.TODO(), user); err != nil {
 				return nil, false, r.wrapErrorWithStatusUpdate(logger, userAcc, r.setStatusUserCreationFailed, err, "failed to create user '%s'", userAcc.Name)
 			}
-			if err := r.setStatusUserCreationOK(userAcc); err != nil {
+			if err := r.setStatusProvisioning(userAcc); err != nil {
 				return nil, false, err
 			}
 			logger.Info("user created successfully", "name", userAcc.Name)
@@ -155,7 +155,7 @@ func (r *ReconcileUserAccount) ensureUser(logger logr.Logger, userAcc *toolchain
 		if err := r.client.Update(context.TODO(), user); err != nil {
 			return nil, false, r.wrapErrorWithStatusUpdate(logger, userAcc, r.setStatusMappingCreationFailed, err, "failed to update user '%s'", userAcc.Name)
 		}
-		if err := r.setStatusMappingCreationOK(userAcc); err != nil {
+		if err := r.setStatusProvisioning(userAcc); err != nil {
 			return nil, false, err
 		}
 		logger.Info("user updated successfully", "name", userAcc.Name)
@@ -180,7 +180,7 @@ func (r *ReconcileUserAccount) ensureIdentity(logger logr.Logger, userAcc *toolc
 			if err := r.client.Create(context.TODO(), identity); err != nil {
 				return nil, false, r.wrapErrorWithStatusUpdate(logger, userAcc, r.setStatusIdentityCreationFailed, err, "failed to create identity '%s'", name)
 			}
-			if err := r.setStatusIdentityCreationOK(userAcc); err != nil {
+			if err := r.setStatusProvisioning(userAcc); err != nil {
 				return nil, false, err
 			}
 			logger.Info("identity created successfully", "name", name)
@@ -203,7 +203,7 @@ func (r *ReconcileUserAccount) ensureIdentity(logger logr.Logger, userAcc *toolc
 		if err := r.client.Update(context.TODO(), identity); err != nil {
 			return nil, false, r.wrapErrorWithStatusUpdate(logger, userAcc, r.setStatusMappingCreationFailed, err, "failed to update identity '%s'", name)
 		}
-		if err := r.setStatusMappingCreationOK(userAcc); err != nil {
+		if err := r.setStatusProvisioning(userAcc); err != nil {
 			return nil, false, err
 		}
 		logger.Info("identity updated successfully", "name", name)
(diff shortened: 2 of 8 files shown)


pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -107,40 +115,46 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
 			user = newUser(userAcc)
 			if err := controllerutil.SetControllerReference(userAcc, user, r.scheme); err != nil {
-				return nil, false, r.wrapErrorWithStatusUpdate(logger, userAcc, err, "failed to set controller reference for user '%s'", userAcc.Name)
+				return nil, false, r.wrapErrorWithStatusUpdate(logger, userAcc, r.setStatusUserCreationFailed, err, "failed to set controller reference for user '%s'", userAcc.Name)
 			}
 			if err := r.client.Create(context.TODO(), user); err != nil {
-				return nil, false, r.wrapErrorWithStatusUpdate(logger, userAcc, err, "failed to create user '%s'", userAcc.Name)
+				return nil, false, r.wrapErrorWithStatusUpdate(logger, userAcc, r.setStatusUserCreationFailed, err, "failed to create user '%s'", userAcc.Name)
+			}
+			if err := r.setStatusUserCreationOK(userAcc); err != nil {
```
MatousJobanek: I'm not really a fan of updating the status every time when either User, Identity or IdentityMapping is created - remember that every update causes an update of the MUR and status synchronization, so at least 5 times per one user provisioning which we don't really need (two are more than enough :-)). 
I think that we could modify the method so it sets the condition status only if the condition already exists - to cover the case when it would fail in a previous reconcile. WDYT?

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -191,29 +211,148 @@ func (r *ReconcileUserAccount) ensureIdentity(logger logr.Logger, userAcc *toolc
...
+			Message: message,
+		},
+		toolchainv1alpha1.Condition{
+			Type:   toolchainv1alpha1.UserAccountReady,
+			Status: apiv1.ConditionFalse,
+			Reason: reason,
+		})
+}
+
+func (r *ReconcileUserAccount) setStatusUserCreationOK(userAcc *toolchainv1alpha1.UserAccount) error {
```
MatousJobanek: I believe that this method and the following two can be merged into one:
```go 
func (r *ReconcileUserAccount) setStatusOK(userAcc *toolchainv1alpha1.UserAccount, conType toolchainv1alpha1.ConditionType) error {
	return r.updateStatusConditions(
		userAcc,
		toolchainv1alpha1.Condition{
			Type:   conType,
			Status: apiv1.ConditionFalse,
		})
}
```

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -191,29 +211,148 @@ func (r *ReconcileUserAccount) ensureIdentity(logger logr.Logger, userAcc *toolc
 	return identity, false, nil
 }
 
-// updateStatus updates user account status to given status with errMsg but only if the current status doesn't match
-// If the current status already set to desired state then this method does nothing
-func (r *ReconcileUserAccount) updateStatus(userAcc *toolchainv1alpha1.UserAccount, status toolchainv1alpha1.StatusUserAccount, errMsg string) error {
-	if userAcc.Status.Status == status && userAcc.Status.Error == errMsg {
-		// Nothing changed
+// wrapErrorWithStatusUpdate wraps the error and update the user account status. If the update failed then log the error.
+func (r *ReconcileUserAccount) wrapErrorWithStatusUpdate(logger logr.Logger, userAcc *toolchainv1alpha1.UserAccount, statusUpdater func(userAcc *toolchainv1alpha1.UserAccount, message string) error, err error, format string, args ...interface{}) error {
```
MatousJobanek: it looks like the number of parameters is growing - we could think about creating a new struct that would keep apart from the client also the logger and UserAccount so we don't have to pass it to all methods every time (it doesn't have to be part of this PR of course)

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -245,3 +384,48 @@ func newIdentity(userAcc *toolchainv1alpha1.UserAccount, user *userv1.User) *use
...
+func AddOrUpdateStatusCondition(currentConditions []toolchainv1alpha1.Condition, newCondition toolchainv1alpha1.Condition) ([]toolchainv1alpha1.Condition, bool) {
+	var oldCondition *toolchainv1alpha1.Condition
+	var index int
+	var newConditions []toolchainv1alpha1.Condition
+	if currentConditions == nil {
+		newConditions = make([]toolchainv1alpha1.Condition, 0, 1)
+	} else {
+		newConditions = make([]toolchainv1alpha1.Condition, 0, len(currentConditions))
+		for i, condition := range currentConditions {
+			newConditions[i] = currentConditions[i]
```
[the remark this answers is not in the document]
alexeykazakov: Yeah, I had something similar to that at first. But then decided that it’s probably better not to modify the current conditions param/slice and always create a new one. Just to avoid potential situations when such modification can be unexpected/undesired.
But maybe it’s not that critical and having a simpler/clearer function implementation is worth it...

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -245,3 +384,48 @@ func newIdentity(userAcc *toolchainv1alpha1.UserAccount, user *userv1.User) *use
...
+	if oldCondition.Status == newCondition.Status &&
+		oldCondition.Reason == newCondition.Reason &&
+		oldCondition.Message == newCondition.Message {
+		// Nothing changed. No need to update.
+		return newConditions, false
+	}
+
+	// Update LastTransitionTime if the status changed
+	if newCondition.Status == oldCondition.Status {
+		newCondition.LastTransitionTime = oldCondition.LastTransitionTime
```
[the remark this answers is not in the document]
alexeykazakov: Correct. If I understand LastTransitionTime correctly, we are supposed to change it only if its status changed (true/false/unknown).

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -191,29 +211,148 @@ func (r *ReconcileUserAccount) ensureIdentity(logger logr.Logger, userAcc *toolc
...
-	if err == nil {
-		return nil
+// updateStatusConditions updates user account status conditions with the new conditions
+func (r *ReconcileUserAccount) updateStatusConditions(userAcc *toolchainv1alpha1.UserAccount, newConditions ...toolchainv1alpha1.Condition) error {
+	var atLeastOneUpdated bool
+	conditions := userAcc.Status.Conditions
+	var updated bool
+	for _, condition := range newConditions {
+		conditions, updated = AddOrUpdateStatusCondition(conditions, condition)
+		atLeastOneUpdated = atLeastOneUpdated || updated
```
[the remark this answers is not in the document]
alexeykazakov: Yeah. Good idea.

pkg/controller/useraccount/useraccount_controller.go
alexeykazakov: Well, it’s a question about optimization vs. more detailed status updates. I would prefer the later for now because of:
- user provisioning is not that frequent. So I don’t think that optimizations will give us much.
- in the beginning I think more details might be more useful than optimization. Later on we can revisit it.

pkg/controller/useraccount/useraccount_controller.go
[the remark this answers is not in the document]
MatousJobanek: I see. Well, it's still possible to always return a new array - just instead of setting to the specific index:
```go
currentConditions[i] = newCondition
return currentConditions, true
```
you can do something like this:
```go
return append(append(currentConditions[:i], newCondition), currentConditions[i+1:]...), true
```

pkg/controller/useraccount/useraccount_controller.go
MatousJobanek: Ok

pkg/controller/useraccount/useraccount_controller.go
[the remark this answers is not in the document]
alexeykazakov: good point

pkg/controller/useraccount/useraccount_controller.go
alexeykazakov: While the current implementation is more verbose (I agree) the usage IMO is a bit clearer:

```
setStatusUserCreationOK(userAcc)
```
vs.
```
setStatusOK(userAcc, toolchainv1alpha1.UserAccountUserNotReady)
```

pkg/controller/useraccount/useraccount_controller.go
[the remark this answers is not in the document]
alexeykazakov: Done in https://github.com/codeready-toolchain/toolchain-common/pull/16

pkg/controller/useraccount/useraccount_controller.go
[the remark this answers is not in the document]
alexeykazakov: Done in https://github.com/codeready-toolchain/toolchain-common/pull/16

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -206,82 +275,101 @@ func prepareReconcile(t *testing.T, username string, initObjs ...runtime.Object)
...
+			},
+			{
+				Type:   toolchainv1alpha1.UserAccountReady,
+				Status: corev1.ConditionTrue,
+			},
+		}
 
-		err := reconciler.updateStatus(userAcc, toolchainv1alpha1.StatusProvisioning, "")
+		// when
+		err := reconciler.updateStatusConditions(userAcc, conditions...)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -206,82 +275,101 @@ func prepareReconcile(t *testing.T, username string, initObjs ...runtime.Object)
...
 		}
+		conditions := []toolchainv1alpha1.Condition{{
+			Type:   toolchainv1alpha1.UserAccountProvisioning,
+			Status: corev1.ConditionFalse,
+		}}
+		userAcc.Status.Conditions = conditions
 
-		err := reconciler.updateStatus(userAcc, toolchainv1alpha1.StatusProvisioning, "some error")
+		// when
+		err := reconciler.updateStatusConditions(userAcc, conditions...)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/controller/useraccount/useraccount_controller.go
alexeykazakov: On the other hand maybe just having Provisioning and Ready conditions would be enough... All the error messages with reasons can be in Ready condition.

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -118,6 +118,7 @@ func TestReconcileOK(t *testing.T) {
 				toolchainv1alpha1.Condition{
 					Type:   toolchainv1alpha1.UserAccountReady,
 					Status: corev1.ConditionFalse,
+					Reason: "Provisioning",
```
nurali-techie: I guess, we can use const introduced at https://github.com/codeready-toolchain/member-operator/pull/30/commits/90db32a057df8635ec0f412c5aaeccb843a6bb74#diff-25293d17133b9d89712a15c8e56795a3R35

nurali-techie: done https://github.com/codeready-toolchain/member-operator/pull/30/commits/4d9628b13cc8cff29b8e5f1928135fa306069c28

nurali-techie:
@alexeykazakov one confusion, why we need both `type` and `reason` ??
To me, we can manage to convey all the info with one field.

Below is current types:
```
	UserAccountProvisioning ConditionType = "Provisioning"
	UserAccountUserNotReady ConditionType = "UserNotReady"
	UserAccountIdentityNotReady ConditionType = "IdentityNotReady"
	UserAccountUserIdentityMappingNotReady ConditionType = "UserIdentityMappingNotReady"
	UserAccountNSTemplateSetNotReady ConditionType = "NSTemplateSetNotReady"
	UserAccountReady ConditionType = "Ready"
```

Below is current reasons:
```
	unableToCreateUserReason     = "UnableToCreateUser"
	unableToCreateIdentityReason = "UnableToCreateIdentity"
	unableToCreateMappingReason  = "UnableToCreateMapping"
	provisioningReason           = "Provisioning"
	provisionedReason            = "Provisioned"
```

While I am still understanding how k8s ideally handle all these stuff .. I just want to check with you if can just replace `UserAccountReady` with `UserAccountProvisioned` (note **ed**) and can work with type only ??

alexeykazakov:
> why we need both type and reason

Reason describes why we have this condition. There can be multiple reasons.

>  can just replace UserAccountReady with UserAccountProvisioned (note ed) and can work with type only ??

If we have only Provisioned then it could be ambiguous. Why it's not provisioned? Maybe it's because it's still provisioning. Or maybe because it failed. And failed can mean both "failed but still provisioning" or "failed and not trying any more".

But I agree there is some overlapping and I also wondering if we are over-complicating the conditions...

As alternative I was thinking about reducing the condition types to:
- *UserAccountProvisioning*
- *UserAccountReady*

If anything failed then UserAccountProvisioning would be true (if we still trying) and UserAccountReady would be false with reason == UserCreationFailed/IdentityCreationFailed/...
But it such case our conditions would be that accurate/detailed as they are now: If there was a problem with User creation the status Ready would keep the reason "UserCreationFailed" even when the problem is fixed but before we check all the secondary resources. Only after all successful checks we would set the status to Ready and remove the failed reason.
Not sure if it's worth it.

alexeykazakov:
> As alternative I was thinking about reducing the condition types to:
>- UserAccountProvisioning
>- UserAccountReady

@nurali-techie this is what I'm talking about: https://github.com/alexeykazakov/member-operator/commit/70ca4d34da258971a275e3dae899295172dfabba

But again... not sure if such simplification is worth it.

deploy/crds/toolchain_v1alpha1_useraccount.yaml
```diff
@@ -11,6 +11,27 @@ spec:
     kind: UserAccount
     plural: useraccounts
   scope: Namespaced
+  additionalPrinterColumns:
```
xcoulon: This file is generated after https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/useraccount_types.go, and I'm afraid these new additions to this yaml file might be lost next time we make an update in the API.

alexeykazakov: Hm.. right @xcoulon. I missed that important aspect. We will have to do something in our generator (or at least document it for now somewhere) to avoid that.

But these additions are important. We have to keep them in CRD.

PR #32, opened 2019-07-03
Fix race condition in UserAccount controller

alexeykazakov:
We currently "crate" `IdentityMapping`. Which is a virtual resource. The mapping rest endpoint modifies the corresponding User & Identity when the mapping is "created". It causes two reconciles in UserAccount controller which happen in parallel:

- **T1.** UserAcc created.
- **T2**. Reconcile 1 triggered. Create User. Exit Reconcile.
- **T3.** Reconcile 2. Create Identity. Exit.
- **T4.** Reconcile 3. "Create" Mapping (it modifies User and Identity, so it triggers two reconciles!). Exit reconcile.
- **T5-T6.** Race condition. Two reconciles (one because Identity modified and one because User modified)

This PR drops any `UserIdentityMapping` REST API usage. We now handle user & identity updates (for establishing a mapping between them) explicitly from the controller.

### DIFF
--- a/pkg/controller/useraccount/useraccount_controller.go
+++ b/pkg/controller/useraccount/useraccount_controller.go
@@ -5,8 +5,8 @@ import (
 	"fmt"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
-
 	"github.com/codeready-toolchain/member-operator/pkg/config"
+
 	"github.com/go-logr/logr"
 	userv1 "github.com/openshift/api/user/v1"
 	"github.com/operator-framework/operator-sdk/pkg/predicate"
@@ -97,18 +97,13 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	var created bool
+	var createdOrUpdated bool
 	var user *userv1.User
-	if user, created, err = r.ensureUser(reqLogger, userAcc); err != nil || created {
+	if user, createdOrUpdated, err = r.ensureUser(reqLogger, userAcc); err != nil || createdOrUpdated {
 		return reconcile.Result{}, err
 	}
 
-	var identity *userv1.Identity
-	if identity, created, err = r.ensureIdentity(reqLogger, userAcc); err != nil || created {
-		return reconcile.Result{}, err
-	}
-
-	if created, err = r.ensureMapping(reqLogger, userAcc, user, identity); err != nil || created {
+	if _, createdOrUpdated, err = r.ensureIdentity(reqLogger, userAcc, user); err != nil || createdOrUpdated {
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -212,38 +221,25 @@ func newUser(userAcc *toolchainv1alpha1.UserAccount) *userv1.User {
 		ObjectMeta: metav1.ObjectMeta{
 			Name: userAcc.Name,
 		},
+		Identities: []string{getIdentityName(userAcc)},
```
nurali-techie: @alexeykazakov FYI, you setting `Identities` in `newUser()` is enough to established mapping from user to identity and the logic here [link](https://github.com/codeready-toolchain/member-operator/blob/6be0efb12395e5d45ac450e8fbc0eaa55e012570/pkg/controller/useraccount/useraccount_controller.go#L136-L147) is never get executed.

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -212,38 +221,25 @@ func newUser(userAcc *toolchainv1alpha1.UserAccount) *userv1.User {
...
-func newMapping(user *userv1.User, identity *userv1.Identity) *userv1.UserIdentityMapping {
-	name := identity.Name
-	mapping := &userv1.UserIdentityMapping{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: name,
-		},
 		User: corev1.ObjectReference{
 			Name: user.Name,
 			UID:  user.UID,
 		},
```
nurali-techie: @alexeykazakov same here, setting `User` in `newIdentity()` is enough to established mapping from Identity to User.  The logic here [link](https://github.com/codeready-toolchain/member-operator/blob/6be0efb12395e5d45ac450e8fbc0eaa55e012570/pkg/controller/useraccount/useraccount_controller.go#L175-L189) never get executed.

nurali-techie:
@alexeykazakov the only thing I would like us to note that we are setting reference to `Identities` in `ensureUser()` before actual `Identity` object is created.  It's like creating foreign key before the reference object exists.  I am not sure if that will create any issue now or in future at all.

In ideal case, the sequence would be ..
- create_user
- create_identity with user reference
- update_user with identity reference

pkg/controller/useraccount/useraccount_controller.go
alexeykazakov: It does get executed. This logic is for updating the existing User in case it's missing the mapping.
Example:
1. Create UserAccount and let the controller create User, Identity with mapping.
2. Update the created User and remove the reference to the Identity.
3. The controller will restore the mapping.

You can remove this code and see that the test for updating the user is now failing.

pkg/controller/useraccount/useraccount_controller.go
alexeykazakov: Same as above. It's needed for updates.

alexeykazakov:
> the only thing I would like us to note that we are setting reference to Identities in ensureUser() before actual Identity object is created. It's like creating foreign key before the reference object exists. I am not sure if that will create any issue now or in future at all.

Yes, I have the same concern. But it looks like OpenShift doesn't really care and you can set references to not yet existing Identity. I would keep it as is because it really simplifies our code. We can always change it if we see some issues. For now, e2e tests for that should proof we are good.

PR #33, opened 2019-07-12
WIP Implement Garbage Collection for UserAccount

tinakurian:
This is still a WIP. It currently does not work :-)


pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -39,13 +46,25 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 }
 
 func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	pred := &GenerationChangedPredicate{
+		predicate.GenerationChangedPredicate{
```
golangcibot: composites: `github.com/codeready-toolchain/member-operator/vendor/github.com/operator-framework/operator-sdk/pkg/predicate.GenerationChangedPredicate` composite literal uses unkeyed fields (from `govet`)

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -107,6 +108,64 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
 	}
 
+	// Finalizer name
+	userAccFinalizerName := "finalizer.toolchain.dev.openshift.com"
+
+	// Add the finalizer if it is not present
+	if !containsString(userAcc.ObjectMeta.Finalizers, userAccFinalizerName) {
+		userAcc.ObjectMeta.Finalizers = append(userAcc.ObjectMeta.Finalizers, userAccFinalizerName)
+		if err := r.client.Update(context.TODO(), userAcc); err != nil {
+			return reconcile.Result{Requeue: true}, err
```
nurali-techie: Minor: No need to set `Requeue: true`.
This should work `return reconcile.Result{}, err`.  Why?? bcoz, we did `Update` to userAcc so `Reconcile()` will be called back for the same.

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -107,6 +108,64 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+		}
+
+		// Remove finalizer from UserAccount
+		userAcc.ObjectMeta.Finalizers = removeString(userAcc.ObjectMeta.Finalizers, userAccFinalizerName)
+		if err := r.client.Update(context.Background(), userAcc); err != nil {
+			return reconcile.Result{Requeue: true}, err
+		}
+
+		return reconcile.Result{}, nil
```
nurali-techie: Design change - good to have: Can we move this new_logic out of `Reconcile()` to diff method.  Currently, `Reconcile()` mostly call other method so it can reflect the overall flow rather internal logic.

nurali-techie:
@tinakurian can you plz explain in desc on why userAcc not getting delete and what we are planning to implement to make it happen :+1:

pkg/controller/useraccount/useraccount_controller.go
tinakurian: Thanks Nurali - addressed as part of d8498c421eaf5dd07ecf330e6cdd4063ce57e21d

pkg/controller/useraccount/useraccount_controller.go
tinakurian: Yes! Makes sense - addressed as part of d8498c421eaf5dd07ecf330e6cdd4063ce57e21d

PR #35, opened 2019-07-17
add make target for deploy host and create kubefedcluster for e2e

dipak-pawar:
*Changes Proposed in this PR*
* Add make target to deploy host-operator
* Add make target to create kubefedcluster cr
* install `jq` tool for k8s yaml parsing

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -46,12 +46,16 @@ upload-codecov-report:
 ###########################################################
 
 .PHONY: test-e2e
-test-e2e:  e2e-setup
-	sed -e 's|REPLACE_IMAGE|${IMAGE_NAME}|g' ./deploy/operator.yaml  | oc apply -f -
+test-e2e:  deploy-host e2e-setup setup-kubefed
 	# This is hack to fix https://github.com/operator-framework/operator-sdk/issues/1657
 	echo "info: Running go mod vendor"
 	go mod vendor
 	operator-sdk test local ./test/e2e --no-setup --namespace $(TEST_NAMESPACE) --go-test-flags "-v -timeout=15m"
+	# remove me once verified
+	oc get kubefedcluster -n $(TEST_NAMESPACE)
+	oc get kubefedcluster -n $(HOST_NS)
+	oc logs $(oc get pods -o name) -n $(HOST_NS)
+	oc logs $(oc get pods -o name) -n $(TEST_NAMESPACE)
 
 .PHONY: e2e-setup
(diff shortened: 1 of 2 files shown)


make/test.mk
```diff
@@ -46,12 +46,14 @@ upload-codecov-report:
...
-test-e2e:  e2e-setup
-	sed -e 's|REPLACE_IMAGE|${IMAGE_NAME}|g' ./deploy/operator.yaml  | oc apply -f -
+test-e2e:  deploy-host e2e-setup setup-kubefed
 	# This is hack to fix https://github.com/operator-framework/operator-sdk/issues/1657
 	echo "info: Running go mod vendor"
 	go mod vendor
 	operator-sdk test local ./test/e2e --no-setup --namespace $(TEST_NAMESPACE) --go-test-flags "-v -timeout=15m"
+	# remove me once verified
+	oc get kubefedcluster -n $(TEST_NAMESPACE)
+	oc get kubefedcluster -n $(HOST_NS)
```
MatousJobanek: do we still need them?

make/test.mk
```diff
@@ -82,4 +90,23 @@ get-test-namespace: $(OUT_DIR)/test-namespace
...
+deploy-host:
+	$(eval HOST_NS := $(shell echo -n "host-operator-$(shell date +'%s')"))
+	rm -rf /tmp/host-operator
+	# cloning shallow as don't want to maintain it for every single change in deploy directory of host-operator
+	git clone https://github.com/codeready-toolchain/host-operator.git --depth 1 /tmp/host-operator
+	oc new-project $(HOST_NS)
+	oc apply -f /tmp/host-operator/deploy/service_account.yaml
+	oc apply -f /tmp/host-operator/deploy/role.yaml
+	cat /tmp/host-operator/deploy/role_binding.yaml | sed s/\REPLACE_NAMESPACE/$(HOST_NS)/ | oc apply -f -
+	oc apply -f /tmp/host-operator/deploy/crds
```
MatousJobanek: wouldn't be better to call the make target?

make/test.mk
dipak-pawar: I was thinking to remove it, but later on realized it would be useful to debug, if something goes wrong, at-least you would be sure about cluster is syncing, otherwise we need to wait for another more 30 minutes to verify it.

make/test.mk
dipak-pawar: I have written this in this repo as this way it's quite easy to follow, otherwise you need to see what's in make target in another project by changing the project

PR #39, opened 2019-07-19
CRT-175: Implement garbage collection for UserAccount

tinakurian:
UserAccount owns the following resources:

User
Identity
NSTemplateSet (we don't use it yet)

However,  when we delete a UserAccount the owned User and Identity are not deleted automatically despite the fact that we set the owner reference correctly. This may be due to the fact that User and Identity are cluster scoped whereas UserAccount is namespace scoped. 

Related to task: https://jira.coreos.com/browse/CRT-175

_Resources:_
1. https://hackernoon.com/level-triggering-and-reconciliation-in-kubernetes-1f17fe30333d
2. https://github.com/kubernetes-sigs/controller-runtime/issues/261#issuecomment-448593754
3. https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#finalizers
4. https://book.kubebuilder.io/reference/using-finalizers.html
5. https://github.com/hossainemruz/k8s-initializer-finalizer-practice/blob/master/pkg/controller/deploymentcontroller.go
6. https://github.com/istio/operator/blob/master/pkg/controller/istiocontrolplane/istiocontrolplane_controller.go


alexeykazakov:
Has any research been done why the built in garbage collection doesn’t work for UserAccont - User/Identity? Is it because User/Identity are cluster scoped? Do we know it for sure or it’s just a guess? Do we know if it’s possible to still get them garbage collected without implementing our own finalizer? Maybe some additional properties/fields?

Just want to make sure we fully understand how it works.

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -213,6 +230,69 @@ func (r *ReconcileUserAccount) ensureIdentity(logger logr.Logger, userAcc *toolc
...
+		if err := r.client.Update(context.TODO(), userAcc); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Handles finalizer logic for UserAccount
+func (r *ReconcileUserAccount) finalizer(logger logr.Logger, userAcc *toolchainv1alpha1.UserAccount) (bool, error) {
```
sbryzak: Could we have a comment here describing what the `bool` return value means?

sbryzak:
In regards to the comment from @alexeykazakov, are we doing a foreground or background delete?  A background delete should take care of the garbage collection itself.  I did some research and couldn't find any information in relation to deleting cluster-scoped objects where the ownerReference is namespace-scoped.

pkg/controller/useraccount/useraccount_controller.go
tinakurian: Thanks @sbryzak! Fixed here: d744e71b1270e079f720db29090851e268f8585d

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -107,6 +110,20 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
 	}
 
+	// Add the finalizer if it is not present
+	if err := r.setFinalizers(userAcc); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Check if the UserAccount has been deleted
+	if !userAcc.ObjectMeta.DeletionTimestamp.IsZero() {
+		rec, err := r.finalizer(reqLogger, userAcc)
```
nurali-techie: should we also check if `err != nil` here ??

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -213,6 +230,67 @@ func (r *ReconcileUserAccount) ensureIdentity(logger logr.Logger, userAcc *toolc
...
+
+	// Get the Identity associated with the UserAccount
+	identity := &userv1.Identity{}
+	identityName := ToIdentityName(userAcc.Spec.UserID)
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: identityName}, identity); err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return false, nil
```
nurali-techie: if `errors.IsNotFound(err)` true then I guess we can continue rather returning.  It's like you want to remove Identity and User as part of finalizer() and you found that Identity is not there (may be already deleted) which is fine and you should continue deleting User.  So we should not return from here.

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -213,6 +230,67 @@ func (r *ReconcileUserAccount) ensureIdentity(logger logr.Logger, userAcc *toolc
...
+	logger.Info("deleting UserAccount and subsequent resources", "name", userAcc.Name)
+
+	// Get the Identity associated with the UserAccount
+	identity := &userv1.Identity{}
+	identityName := ToIdentityName(userAcc.Spec.UserID)
+	if err := r.client.Get(context.TODO(), types.NamespacedName{Name: identityName}, identity); err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
```
nurali-techie: I guess this comment are not applicable here.  It's comes from copy/paste and not relevant and not required here.

nurali-techie: I guess this comment are not applicable here.  It's comes from copy/paste and not relevant and not required here.

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -213,6 +230,67 @@ func (r *ReconcileUserAccount) ensureIdentity(logger logr.Logger, userAcc *toolc
...
+			return false, nil
+		}
+		return false, err
+	}
+
+	// Delete Identity associated with UserAccount
+	if err := r.client.Delete(context.TODO(), identity); err != nil {
+		return true, err
```
nurali-techie: if error is not_nil, no need to set requeue to true.  check the doc of `Reconcile()` method

```
// The Controller will requeue the Request to be processed again if the returned error is non-nil or
// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
```

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -327,6 +327,54 @@ func TestReconcile(t *testing.T) {
...
+			assert.Equal(t, fmt.Sprintf("%s:%s", config.GetIdP(), userAcc.Spec.UserID), identity.Name)
+
+			// Check the user identity mapping
+			checkMapping(t, preexistingUser, identity)
+
+			err = r.client.Delete(context.TODO(), userAcc)
+			require.NoError(t, err)
+
+			err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: req.Namespace, Name: userAcc.Name}, userAcc)
+			require.Error(t, err)
```
nurali-techie: you can also assert.True(t, errors.IsNotFound(err))

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -213,6 +230,67 @@ func (r *ReconcileUserAccount) ensureIdentity(logger logr.Logger, userAcc *toolc
...
+		if err := r.client.Update(context.TODO(), userAcc); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Handles finalizer logic for UserAccount. Returns bool to indicate Requeue
+func (r *ReconcileUserAccount) finalizer(logger logr.Logger, userAcc *toolchainv1alpha1.UserAccount) (bool, error) {
```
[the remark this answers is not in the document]
alexeykazakov: Also, `finalizer` is not very clear name here IMO... it's about executing the finalizer(s), right? So, maybe "finalize()" would be better? Let's find some examples to see if there any convention or common naming in the community.

pkg/controller/useraccount/useraccount_controller.go
alexeykazakov: Yeah.. Also I think we don't have to load both identity and user before deleting them. We should do something like:
- Get User. If doesn't exist then proceed. If exists then delete it and exit the reconcile.
- When the user is deleted it will trigger another reconcile. Get Identity. If exists then delete it and exit reconcile.
- When identity is deleted then one more reconcile is triggered. Everything is deleted then so we can delete the finalizer from meta and exit.

pkg/controller/useraccount/useraccount_controller.go
tinakurian: Refactored the code - hopefully this is alright: 51e17d72b87aa88373f9bdb8dd5dd696b2b29847

pkg/controller/useraccount/useraccount_controller.go
tinakurian: @nurali-techie Fixed here: 51e17d72b87aa88373f9bdb8dd5dd696b2b29847

tinakurian: @nurali-techie Fixed here: 51e17d72b87aa88373f9bdb8dd5dd696b2b29847

pkg/controller/useraccount/useraccount_controller.go
tinakurian: @nurali-techie Fixed here: 51e17d72b87aa88373f9bdb8dd5dd696b2b29847

pkg/controller/useraccount/useraccount_controller_test.go
tinakurian: @nurali-techie Fixed here: 2cdc12f48fc46eeccbc8b0e01abf9ca7d1717975

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -213,6 +230,67 @@ func (r *ReconcileUserAccount) ensureIdentity(logger logr.Logger, userAcc *toolc
 	return identity, false, nil
 }
 
+// Sets the finalizers for UserAccount
```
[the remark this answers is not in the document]
tinakurian: @alexeykazakov Fixed here: 539d8273b44986cb0de8e103acdfab001912cc73

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -213,6 +230,67 @@ func (r *ReconcileUserAccount) ensureIdentity(logger logr.Logger, userAcc *toolc
...
+		userAcc.ObjectMeta.Finalizers = append(userAcc.ObjectMeta.Finalizers, userAccFinalizerName)
+		if err := r.client.Update(context.TODO(), userAcc); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Handles finalizer logic for UserAccount. Returns bool to indicate Requeue
```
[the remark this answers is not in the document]
tinakurian: @alexeykazakov Fixed here: 539d8273b44986cb0de8e103acdfab001912cc73

pkg/controller/useraccount/useraccount_controller.go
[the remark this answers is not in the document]
tinakurian: @alexeykazakov Agreed, i removed the bool when i was refactoring the code to address Nurali's comments! That was done here: 51e17d72b87aa88373f9bdb8dd5dd696b2b29847

I have also updated the func name to ```finalize``` instead here: 7f7962061e1c670b263f24104e66c64e0a12c3a3. 

I did see an example which is what i based these changes off of, let me see if in can find it again! Will add it to this comment when i find it!

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -327,6 +329,62 @@ func TestReconcile(t *testing.T) {
...
+		require.NoError(t, err)
+		assert.Equal(t, fmt.Sprintf("%s:%s", config.GetIdP(), userAcc.Spec.UserID), identity.Name)
+
+		// Check the user identity mapping
+		checkMapping(t, preexistingUser, identity)
+
+		userAcc.DeletionTimestamp = &metav1.Time{time.Now()} //nolint: govet
+		err = r.client.Update(context.TODO(), userAcc)
+		require.NoError(t, err)
+		res, err = r.Reconcile(req)
```
golangcibot: ineffectual assignment to `res` (from `ineffassign`)

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -327,6 +329,62 @@ func TestReconcile(t *testing.T) {
...
+		require.NoError(t, err)
+		res, err = r.Reconcile(req)
```
golangcibot: ineffectual assignment to `res` (from `ineffassign`)

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -327,6 +329,62 @@ func TestReconcile(t *testing.T) {
...
+		require.NoError(t, err)
+		res, err = r.Reconcile(req)
```
golangcibot: ineffectual assignment to `res` (from `ineffassign`)

pkg/controller/useraccount/useraccount_controller.go
tinakurian: @alexeykazakov thanks for pointing this out. I have refactored the code to follow this solution instead

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -107,16 +110,39 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+		if user, createdOrUpdated, err = r.ensureUser(reqLogger, userAcc); err != nil || createdOrUpdated {
+			return reconcile.Result{}, err
+		}
+
+		if _, createdOrUpdated, err = r.ensureIdentity(reqLogger, userAcc, user); err != nil || createdOrUpdated {
+			return reconcile.Result{}, err
+		}
+
+		// Add the finalizer if it is not present
+		if err := r.setFinalizers(userAcc); err != nil {
```
nurali-techie: call to `setFinalizers` should be first .. i.e. before `ensureUser` and `ensureIdentity`

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -213,6 +239,79 @@ func (r *ReconcileUserAccount) ensureIdentity(logger logr.Logger, userAcc *toolc
...
+
+	return nil
+}
+
+// deleteUser deletes the user resource
+func (r *ReconcileUserAccount) deleteUser(logger logr.Logger, userAcc *toolchainv1alpha1.UserAccount) (error, bool) {
+	// Get the User associated with the UserAccount
+	user := &userv1.User{}
+	userErr := r.client.Get(context.TODO(), types.NamespacedName{Name: userAcc.Name}, user)
+	if userErr != nil {
```
nurali-techie: If you don't mind, can we use `err` for any error unless we want to distinguish two different error in same scope.   I am asking as generally in our team `err` is used rather `<prefix>err<suffix>` and that's also idiomatic go way.  This is to make sure we have consistence coding style throughout our codebase.

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -213,6 +239,79 @@ func (r *ReconcileUserAccount) ensureIdentity(logger logr.Logger, userAcc *toolc
...
+	identity := &userv1.Identity{}
+	identityName := ToIdentityName(userAcc.Spec.UserID)
+	identityErr := r.client.Get(context.TODO(), types.NamespacedName{Name: identityName}, identity)
+	if identityErr != nil {
+		if !errors.IsNotFound(identityErr) {
+			return identityErr, false
+		}
+	}
+
+	// Delete Identity associated with UserAccount if the identity exists
```
nurali-techie: comment mis-placed or redundant ;-)

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -107,16 +110,39 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+		if err, deleted = r.deleteUser(reqLogger, userAcc); err != nil || deleted {
+			return reconcile.Result{}, err
+		}
+
+		if err = r.deleteFinalizer(reqLogger, userAcc); err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 
+	if userAcc.ObjectMeta.DeletionTimestamp.IsZero() {
```
nurali-techie: this can just go into `else` part.

nurali-techie:
@tinakurian the code looks better and clean, thank you :-)  Let me know when everything is ready so that I will take this PR on my machine and will do one round of manual testing :+1:

test/e2e/useraccount_test.go
```diff
@@ -109,6 +113,32 @@ func TestUserAccount(t *testing.T) {
 		assert.NoError(t, err)
 	})
 
+	t.Run("delete_useraccount_ok", func(t *testing.T) {
+		lucyAcc := &toolchainv1alpha1.UserAccount{}
+		err := client.Get(context.TODO(), types.NamespacedName{Name: lucyUserAcc.Name, Namespace: namespace}, lucyAcc)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

test/e2e/useraccount_test.go
```diff
@@ -109,6 +113,32 @@ func TestUserAccount(t *testing.T) {
...
+		err = client.Update(context.TODO(), lucyAcc)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

test/e2e/useraccount_test.go
```diff
@@ -109,6 +113,32 @@ func TestUserAccount(t *testing.T) {
...
+
+		err = client.Get(context.TODO(), types.NamespacedName{Name: lucyAcc.Name, Namespace: namespace}, lucyAcc)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

test/e2e/useraccount_test.go
```diff
@@ -109,6 +113,32 @@ func TestUserAccount(t *testing.T) {
...
+		err = client.Update(context.TODO(), lucyAcc)
+
+		client.Delete(context.TODO(), lucyAcc)
```
golangcibot: Error return value of `client.Delete` is not checked (from `errcheck`)

pkg/controller/useraccount/useraccount_controller.go
tinakurian: Fixed in refactor! :-)

pkg/controller/useraccount/useraccount_controller.go
tinakurian: Fixed here: 008771ba402bc3dafb894c1eb8d0a7851f07df0d

pkg/controller/useraccount/useraccount_controller.go
tinakurian: fixed here: 008771ba402bc3dafb894c1eb8d0a7851f07df0d

pkg/controller/useraccount/useraccount_controller.go
tinakurian: Fixed here: 008771ba402bc3dafb894c1eb8d0a7851f07df0d

make/test.mk
```diff
@@ -44,14 +44,14 @@ upload-codecov-report:
...
-
 .PHONY: test-e2e
 test-e2e:  e2e-setup
 	sed -e 's|REPLACE_IMAGE|${IMAGE_NAME}|g' ./deploy/operator.yaml  | oc apply -f -
 	# This is hack to fix https://github.com/operator-framework/operator-sdk/issues/1657
 	echo "info: Running go mod vendor"
 	go mod vendor
-	operator-sdk test local ./test/e2e --no-setup --namespace $(TEST_NAMESPACE) --go-test-flags "-v -timeout=15m"
+	oc login -u system:admin
+	operator-sdk test local ./test/e2e --namespace $(TEST_NAMESPACE) --up-local --go-test-flags "-v -timeout=15m"
```
nurali-techie: @tinakurian what's the need of these changes ?? Also, removal of `--no-setup` is intentional ??

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -107,16 +110,37 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, err
 	}
 
-	var createdOrUpdated bool
-	var user *userv1.User
-	if user, createdOrUpdated, err = r.ensureUser(reqLogger, userAcc); err != nil || createdOrUpdated {
-		return reconcile.Result{}, err
-	}
+	// Check if the UserAccount has been deleted. If it has been deleted, delete
```
nurali-techie: this comment is for `else` part and not for `if` part.  I guess, it's due to refactoring code and swap if-else but you forgot to move comment :-(

test/e2e/wait_util.go
```diff
@@ -77,3 +77,45 @@ func waitForUserAccStatusConditions(t *testing.T, client client.Client, namespac
...
+			return false, err
+		}
+		return false, nil
+	})
+}
+
+func waitForDeletedUser(t *testing.T, client client.Client, name string, namespace string) error {
+	return wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+		user := &toolchainv1alpha1.UserAccount{}
+		if err := client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, user); err != nil {
```
nurali-techie: namespace not required, plz remove for both - user and identity.

test/e2e/wait_util.go
```diff
@@ -77,3 +77,45 @@ func waitForUserAccStatusConditions(t *testing.T, client client.Client, namespac
...
+			}
+			return false, err
+		}
+		return false, nil
+	})
+}
+
+func waitForDeletedUser(t *testing.T, client client.Client, name string, namespace string) error {
+	return wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+		user := &toolchainv1alpha1.UserAccount{}
```
nurali-techie: this should be `&userv1.User{}`

test/e2e/wait_util.go
```diff
@@ -77,3 +77,45 @@ func waitForUserAccStatusConditions(t *testing.T, client client.Client, namespac
...
+		if err := client.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, user); err != nil {
+			if errors.IsNotFound(err) {
+				t.Logf("deleted user '%s'", name)
+				return true, err
```
nurali-techie: I guess, here rather returning `err` we should return nil to complete polling immediately ??

test/e2e/wait_util.go
```diff
@@ -77,3 +77,45 @@ func waitForUserAccStatusConditions(t *testing.T, client client.Client, namespac
...
+				t.Logf("deleted user account '%s'", name)
+				return true, err
+			}
+			return false, err
+		}
+		return false, nil
+	})
+}
+
+func waitForDeletedUser(t *testing.T, client client.Client, name string, namespace string) error {
```
nurali-techie: `namespace` input param not needed.

test/e2e/wait_util.go
```diff
@@ -77,3 +77,45 @@ func waitForUserAccStatusConditions(t *testing.T, client client.Client, namespac
...
+			}
+			return false, err
+		}
+		return false, nil
+	})
+}
+
+func waitForDeletedIdentity(t *testing.T, client client.Client, name string, namespace string) error {
+	return wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+		user := &userv1.Identity{}
```
nurali-techie: the variable name `identity` and NOT `user`

make/test.mk
alexeykazakov: Yeah... `oc login -u system:admin` looks suspicious. Maybe our SA is missing some permissions? cc: @xcoulon

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -427,11 +485,13 @@ func TestUpdateStatus(t *testing.T) {
...
+	finalizers := []string{userAccFinalizerName}
 	userAcc := &toolchainv1alpha1.UserAccount{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      userName,
-			Namespace: "toolchain-member",
-			UID:       types.UID(uuid.NewV4().String()),
+			Name:       userName,
+			Namespace:  "toolchain-member",
+			UID:        types.UID(uuid.NewV4().String()),
+			Finalizers: finalizers,
```
alexeykazakov: Don't add finalizers for every preexisting account. We must test the logic which adds the finalizer in our controller during reconciling.

pkg/controller/useraccount/useraccount_controller.go
tinakurian: Sorry, i didn't forget to move the comment. I left it there to explain the else block and there was no comment for the addition of user/identity. But i can add one to the comment

test/e2e/wait_util.go
tinakurian: i wanted to return error so the test was more clear on what was being checked. however, if you would like me to return nil and the test not verify that the error is "IsNotFound" i can do that as well...

make/test.mk
tinakurian: Sorry, this was from an earlier commit. It's now reverted back to what we had!

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -327,6 +329,62 @@ func TestReconcile(t *testing.T) {
...
+		r, req, _ := prepareReconcile(t, username, userAcc, preexistingUser, preexistingIdentity)
+
+		//when
+		res, err := r.Reconcile(req)
+
+		//then
+		require.NoError(t, err)
+		assert.Equal(t, reconcile.Result{}, res)
+
+		// Check the created user
```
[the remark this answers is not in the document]
tinakurian: Done above!

pkg/controller/useraccount/useraccount_controller_test.go
tinakurian: I need to look further into this. Without adding it here, many of the other tests fail

test/e2e/wait_util.go
tinakurian: hmm.. i've updated it to remove the error. i guess the method name is clear enough to describe what is going on. removed the checks from the test as well!

tinakurian:
> Has any research been done why the built in garbage collection doesn’t work for UserAccont - User/Identity? Is it because User/Identity are cluster scoped? Do we know it for sure or it’s just a guess? Do we know if it’s possible to still get them garbage collected without implementing our own finalizer? Maybe some additional properties/fields?
> 
> Just want to make sure we fully understand how it works.

@alexeykazakov @nurali-techie it seems this is disallowed by design - from this link: https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents

_**Note:** Cross-namespace owner references are disallowed by design. This means: 1) Namespace-scoped dependents can only specify owners in the same namespace, and owners that are cluster-scoped. 2) Cluster-scoped dependents can only specify cluster-scoped owners, but not namespace-scoped owners._

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -156,6 +158,13 @@ func TestReconcile(t *testing.T) {
 				})
 		})
 		t.Run("update", func(t *testing.T) {
+			r, req, _ := prepareReconcile(t, username, userAcc, preexistingUser, preexistingIdentity)
+			res, err := r.Reconcile(req)
```
golangcibot: ineffectual assignment to `res` (from `ineffassign`)

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -327,6 +329,62 @@ func TestReconcile(t *testing.T) {
...
+		checkMapping(t, user, preexistingIdentity)
+
+		// Check the created/updated identity
+		identity := &userv1.Identity{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: ToIdentityName(userAcc.Spec.UserID)}, identity)
+		require.NoError(t, err)
+		assert.Equal(t, fmt.Sprintf("%s:%s", config.GetIdP(), userAcc.Spec.UserID), identity.Name)
+
+		// Check the user identity mapping
+		checkMapping(t, preexistingUser, identity)
```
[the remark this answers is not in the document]
tinakurian: @alexeykazakov i have removed this!

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -317,3 +407,22 @@ func newIdentity(userAcc *toolchainv1alpha1.UserAccount, user *userv1.User) *use
 func ToIdentityName(userID string) string {
 	return fmt.Sprintf("%s:%s", config.GetIdP(), userID)
 }
+
+func ContainsString(items []string, value string) bool {
```
tinakurian: Perhaps these utility funcs should have a better home?

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -317,3 +407,22 @@ func newIdentity(userAcc *toolchainv1alpha1.UserAccount, user *userv1.User) *use
 func ToIdentityName(userID string) string {
 	return fmt.Sprintf("%s:%s", config.GetIdP(), userID)
 }
+
+func containsString(items []string, value string) bool {
```
tinakurian: Perhaps these utility funcs should have a better home?

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -107,16 +111,39 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+		var user *userv1.User
+		if user, createdOrUpdated, err = r.ensureUser(reqLogger, userAcc); err != nil || createdOrUpdated {
+			return reconcile.Result{}, err
+		}
 
+		if _, createdOrUpdated, err = r.ensureIdentity(reqLogger, userAcc, user); err != nil || createdOrUpdated {
+			return reconcile.Result{}, err
+		}
+	} else {
+		if containsString(userAcc.ObjectMeta.Finalizers, userAccFinalizerName) {
```
xcoulon: minor suggestion: merge the `else` and `if` statements

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -107,16 +111,39 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
-	var createdOrUpdated bool
-	var user *userv1.User
-	if user, createdOrUpdated, err = r.ensureUser(reqLogger, userAcc); err != nil || createdOrUpdated {
-		return reconcile.Result{}, err
-	}
+	// If the UserAccount has not been deleted, create or update user and identity resources.
+	// If the UserAccount has been deleted, delete secondary resources identity and user.
+	if userAcc.ObjectMeta.DeletionTimestamp.IsZero() {
+		// Add the finalizer if it is not present
+		if err := r.setFinalizers(userAcc); err != nil {
```
xcoulon: minor suggestion: pass the finalizer name as an arg here, so we can "match" it with the call in line 132 (you could even use a constant for its name)

pkg/controller/useraccount/useraccount_controller.go
xcoulon: yes ;)

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -317,3 +408,22 @@ func newIdentity(userAcc *toolchainv1alpha1.UserAccount, user *userv1.User) *use
...
+	for _, item := range items {
+		if item == value {
+			return true
+		}
+	}
+	return false
+}
+
+func removeString(items []string, value string) (result []string) {
+	for _, item := range items {
```
xcoulon: minor suggestion: you could probably initialize the `result` array with a capacity of `len(items) - 1`

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -317,3 +408,22 @@ func newIdentity(userAcc *toolchainv1alpha1.UserAccount, user *userv1.User) *use
...
+func containsString(items []string, value string) bool {
+	for _, item := range items {
+		if item == value {
+			return true
+		}
+	}
+	return false
+}
+
+func removeString(items []string, value string) (result []string) {
```
xcoulon: minor suggestion: we don't usually name the returned value

alexeykazakov:
> it seems this is disallowed by design

I see. This explains why Account - User/Identity owner reference doesn't work. Thanks for the link.

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -327,6 +331,144 @@ func TestReconcile(t *testing.T) {
...
+	})
+	// Delete useraccount and ensure related resources are also removed
+	t.Run("set finalizer fails", func(t *testing.T) {
+		// given
+		userAcc := newUserAccount(username, userID)
+		r, req, client := prepareReconcile(t, username, userAcc, preexistingUser, preexistingIdentity)
+
+		// Mock setting finalizer failure
+		client.MockUpdate = func(obj runtime.Object) error {
+			return errors.New(fmt.Sprintf("unable to set finalizer for user account %s", userAcc.Name))
```
golangcibot: S1028: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (from `gosimple`)

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -327,6 +331,144 @@ func TestReconcile(t *testing.T) {
...
+		// Check that the associated user has been deleted
+		// when reconciling the useraccount with a deletion timestamp
+		user := &userv1.User{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: userAcc.Name}, user)
+		require.Error(t, err)
+		assert.True(t, apierros.IsNotFound(err))
+
+		// Mock finalizer reoval failure
+		client.MockUpdate = func(obj runtime.Object) error {
+			return errors.New(fmt.Sprintf("unable to delete finalizer for user account %s", userAcc.Name))
```
golangcibot: S1028: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (from `gosimple`)

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -389,6 +389,86 @@ func TestReconcile(t *testing.T) {
...
+		// Mock setting finalizer failure
+		client.MockUpdate = func(obj runtime.Object) error {
+			return errors.New(fmt.Sprintf("unable to set finalizer for user account %s", userAcc.Name))
+		}
+
+		//when
+		res, err := r.Reconcile(req)
+
+		//then
+		require.Error(t, err)
```
alexeykazakov: Please also check the error message to make sure you got the expected error.

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -389,6 +389,86 @@ func TestReconcile(t *testing.T) {
...
+		assert.True(t, apierros.IsNotFound(err))
+
+		// Mock finalizer reoval failure
+		client.MockUpdate = func(obj runtime.Object) error {
+			return errors.New(fmt.Sprintf("unable to delete finalizer for user account %s", userAcc.Name))
+		}
+
+		res, err = r.Reconcile(req)
+		assert.Equal(t, reconcile.Result{}, res)
+		require.Error(t, err)
```
alexeykazakov: Same here.

dipak-pawar:
@tinakurian, Thank you for your work. Can you please take a look into https://github.com/redhat-cop/operator-utils/blob/master/pkg/util/finalizer.go and see if you can improve this implementation. I feel it's more readable and optimized.

test/e2e/useraccount_test.go
```diff
@@ -109,6 +112,24 @@ func TestUserAccount(t *testing.T) {
...
+		err = client.Delete(context.TODO(), lucyAcc)
+		assert.NoError(t, err)
+
+		err = waitForDeletedUserAccount(t, client, lucyAcc.Name, namespace)
+		assert.NoError(t, err)
+
+		err = waitForDeletedUser(t, client, lucyAcc.Name)
+		assert.NoError(t, err)
+
+		err = waitForDeletedIdentity(t, client, lucyAcc.Name)
```
dipak-pawar: Can you please verify that `useridentitymapping` is also deleting? https://docs.openshift.com/container-platform/3.7/rest_api/apis-user.openshift.io/v1.UserIdentityMapping.html#Post-apis-user.openshift.io-v1-useridentitymappings

tinakurian: Looks like i cannot get the useridentitymapping resource. I see the following error: 
```
tkurian@localhost host-operator]$ oc get useridentitymapping
Error from server (MethodNotAllowed): the server does not allow this method on the requested resource
```
It is also outlined in this defect: https://bugzilla.redhat.com/show_bug.cgi?id=1684311

But as you suggested, i have tried adding and removing and re-adding the same userAccount and nothing fails!

alexeykazakov: @dipak-pawar useridentitymapping is a special resource. It's just a REST API for CREATE. In reality it modifies User and Identity. It doesn't have it's own etcd representation (nor get, watch, etc API). So, if both User and Identity are deleted then there is no any UserIdentityMapping either.

tinakurian:
> @tinakurian, Thank you for your work. Can you please take a look into https://github.com/redhat-cop/operator-utils/blob/master/pkg/util/finalizer.go and see if you can improve/reuse this implementation. I feel it's more readable and optimized. You can check how they're using it https://github.com/redhat-cop/operator-utils#managing-cr-finalization

Thanks for pointing me to this library Dipak. I decided to go with using the library. The only part i felt a little uncomfortable with was having the HasFinalizer check being called twice (once by our code and once by the library) when attempting to add the finalizer. But overall, it is a lot cleaner

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -107,16 +112,28 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+		var user *userv1.User
+		if user, createdOrUpdated, err = r.ensureUser(reqLogger, userAcc); err != nil || createdOrUpdated {
+			return reconcile.Result{}, err
+		}
 
+		if _, createdOrUpdated, err = r.ensureIdentity(reqLogger, userAcc, user); err != nil || createdOrUpdated {
+			return reconcile.Result{}, err
+		}
+	} else if util.HasFinalizer(userAcc, userAccFinalizerName) {
+		if err = r.manageCleanUpLogic(reqLogger, userAcc); err != nil {
```
dipak-pawar: Can we rename `manageCleanUpLogic` ?

tinakurian: @dipak-pawar this name is from the library as the method to implement. Not sure we should renmae this if we are going the library approach?

dipak-pawar: @tinakurian, my point here is if we can name a function from which we can understand what's happening inside the function. e.g.`cleanUpUserAndIdentiy()`, then it'd be good.

PR #40, opened 2019-07-26
User namespace provisioning

nurali-techie:
When User is Approved, new UserAccount created.
When UserAccount created, we also need to provision set of namespaces for the User depending on Tier assigned.  For ex - ide, cicd, stage etc.

This PR ..
- enhance UserAccount controller to create NSTemplateSet entity.
- introduce new controller for NSTemplateSet entity.
- create set of namespaces in NSTemplateSet controller.

TODO:
- [ ] create all namespaces (currently only one namespace created <username>-stage)
- [ ] support NSTemplateSet.Status
- [ ] auto test


deploy/crds/toolchain_v1alpha1_useraccount.yaml
```diff
@@ -58,34 +58,12 @@ spec:
...
-                tierName:
-                  description: The name of the tier represented by this template set
-                  type: string
-              required:
-              - tierName
-              - namespaces
+            nsTemplateSetName:
+              description: Reference to NSTemplateSet for this user
               type: object
+            tierName:
```
alexeykazakov: `tierName` is part of `NSTemplateSet`. Why do we also need it in `UserAccount`?

deploy/crds/toolchain_v1alpha1_useraccount.yaml
```diff
@@ -94,7 +72,7 @@ spec:
           required:
           - userID
           - nsLimit
-          - nsTemplateSet
```
alexeykazakov: `nsTemplateName` should be required, shouldn't it?

pkg/apis/apis.go
```diff
@@ -12,6 +14,8 @@ var AddToSchemes runtime.SchemeBuilder
 func AddToScheme(s *runtime.Scheme) error {
 	// add openshift specific resource
 	AddToSchemes = append(AddToSchemes, userv1.AddToScheme)
+	AddToSchemes = append(AddToSchemes, templatev1.AddToScheme)
```
alexeykazakov: Hm.. Why do we need templates and project schemes here?

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -0,0 +1,140 @@
...
+		return reconcile.Result{}, err
+	}
+
+	result, err := r.ensureNamespaces(reqLogger, nsTeplSet)
+	return result, err
+}
+
+func (r *ReconcileNSTemplateSet) ensureNamespaces(logger logr.Logger, nsTeplSet *toolchainv1alpha1.NSTemplateSet) (reconcile.Result, error) {
+	// TODO create multiple namespaces
+	name := fmt.Sprintf("%s-stage", nsTeplSet.Name)
```
alexeykazakov: I wonder if we should use the NS Template namespace type to construct the name: `<nsTemplateSet.Name>-<nsTemplateSet.namespace.type>`

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -314,6 +375,25 @@ func newIdentity(userAcc *toolchainv1alpha1.UserAccount, user *userv1.User) *use
...
+	}
+	return nsTeplSet
+}
+
 func ToIdentityName(userID string) string {
 	return fmt.Sprintf("%s:%s", config.GetIdP(), userID)
 }
+
+func ToNSTemplateSetName(name string) string {
+	return name
```
alexeykazakov: We will probably need to do some validation of the username here. Not sure if we always can use the username as the namespace name + some prefix (stage, ide, etc). Can we exceed the namespace name length? Can username contain symbols that are not allowed in namespace names (probably not because both of them are k8s resource names), etc.

deploy/crds/toolchain_v1alpha1_useraccount.yaml
alexeykazakov: I think I know why you added it but see my comments above.

nurali-techie: @alexeykazakov NSTemplateSet is created by UserAccountCtrl, to set NSTemplateSet.TierName we need UserAccount.TierName.

deploy/crds/toolchain_v1alpha1_useraccount.yaml
nurali-techie: UserAccount is created by MUR_ctrl .. I made this `nsTemplateName` optional so MUR_ctrl no need to create anything related to NSTemplateSet and creation of NSTemplateSet is totally handler by UserAccountCtrl.

pkg/apis/apis.go
nurali-techie: We are using it here [link](https://github.com/codeready-toolchain/member-operator/pull/40/files/e57cdfdbf3c8fc3dc72e46d8b6efe0f3863fc243#diff-cffea82f7fdc7427b8c9f0d0fdd55aedR13) in NSTemplateSetCtrl.  We need to add all scheme which are used in operator.  Let me know if you asking something more.

pkg/controller/nstemplateset/nstemplateset_controller.go
nurali-techie: this is temp, it will be addressed :+1:

nurali-techie:
Replaced by https://github.com/codeready-toolchain/member-operator/pull/53

PR #41, opened 2019-08-01
add dev cluster deploymentconfig

dipak-pawar:
**Changes Proposed in this PR**

- Adds ImageStream which points to image from openshift CI cluster.
- Add DeploymentConfig which uses above ImageStream
- Fix minor make targets

### DIFF
--- a/deploy/operator_dev_cluster.yaml
+++ b/deploy/operator_dev_cluster.yaml
@@ -22,7 +22,7 @@ metadata:
   name: member-operator
   namespace: toolchain-member-operator
 spec:
-  replicas: 1
+  replicas: 2
   revisionHistoryLimit: 10
   selector:
     app: member-operator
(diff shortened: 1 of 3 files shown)


deploy/operator_dev_cluster.yaml
```diff
@@ -0,0 +1,93 @@
...
+---
+apiVersion: apps.openshift.io/v1
+kind: DeploymentConfig
+metadata:
+  labels:
+    app: member-operator
+  name: member-operator
+  namespace: toolchain-member-operator
+spec:
+  replicas: 1
```
nurali-techie: would it be better to keep `replicas=2`, so that there will be two pod running for member-operator and all our code test against concurrency.  Also, to test sync between member and host .. replica=2 is more meaningful.

dipak-pawar: @nurali-techie, good point, but this is going to work with leader election, this means at any given time, there'll be just one controller which is acting and others are waiting to get elected as master

nurali-techie: @dipak-pawar as discussed, either keep `replica=2` OR find some rolling update strategy to make sure we have High-Availability during rolling update :+1:

PR #44, opened 2019-08-12
fix broken build as python3.6 is retired in centos7

dipak-pawar:
We need python3.6, for yamllint which needs to formal yaml files in lint goal. We have issues while installing python3.6 as it's retired from centos7. you can find issue link https://bugzilla.redhat.com/show_bug.cgi?id=1739804.

Now the workaround which we are using here is IUS, which stands for Inline with Upstream Stable rpm provider. I have tested this and it worked. it's just another way to install python3 rpm from another repo.

Found from how to install python3 blog https://www.digitalocean.com/community/tutorials/how-to-install-python-3-and-set-up-a-local-programming-environment-on-centos-7


dipak-pawar:
closing this as packages are back

PR #47, opened 2019-08-15
feat: add/modify make targets to clean e2e namespaces

MatousJobanek:
* add make target to easily delete all namespaces that were created by e2e tests
* add automatic deletion of namespaces after e2e test
* add a new target that skips deletion of namespaces after e2e test

### DIFF
--- a/make/test.mk
+++ b/make/test.mk
@@ -69,11 +69,17 @@ PULL_NUMBER := $(shell echo $$CLONEREFS_OPTIONS | jq '.refs[0].pulls[0].number')
 #
 ###########################################################
 
+.PHONY: test-e2e-keep-namespaces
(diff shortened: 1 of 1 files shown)


alexeykazakov:
Shouldn't our e2e tests clean up the namespaces? I mean it's fine to have a separate target for that but our e2e tests should clean up the namespaces after finishing too.

MatousJobanek:
@alexeykazakov Based on our discussion I:
* changed the current `test-e2e` target to automatically delete the created namespaces
* added a new target `test-e2e-keep-namespaces` that skips the deletion of the namespaces

I also decided to keep the `clean-e2e-namespaces` because when either test or test setup fails then the execution is stopped and namespaces are not deleted. This can end up with multiple namespaces stayed as zombies after unsuccessful runs

PR #48, opened 2019-08-19
remove vendor support

nurali-techie:
As operator-sdk 0.10.x has `go module` support, we can remove `vendor support`.


dipak-pawar:
@nurali-techie I would suggest you to remove vendor goal only from `up-local` make goal, and keep `clean`, `vendor` goals and gitignore as it is. The reason behind this is, you need vendor goal, to download all dependencies and use it from you vscode or goland. I do remember, this is why @xcoulon  decided to put vendor in this, even we started with go modules

nurali-techie:
@dipak-pawar make sense to support local dev environment :+1: 
Revert few changes, kindly review :-)

nurali-techie:
@dipak-pawar thanks for your review and feedback :-)

PR #49, opened 2019-08-20
WIP: process and apply templates

dipak-pawar:
process the template given its name (eg: basic)
apply the variables (hard-coded for now)
create the resources (here, a project and a role binding)

fixes CRT-197


pkg/templates/basic-user-template.yml
```diff
@@ -0,0 +1,43 @@
+---
```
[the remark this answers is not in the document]
dipak-pawar: @nurali-techie, I believe, we need this templates to be created in host repository. where we'll be managing  it using `templatetier`, this is just for the testing purpose. this is why I don't want to create top level directory for templates.

pkg/controller/nstemplateset/nstemplate_controller_test.go
```diff
@@ -0,0 +1,142 @@
...
+		tmplSet := &toolchainv1alpha1.NSTemplateSet{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: namespace,
+				Name:      name,
+			},
+			Spec: toolchainv1alpha1.NSTemplateSetSpec{
+				TierName: "basic",
+			},
+		}
+		cl.Create(context.TODO(), tmplSet)
```
golangcibot: Error return value of `cl.Create` is not checked (from `errcheck`)

pkg/controller/nstemplateset/nstemplate_controller_test.go
```diff
@@ -0,0 +1,142 @@
...
```
golangcibot: Error return value of `cl.Create` is not checked (from `errcheck`)

pkg/controller/nstemplateset/nstemplate_controller_test.go
```diff
@@ -0,0 +1,142 @@
...
+		verifyRoleBinding(t, r.client, namespace)
+
+		// delete rolebinding to create scenario, of rolebinding failed to create in first run.
+		rb, err := getRoleBinding(r.client, namespace)
+		require.NoError(t, err)
+
+		err = cl.Delete(context.TODO(), rb)
+		require.NoError(t, err)
+
+		result, err = r.Reconcile(req)
```
golangcibot: ineffectual assignment to `result` (from `ineffassign`)

pkg/templates/bindata.go
```diff
@@ -0,0 +1,249 @@
+// Code generated by go-bindata.
+// sources:
+// pkg/templates/basic-user-template.yml
+// DO NOT EDIT!
+
+package templates
```
alexeykazakov: Do we really need to keep that generated bindata in github? Let's generate it during build and keep in git ignore.

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -0,0 +1,125 @@
...
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	log.Info("processing NSTemplateSet...")
+	// TODO: use values from the request
+	values := map[string]string{
+		"PROJECT_NAME":    request.Namespace,
+		"ADMIN_USER_NAME": "developer",
```
alexeykazakov: It looks like we need to add `username` field to NSTemplateSet CRD.

Also, why `ADMIN_USER_NAME`? Wouldn't `USERNAME` be better? We have the only user in the template context.

pkg/templates/basic-user-template.yml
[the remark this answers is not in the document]
alexeykazakov: Fear enough. But in this case we would need to add a //TODO with the link to JIRA which will take care of that later :)

pkg/template/processor_test.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func logger(t *testing.T) logr.Logger {
+	zapLog, err := zap.NewDevelopment()
+	require.NoError(t, err)
+	return zapr.NewLogger(zapLog)
+}
+
+func expectedRoleBinding() string {
+	return `{"apiVersion":"authorization.openshift.io/v1","kind":"RoleBinding","metadata":{"labels":{"provider":"codeready-toolchain","version":"1a2b3c"},"name":"foo-admin","namespace":"foo"},"roleRef":{"name":"admin"},"subjects":[{"kind":"User","name":"developer"}]}
```
MatousJobanek: I have already written one related comment, having the template in a file and then expected processed outcome as a string is not the best

pkg/template/processor.go
```diff
@@ -0,0 +1,111 @@
...
+
+// injectUserVars injects user specified variables into the Template
+func injectUserVars(t *templatev1.Template, values map[string]string, ignoreUnknownParameters bool) error {
+	for param, val := range values {
+		v := templateprocessing.GetParameterByName(t, param)
+		if v != nil {
+			v.Value = val
+			v.Generate = ""
+		} else if !ignoreUnknownParameters {
+			return fmt.Errorf("unknown parameter name %q", param)
```
MatousJobanek: why would you fail if the template doesn't have the parameter? consider that we have 10 pre-configured variables and the templates use only subsets of them. There is no point in failing
In addition, consider custom templates - it shouldn't rely on the number of vars

pkg/template/processor_test.go
```diff
@@ -0,0 +1,223 @@
...
+		require.NoError(t, err, "failed to marshal json for projectrequest")
+		assert.Equal(t, expectedProjectRequest(), string(prJson))
+
+		// role binding
+		rbKind := objs[1].Object.GetObjectKind()
+		require.IsType(t, &unstructured.Unstructured{}, rbKind)
+		roleBinding := rbKind.(*unstructured.Unstructured)
+		rbJson, err := roleBinding.MarshalJSON()
+		require.NoError(t, err, "failed to marshal json for rolebinding")
+		assert.Equal(t, expectedRoleBinding(), string(rbJson))
```
MatousJobanek: I guess that the logic in these two verifications can be used as a custom assertion and call it only for the expected string/type

pkg/template/processor_test.go
```diff
@@ -0,0 +1,223 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"context"
+	authv1 "github.com/openshift/api/authorization/v1"
+	projectv1 "github.com/openshift/api/project/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"github.com/satori/go.uuid"
+	"github.com/codeready-toolchain/member-operator/pkg/apis"
+)
+
+func TestProcess(t *testing.T) {
```
MatousJobanek: I'd like to also see a test with default values set inside of the templates - that they are overridden.

pkg/template/processor_test.go
```diff
@@ -0,0 +1,223 @@
...
+	t.Run("ok", func(t *testing.T) {
+		s := scheme.Scheme
+		err := apis.AddToScheme(s)
+		require.NoError(t, err)
+		utilruntime.Must(apitemplate.Install(s)) // see https://github.com/openshift/oc/blob/master/cmd/oc/oc.go#L77
+
+		templateContent := templateContent(t)
+		pn := uuid.NewV4().String()
+		u := uuid.NewV4().String()
+		c := uuid.NewV4().String()
```
MatousJobanek: please use names for the variables. It's fine for one variable that you use one char, but in this combination it's not clear

pkg/template/processor_test.go
```diff
@@ -0,0 +1,223 @@
...
+		values := map[string]string{
+			"PROJECT_NAME": pn,
+			"COMMIT":       c,
+			"USER_NAME":    u,
+		}
+
+		cl := test.NewFakeClient(t)
+		p := template.NewProcessor(cl, s, logger(t))
```
MatousJobanek: didn't you forget to add `//when` and `// then` comments?

pkg/template/processor.go
```diff
@@ -0,0 +1,111 @@
...
+			rq.Info("template object is nil")
+			continue
+		}
+		gvk := obj.Object.GetObjectKind().GroupVersionKind()
+		rq.Info("processing object", "version", gvk.Version, "kind", gvk.Kind)
+		if err := cl.Create(context.TODO(), obj.Object); err != nil {
+			if errors.IsAlreadyExists(err) {
+				// If client failed to create all resources(few created, few remaining) in the first run, then to avoid
+				// resource already existing errors for later runs.(it's like oc apply, don't fail to exists)
+				continue
```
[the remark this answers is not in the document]
alexeykazakov: Right.
But `oc apply` actually doesn't recreate objects afaik. It creates new ones and patches the existing ones. So, maybe we can update the existing objects instead of removing/re-creating them?

pkg/template/processor.go
alexeykazakov: I agree. But logging a warning could still be useful for debugging.

pkg/template/processor_test.go
dipak-pawar: hmm, let me think how can we change this to make it maintainable in less effort.

make/go.mk
```diff
@@ -19,4 +19,8 @@ $(OUT_DIR)/operator:
 
 .PHONY: vendor
 vendor:
-	$(Q)go mod vendor
\ No newline at end of file
+	$(Q)go mod vendor
+
+.PHONY: generate
+generate:
+	go-bindata -pkg templates -o test/templates/template_data.go -nocompress -ignore=.*go test/templates/...
```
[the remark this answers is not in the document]
dipak-pawar: that's good point.

pkg/template/processor.go
alexeykazakov: We will have multiple templates (even for one user there will be multiple templates for different namespaces). We don't really know what template uses what parameters. If we require the exact match between the passing parameters and the template then the controller will have to prepare the parameters individually for each template which can be quite tricky.
It could just generate the same parameters every time and pass them to the processor for all the templates instead.

dipak-pawar: ok, now got it. will change this to not fail and log it as warning

pkg/template/processor.go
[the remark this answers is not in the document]
MatousJobanek: > I am not sure how do we get conflicts,

in my case conflict == response code 409. You will get it when you send a POST request to OS for an object that already exists (which is exactly what the client does)

Yeah, updating of the object will work too, just be sure that the spec is replaced by the new one - to ensure that all previous configuration is removed and replaced/updated by the new one

pkg/template/processor.go
```diff
@@ -0,0 +1,111 @@
...
+	objs, err := p.Process(tmplContent, values)
+	if err != nil {
+		return err
+	}
+
+	return apply(p.cl, objs, p.log)
+}
+
+func apply(cl client.Client, objs []runtime.RawExtension, rq logr.Logger) error {
+	for _, obj := range objs {
```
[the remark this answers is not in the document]
MatousJobanek: > I would prefer, to sort all template objects as per our need outside of this code.

sorting out of this particular code for sure. I would do it after the process logic when you get the slice of objects.  

> it has to make sure the template is in correct order for addition or deletion(reverse order)

You can take a look at tenant service - there is done exactly this 

> otherwise while getting the template from host should be in correct order of dependency.

I wouldn't really rely on this. I would expect that the template that is retrieved from host can be in a wrong order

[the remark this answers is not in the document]
MatousJobanek: > for namespace creation verification, it makes sense to do it upfront. I will add it.

please, check the tenant service

pkg/template/processor_test.go
MatousJobanek: take a look at tenant service tests - There are the same tests

pkg/template/processor.go
MatousJobanek: I guess that this is the reason why the tenant service didn't use this process logic, but implemented it's own replace - to have the flexibility.

examples/toolchain_v1alpha1_nstemplateset_cr.yaml
```diff
@@ -1,6 +1,7 @@
 apiVersion: toolchain.dev.openshift.com/v1alpha1
 kind: NSTemplateSet
 metadata:
+  name: johnsmith-templateset
```
nurali-techie: I guess, for NSTemplateSet.Name as of now there is NO need to add suffix `-templateset` and we can stay with NSTemplateSet.Name = `username`.

Also check design doc - https://docs.google.com/document/d/19UzaXVqiOo9JqDYz3qUKguJwW-7zceA4UtTr358R_CA/edit#heading=h.autst8l8rdnd

pkg/template/processor_test.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+func logger(t *testing.T) logr.Logger {
+	zapLog, err := zap.NewDevelopment()
+	require.NoError(t, err)
+	return zapr.NewLogger(zapLog)
+}
+
+func templateVars() (string, string, string) {
+	return uuid.NewV4().String(), uuid.NewV4().String(), uuid.NewV4().String()
```
nurali-techie: It would be good to add small prefix, like `prj-<uuid>`, `sha-<uuid>`, 'user-<uuid>', this will help in debugging failed test case in future and add readability to test data.

pkg/template/processor_test.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+func verifyResource(t *testing.T, objKind schema.ObjectKind, vars ...string) {
+	require.IsType(t, &unstructured.Unstructured{}, objKind)
```
nurali-techie: better to use `assert` to make sure that multiple calls to `verifyResource` executes and verify the resource rather stopping at first invalid resource verification.

pkg/template/processor_test.go
```diff
@@ -0,0 +1,345 @@
...
+	return []byte(tmpl + templateParams)
+}
+
+func logger(t *testing.T) logr.Logger {
+	zapLog, err := zap.NewDevelopment()
+	require.NoError(t, err)
+	return zapr.NewLogger(zapLog)
+}
+
+func templateVars() (string, string, string) {
```
nurali-techie: Minor feedback for better usability, incorporate at your will.
This method returns three values and how this three values are used is sort of pre-define.  For ex, first is project, second is commit and third is user.

This method can be enhance to take three input string.  For every blank_string it should generate name with uuid and if string is not blank then return the same string in output with preserve ordering.  With this the usage of this func will be help a lot in your test.

pkg/template/processor.go
```diff
@@ -0,0 +1,111 @@
...
+			return errs.Wrapf(err, "unable to create resource of type %s", gvk.Kind)
+		}
+	}
+	return nil
+}
+
+// Process processes the template with the given content
+func (p Processor) Process(tmplContent []byte, values map[string]string) ([]runtime.RawExtension, error) {
+	scheme := p.scheme
+	codecs := serializer.NewCodecFactory(scheme)
```
nurali-techie: I checked, it seems that creation of `CodecFactory` does lot of processing so it's better to add this as field to `Processor` struct and init only once.

pkg/template/processor.go
```diff
@@ -0,0 +1,111 @@
...
+	decode := codecs.UniversalDeserializer().Decode
+	obj, _, err := decode(tmplContent, nil, nil)
+	if err != nil {
+		p.log.Error(err, "unable to decode template")
+		return nil, err
+	}
+	tmpl := obj.(*templatev1.Template)
```
nurali-techie: better to type cast with assertion to avoid panic.

pkg/template/processor.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// Process processes the template with the given content
+func (p Processor) Process(tmplContent []byte, values map[string]string) ([]runtime.RawExtension, error) {
+	scheme := p.scheme
+	codecs := serializer.NewCodecFactory(scheme)
+	decode := codecs.UniversalDeserializer().Decode
+	obj, _, err := decode(tmplContent, nil, nil)
+	if err != nil {
+		p.log.Error(err, "unable to decode template")
```
nurali-techie: Method of `Processor` should not log any error, it should just return error.  This Processor is like core_functionality and it should just return error .. how to log error should be upto the caller (consumer).  Functionality like this log error is not desirable.

pkg/template/processor.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func apply(cl client.Client, objs []runtime.RawExtension, rq logr.Logger) error {
+	for _, obj := range objs {
+		if obj.Object == nil {
+			rq.Info("template object is nil")
+			continue
+		}
+		gvk := obj.Object.GetObjectKind().GroupVersionKind()
+		rq.Info("processing object", "version", gvk.Version, "kind", gvk.Kind)
```
nurali-techie: This will generate too much and not much useful logs.

pkg/apis/apis.go
```diff
@@ -2,14 +2,20 @@ package apis
...
 )
 
 // AddToScheme adds all Resources to the Scheme
 func AddToScheme(s *runtime.Scheme) error {
 	// add openshift specific resource
-	addToSchemes := append(apis.AddToSchemes, userv1.AddToScheme)
+	addToSchemes := append(apis.AddToSchemes, userv1.Install)
+	addToSchemes = append(addToSchemes, templatev1.Install)
+	addToSchemes = append(addToSchemes, projectv1.Install)
+	addToSchemes = append(addToSchemes, authv1.Install)
```
nurali-techie: This change will come (sooner or later) but if this PR changes doesn't need this then no need to add now here.

pkg/template/processor.go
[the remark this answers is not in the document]
nurali-techie: @dipak-pawar it would be great if you first check if this dependent resource creation is an issue for what usecase/scenario.  Can you please verify this https://jira.coreos.com/browse/CRT-197?focusedCommentId=119466&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-119466

pkg/apis/apis.go
dipak-pawar: @nurali-techie we need this to register, as we are processing/creating objects as a part of our tests

pkg/template/processor.go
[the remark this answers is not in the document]
dipak-pawar: @nurali-techie thanks for reminding me, I have already considered this, we'll be adding sorting logic for this

pkg/template/processor.go
```diff
@@ -47,13 +49,54 @@ func apply(cl client.Client, objs []runtime.RawExtension) error {
...
 		}
+		if gvk.Group == "project.openshift.io" && gvk.Version == "v1" && gvk.Kind == "ProjectRequest" {
+			if _, ok := obj.(runtime.Unstructured); ok {
+				var prq projectv1.ProjectRequest
+				err := p.scheme.Convert(obj, &prq, nil)
+				if err != nil {
+					return err
+				}
+				// wait until the Project exists and is ready
+				_, err = p.waitUntilProjectExists(prq.GetName(), 2*time.Second)
```
nurali-techie: `2 sec` timeout may be less any reference on arriving this number ??  anyway, we are polling on some interval .. so we can safely increase timeout.

pkg/template/processor.go
```diff
@@ -47,13 +49,54 @@ func apply(cl client.Client, objs []runtime.RawExtension) error {
 			continue
 		}
 		gvk := obj.GetObjectKind().GroupVersionKind()
-		if err := createOrUpdateObj(cl, obj); err != nil {
+		if err := createOrUpdateObj(p.cl, obj); err != nil {
 			return errs.Wrapf(err, "unable to create resource of kind: %s, version: %s", gvk.Kind, gvk.Version)
 		}
+		if gvk.Group == "project.openshift.io" && gvk.Version == "v1" && gvk.Kind == "ProjectRequest" {
```
nurali-techie: Other option is to keep this out of for loop.  First, we create project and wait for it to be active and then we run for_loop over other objects.  In this way, we can make sure that the template has ProjectRequest (which is mandatory) for given template.  Also, this if_cond no need to check in for_loop.

In any case, we should add logic that validate that given template has one project_request entry.

pkg/template/processor.go
```diff
@@ -47,13 +49,54 @@ func apply(cl client.Client, objs []runtime.RawExtension) error {
...
+			err := p.cl.Get(context.TODO(), types.NamespacedName{
+				Namespace: namespace,
+				Name:      "",
+			}, &prj)
+			// something wrong happened while checking the project
+			if err != nil && !apierrors.IsNotFound(err) {
+				return projectv1.Project{}, errs.Wrapf(err, "failed to check if project '%s' exists", namespace)
+			}
+			// project exists and is active
+			if err == nil && prj.Status.Phase == corev1.NamespaceActive {
```
nurali-techie: check `err == nil` not required.

pkg/template/processor.go
```diff
@@ -47,13 +49,54 @@ func apply(cl client.Client, objs []runtime.RawExtension) error {
...
+		}
 	}
 	return nil
 }
 
+func (p Processor) waitUntilProjectExists(namespace string, t time.Duration) (projectv1.Project, error) {
+	var prj projectv1.Project
+	ticker := time.NewTicker(t / 5)
+	timeout := time.After(t)
+	defer ticker.Stop()
```
nurali-techie: minor - move it after line no - 75.

pkg/template/processor.go
xcoulon: done in 33dedaa

pkg/template/processor.go
xcoulon: done in 33dedaa

pkg/template/processor.go
xcoulon: well, keeping this inside the loop avoids code duplication (calling the `createOrUpdateObj` func and checking the error afterwards). 
Also, this whole processor could remain generic, and it's probably the responsibility of the caller to verify that the input template has a ProjectRequest. WDYT?

pkg/template/processor.go
xcoulon: ok, I suggest that we use a higher, hard-coded value for now (10 sec?) and work on making this configurable in a subsequent PR. Ok for you?

xcoulon: value set to 10s in 92a2994

pkg/template/processor.go
nurali-techie: For code duplication,
We can arrange code for least code duplication and it should be fine IMO.


For ProjectRequest check, 
I understand that, as a Template processor, it should not bother if `ProjectRequest` is present or mandatory .. but Template processor is only used by our product as of now then I think we can have this check somehow within Template processor else we need to first load this template one more time with `serializer.CodecFactory` and need to check.

xcoulon: > For code duplication,
> We can arrange code for least code duplication and it should be fine IMO.
> 

The objects to create are sorted, so that the ProjectRequest will come first. Honestly, I don't see any benefit in processing the ProjectRequest separately from the other objects. The cost of checking if the object to create or update is a ProjectRequest is negligible.


> For ProjectRequest check,
> I understand that, as a Template processor, it should not bother if `ProjectRequest` is present or mandatory .. but Template processor is only used by our product as of now then I think we can have this check somehow within Template processor else we need to first load this template one more time with `serializer.CodecFactory` and need to check.

I'm not even sure what will be the benefit of checking if there's a ProjectRequest in the template. My point is: is there a chance that we will use this function without having a ProjectRequest object? Also, to avoid loading the template a second time, we could probably split the `ProcessAndApply` func to `Process` + `Apply`, so the caller can get the list of objects to create from the `Process` func and do some verifications before calling the `Apply` func.

pkg/template/processor.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		return nil
+	}
+	return nil
+}
+
+// Process processes the template with the given content
+func (p Processor) Process(tmplContent []byte, values map[string]string) ([]runtime.RawExtension, error) {
+	decode := p.codecFactory.UniversalDeserializer().Decode
+	obj, _, err := decode(tmplContent, nil, nil)
```
tinakurian: can this just be `obj, _, err :=p.codecFactory.UniversalDeserializer().Decode(tmplContent, nil, nil)`?

pkg/template/processor.go
```diff
@@ -0,0 +1,156 @@
...
+	"math/rand"
+	"sort"
+	"time"
+
+	projectv1 "github.com/openshift/api/project/v1"
+	templatev1 "github.com/openshift/api/template/v1"
+	"github.com/openshift/library-go/pkg/template/generator"
+	"github.com/openshift/library-go/pkg/template/templateprocessing"
+	errs "github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
```
tinakurian: should we add a space to separate and organize our imports?

pkg/template/processor.go
```diff
@@ -0,0 +1,156 @@
...
+	cl           client.Client
+	scheme       *runtime.Scheme
+	codecFactory serializer.CodecFactory
+}
+
+func NewProcessor(cl client.Client, scheme *runtime.Scheme) *Processor {
+	return &Processor{cl: cl, scheme: scheme, codecFactory: serializer.NewCodecFactory(scheme)}
+}
+
+func (p Processor) ProcessAndApply(tmplContent []byte, values map[string]string) error {
```
tinakurian: should we add some documentation on our public funcs?

pkg/template/processor.go
```diff
@@ -0,0 +1,156 @@
...
+	for _, rawObj := range objs {
+		obj := rawObj.Object
+		if obj == nil {
+			continue
+		}
+		gvk := obj.GetObjectKind().GroupVersionKind()
+		if err := createOrUpdateObj(p.cl, obj); err != nil {
+			return errs.Wrapf(err, "unable to create resource of kind: %s, version: %s", gvk.Kind, gvk.Version)
+		}
+		if gvk.Group == "project.openshift.io" && gvk.Version == "v1" && gvk.Kind == "ProjectRequest" {
```
tinakurian: can "ProjectRequest" use the const defined in sort.go?

pkg/template/processor.go
```diff
@@ -0,0 +1,156 @@
...
+			var prq projectv1.ProjectRequest
+			err := p.scheme.Convert(obj, &prq, nil)
+			if err != nil {
+				return errs.Wrapf(err, "unable to convert object of kind: %s, version: %s to a ProjectRequests", gvk.Kind, gvk.Version)
+			}
+			// wait until the Project exists and is ready
+			_, err = p.waitUntilProjectExists(prq.GetName(), 10*time.Second)
```
alexeykazakov: I'm a bit concerned about the fact that we wait 10 seconds for the project to be created and active. What if it takes longer for whatever reason? It would require proper alignment with the NSTemplateSet controller logic which is to use this template processor.
But TBH I don't know how we could do better without introducing too much of complexity.

pkg/template/processor.go
xcoulon: sure, let me change that

pkg/template/processor.go
xcoulon: not sure what you mean. All stdlib imports are at the top and there's a blank line to separate the "vendor" libs:
```

import (
	"context"
	"fmt"
	"math/rand"
	"sort"
	"time"

	projectv1 "github.com/openshift/api/project/v1"
	templatev1 "github.com/openshift/api/template/v1"
	"github.com/openshift/library-go/pkg/template/generator"
	"github.com/openshift/library-go/pkg/template/templateprocessing"
	errs "github.com/pkg/errors"
	corev1 "k8s.io/api/core/v1"
	apierrors "k8s.io/apimachinery/pkg/api/errors"
	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/apimachinery/pkg/runtime/serializer"
	"k8s.io/apimachinery/pkg/types"
	"sigs.k8s.io/controller-runtime/pkg/client"
)
```
if I need to import our own packages, I'll do it in-between the stdlib and the vendor pkgs, with a blankline to separate each group.

pkg/template/processor.go
xcoulon: sure, let me do that

xcoulon: done in e4cc320

pkg/template/processor.go
xcoulon: done in e4cc320

pkg/template/processor.go
xcoulon: in e4cc320 I've added a param to the `Apply` func to configure the project creation timeout, so it's up to the controller to decide how long to wait. I suggest we externalize this setting in the deployment environment (https://github.com/codeready-toolchain/member-operator/blob/master/deploy/operator.yaml#L23), but let's do that in a separate PR, ok?

xcoulon:
> We can also improve tests. There is a few cases not covered (mostly error handling).

ok, let me check that and add more tests ;)

xcoulon:
> > We can also improve tests. There is a few cases not covered (mostly error handling).
> 
> ok, let me check that and add more tests ;)

@alexeykazakov I added a test on an error in a4e8a79

pkg/template/processor.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+}
+
+func createOrUpdateObj(cl client.Client, obj runtime.Object) error {
+	if err := cl.Create(context.TODO(), obj); err != nil {
+		if !apierrors.IsAlreadyExists(err) {
+			return errs.Wrapf(err, "failed to create object %v", obj)
+		}
+
+		if err = cl.Update(context.TODO(), obj); err != nil {
```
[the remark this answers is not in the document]
xcoulon: I believe this is done by the client in https://github.com/kubernetes/client-go/blob/master/tools/cache/thread_safe_store.go#L76

pkg/template/processor.go
```diff
@@ -0,0 +1,136 @@
...
+	// sorting before applying to maintain correct order
+	sort.Sort(ByKind(objs))
+
+	for _, rawObj := range objs {
+		obj := rawObj.Object
+		if obj == nil {
+			continue
+		}
+		gvk := obj.GetObjectKind().GroupVersionKind()
+		if err := createOrUpdateObj(p.cl, obj); err != nil {
```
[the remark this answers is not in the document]
xcoulon: I'm not sure if you're asking for a change here :/

pkg/template/processor.go
xcoulon: @nurali-techie I slightly refactored the code and there's no `ProcessAndApply` func anymore. You'll need to call `Process` first and get the `objects` to create (or an error), then pass those objects to the `Apply` func. So you can add some validation logic in between, to ensure that there's a `ProjectRequest` at the first position.

pkg/template/processor.go
alexeykazakov: Sure. Having a better way of controlling timeouts and delays is useful but it doesn't eliminate the potential problem with too long namespace/project creation completely. We would need to drop our own custom waits and relay on controller reconcile events for that. But as I said it would probably bring a lot of complexity in our template process flow.
So, I'm fine with using that wait for now but wanted to share my overall concern with such approach :)

pkg/template/processor.go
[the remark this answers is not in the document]
alexeykazakov: Yeah.. the client seems to handle object updates properly if they have been modified using the same client. But if they were modified by another client then there would be a conflict (which is not the case here).

pkg/template/processor.go
tinakurian: i was thinking to take it one step further and separate based on orgs for the vendor pkgs. so for instance separating openshift from k8s. but it's minor and i'm unsure if thats normal in the go world...i think i was shown a document on this being the best practice - but i can't find the doc :-) either way is fine.

pkg/template/processor.go
```diff
@@ -0,0 +1,119 @@
...
+	for _, obj := range objs {
+		if filter(obj) {
+			result = append(result, obj)
+		}
+	}
+	return result
+}
+
+// Apply applies the objects, ie, creates or updates them on the cluster
+func (p Processor) Apply(objs []runtime.RawExtension, projectActiveTimeout time.Duration) error {
```
nurali-techie: `projectActiveTimeout` no more needed ??

xcoulon: good point, I missed that. Let me see if I can configure the linter to reveal that kind of error.

pkg/template/processor.go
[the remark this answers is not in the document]
MatousJobanek: Yes, I was asking for a change, but I guess that with the new flow the caller of this function (the controller) should be responsible for checking if the namespace already exists or not. If it does, then it should not call this function and just fail...

pkg/template/processor_test.go
MatousJobanek: Since the function doesn't take any parameter, then I don't see any reason why we should have it. We can easily generate the values as global variables and use them in the tests. I know that it generates the name with the timestamp, but is it really needed?

pkg/template/processor_test.go
```diff
@@ -0,0 +1,604 @@
...
+	return fmt.Sprintf("%s-%d", prefix, time.Now().UnixNano())
+}
+
+func verifyResource(t *testing.T, objKind schema.ObjectKind, vars ...string) {
+	require.IsType(t, &unstructured.Unstructured{}, objKind)
+	obj := objKind.(*unstructured.Unstructured)
+	objJson, err := obj.MarshalJSON()
+	require.NoError(t, err, "failed to marshal json from unstructured object")
+	for _, v := range vars {
+		assert.Contains(t, string(objJson), v, fmt.Sprintf("missing %s", v))
```
MatousJobanek: Shouldn't we verify that the variables are written at the correct places? Here it "only" verifies that the whole object contains the variable values, but doesn't verify that are used/set correctly.
Apart from that, maybe it would be good that all variables are replaced - in other words, that the object doesn't contain any string matching `${...}`

pkg/template/processor_test.go
```diff
@@ -0,0 +1,604 @@
...
+		require.Len(t, objs, 2)
+
+		// namespace
+		verifyResource(t, objs[0].Object.GetObjectKind(), namespace, commit, user)
+
+		// role binding
+		verifyResource(t, objs[1].Object.GetObjectKind(), namespace, user)
+	})
+
+	t.Run("should process template with missing required params", func(t *testing.T) {
```
MatousJobanek:
```suggestion
	t.Run("process template should fail because of missing required parameter", func(t *testing.T) {
```

pkg/template/processor_test.go
```diff
@@ -0,0 +1,604 @@
...
+
+func TestProcess(t *testing.T) {
+	t.Run("should process template successfully", func(t *testing.T) {
+		// given
+		s := addToScheme(t)
+		namespace, commit, user := templateVars()
+		values := paramsKeyValues(namespace, commit, user)
+
+		cl := test.NewFakeClient(t)
+		p := template.NewProcessor(cl, s)
```
MatousJobanek: I believe that the client, processor, and scheme can be declared out of the subtests as part of the test setup of the parent test (lines after 80) and then reuse it in all subtests

pkg/template/processor_test.go
```diff
@@ -0,0 +1,604 @@
...
+				"name": "rb2",
+			},
+		},
+	}
+
+	t.Run("filter namespaces", func(t *testing.T) {
+
+		t.Run("return one", func(t *testing.T) {
+			// given
+			objs := []runtime.RawExtension{
```
MatousJobanek: for me, the short version:
```go
objs := []runtime.RawExtension{{Object: ns1}, {Object: rb1}}
```
if more readable 
WDYT?

pkg/template/processor_test.go
```diff
@@ -0,0 +1,604 @@
...
+		})
+	})
+}
+
+func TestProcessAndApply(t *testing.T) {
+
+	t.Run("should create namespace alone", func(t *testing.T) {
+
+		// given
+		s := addToScheme(t)
```
MatousJobanek: In this case we can move at least this line out of subtests - to line 371

pkg/template/processor_test.go
```diff
@@ -0,0 +1,604 @@
...
+		namespace, commit, user := templateVars()
+		values := paramsKeyValues(namespace, commit, user)
+		cl := test.NewFakeClient(t)
+		p := template.NewProcessor(cl, s)
+
+		// when
+		objs, err := p.Process(templateContent(namespaceObj), values)
+		require.NoError(t, err)
```
MatousJobanek: it's a detail, but these two lines should belong to "given" part - the test verifies only the "apply" functionality... 
but as I said - it's a detail ;-)

pkg/template/processor_test.go
```diff
@@ -0,0 +1,604 @@
...
+			require.NoError(t, err)
+			verifyNamespace(t, cl, namespace)
+			verifyRoleBinding(t, cl, namespace)
+		})
+
+		t.Run("failure", func(t *testing.T) {
+
+			t.Run("client error", func(t *testing.T) {
+
+				t.Run("namespace does not exist", func(t *testing.T) {
```
MatousJobanek: may I know why so many subtests and just not a longer subtest name such as:
```go
t.Run("apply should fail because of missing namespace", func(t *testing.T) {
```

pkg/template/processor_test.go
```diff
@@ -0,0 +1,604 @@
...
+		require.NoError(t, err)
+		err = p.Apply(objs, 2*time.Second)
+
+		require.NoError(t, err)
+		verifyRoleBinding(t, cl, namespace)
+
+		// when
+		objs, err = p.Process(templateContent(roleBindingObj, newUser), values)
+		require.NoError(t, err)
+		err = p.Apply(objs, 2*time.Second)
```
MatousJobanek: correct me if I'm wrong - the RoleBinding that is processed/applied is the very same as the previous one, so there is no way you could verify that the object was really updated. You should change the object and then verify that the object was really changed.

pkg/template/processor_test.go
```diff
@@ -0,0 +1,604 @@
...
+					cl.MockCreate = func(ctx context.Context, obj runtime.Object) error {
+						if obj.GetObjectKind().GroupVersionKind().Kind == "RoleBinding" {
+							return errors.New("mock error")
+						}
+						return nil
+					}
+					p := template.NewProcessor(cl, s)
+
+					// when
+					objs, err := p.Process(templateContent(namespaceObj, roleBindingObj), values)
```
MatousJobanek: I'm a bit confused - the test name contains "namespace does not exist" but here I see that it creates the namespace

pkg/template/processor_test.go
```diff
@@ -0,0 +1,604 @@
...
+		require.NoError(t, err)
+		err = p.Apply(objs, 2*time.Second)
+
+		// then
+		require.NoError(t, err)
+		binding := verifyRoleBinding(t, cl, namespace)
+		require.Len(t, binding.Subjects, 2)
+	})
+
+	t.Run("should fail to create template object", func(t *testing.T) {
```
MatousJobanek: this test looks very similar to this one: https://github.com/codeready-toolchain/member-operator/pull/49/files#diff-a5ef365842f1e1803e47926eaf743087R435

pkg/template/processor.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	// sorting before applying to maintain correct order
+	objs := result.Objects
+	return objs, nil
+}
+
+// FilterFunc a function to retain an object or not
+type FilterFunc func(runtime.RawExtension) bool
+
+var RetainNamespaces, RetainAllButNamespaces FilterFunc
```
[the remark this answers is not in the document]
xcoulon: yeah, I've seen it too, I'll fix it in my next commit (waiting for the other #62 to be merged, first)

go.mod
```diff
@@ -54,3 +57,8 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20180711000925-0cf8f7e6ed1d
 	k8s.io/kubernetes => k8s.io/kubernetes v1.13.4
 )
+
+replace (
+	git.apache.org/thrift.git => github.com/apache/thrift v0.12.0
```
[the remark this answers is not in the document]
xcoulon: yes, I'll do that when merging #62

pkg/template/processor.go
```diff
@@ -0,0 +1,119 @@
...
+	if err := tmplProcessor.Process(tmpl); len(err) > 0 {
+		return nil, errs.Wrap(err.ToAggregate(), "unable to process template")
+	}
+	var result templatev1.Template
+	if err := p.scheme.Convert(tmpl, &result, nil); err != nil {
+		return nil, errs.Wrap(err, "failed to convert template to external template object")
+	}
+	// sorting before applying to maintain correct order
+	objs := result.Objects
+	return objs, nil
```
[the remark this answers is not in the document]
xcoulon: good point

pkg/template/processor.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// FilterFunc a function to retain an object or not
+type FilterFunc func(runtime.RawExtension) bool
+
+var RetainNamespaces, RetainAllButNamespaces FilterFunc
+
+func init() {
```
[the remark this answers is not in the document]
xcoulon: good point!

pkg/template/processor.go
```diff
@@ -0,0 +1,119 @@
...
+	codecFactory serializer.CodecFactory
+}
+
+// NewProcessor returns a new Processor
+func NewProcessor(cl client.Client, scheme *runtime.Scheme) Processor {
+	return Processor{cl: cl, scheme: scheme, codecFactory: serializer.NewCodecFactory(scheme)}
+}
+
+// Process processes the template (ie, replaces the variables with their actual values)
+func (p Processor) Process(tmplContent []byte, values map[string]string) ([]runtime.RawExtension, error) {
```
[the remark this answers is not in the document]
xcoulon: good point, I can add one or more filters as a trailing argument (making this func variadic)

pkg/template/processor_test.go
xcoulon: I prefer using separate lines, but I guess it's a matter of taste

MatousJobanek: Then keep it as it is ;-),

pkg/template/processor.go
xcoulon: removed in 091cb51

pkg/template/processor.go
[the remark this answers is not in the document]
xcoulon: done in 236feec (and moved in https://github.com/codeready-toolchain/member-operator/pull/49/files#diff-6b3f1e6f1c214f326255255b765a0d1bR7-R16)

pkg/template/processor.go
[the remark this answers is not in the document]
xcoulon: changed the logic to use optional filters before returning the objects.

pkg/template/processor.go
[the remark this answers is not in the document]
xcoulon: done in https://github.com/codeready-toolchain/member-operator/pull/49/files#diff-6b3f1e6f1c214f326255255b765a0d1bR7-R16

pkg/template/processor.go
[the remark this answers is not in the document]
xcoulon: done in 236feec

pkg/template/processor_test.go
xcoulon: good point. I'll refactor the test to compare the whole output, instead

pkg/template/processor_test.go
xcoulon: I'm sorry, the link brings me to other review comments. Could you please give me the title of the test that looks similar to `should fail to create template object` ?

pkg/template/processor.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	var result templatev1.Template
+	if err := p.scheme.Convert(tmpl, &result, nil); err != nil {
+		return nil, errs.Wrap(err, "failed to convert template to external template object")
+	}
+	return Filter(result.Objects, filters...), nil
+}
+
+// Apply applies the objects, ie, creates or updates them on the cluster
+func (p Processor) Apply(objs []runtime.RawExtension) error {
```
alexeykazakov: We should make sure that the controller can manipulate the objects in the controller before `Apply` will create them in API server.
It will be needed for setting the owner reference (and maybe additional labels/annotations) for Namespaces. See https://docs.google.com/document/d/19UzaXVqiOo9JqDYz3qUKguJwW-7zceA4UtTr358R_CA/edit?disco=AAAADp6G_R4

pkg/template/processor_test.go
xcoulon: done in 39a07f3

pkg/template/processor_test.go
xcoulon: done in a693813

pkg/template/processor_test.go
```diff
@@ -0,0 +1,711 @@
...
+	},
+	"subjects": [
+		{
+			"kind": "User",
+			"name": "{{ .Username }}"
+		}
+	]
+}`
+
+	rolebindingWithExtraUserObj = `{
```
golangcibot: `rolebindingWithExtraUserObj` is unused (from `varcheck`)

pkg/template/processor_test.go
```diff
@@ -0,0 +1,711 @@
...
```
golangcibot: `rolebindingWithExtraUserObj` is unused (from `deadcode`)

pkg/template/processor_test.go
xcoulon: done in 16e7265 for the `TestProcessor` func only (when testing the apply, we need to reset the client after each test, so it's simpler to declare a new one each time)

pkg/template/processor_test.go
xcoulon: ok ;)

pkg/template/processor_test.go
xcoulon: done in 16e7265

pkg/template/processor_test.go
xcoulon: fair point, addressed in 718d67c

pkg/template/processor_test.go
xcoulon: done (lost track of the commit, though :/ )

pkg/template/processor_test.go
xcoulon: good point, let me do that

pkg/template/processor_test.go
```diff
@@ -0,0 +1,681 @@
...
```
golangcibot: `rolebindingWithExtraUserObj` is unused (from `varcheck`)

pkg/template/processor_test.go
xcoulon: actually, it was addressed in a previous commit: the updated rolebinding has an extra user:
```
objs, err = p.Process([]byte(namespaceAndRolebindingWithExtraUserTmpl), values)
require.NoError(t, err)
err = p.Apply(objs)
```

pkg/template/processor_test.go
xcoulon: I believe this is now outdated :/

pkg/template/processor_test.go
xcoulon: fixed in 612cbde

pkg/template/processor_test.go
xcoulon: fixed in 612cbde

pkg/template/processor_test.go
xcoulon: fixed in 612cbde

pkg/template/processor.go
xcoulon: done in 6480e4c

pkg/template/processor.go
```diff
@@ -77,6 +77,7 @@ func (p Processor) Apply(objs []runtime.RawExtension) error {
 }
 
 func createOrUpdateObj(cl client.Client, obj runtime.Object) error {
+	fmt.Printf("creating or updating %v\n", obj)
```
alexeykazakov: Remove Printf. You can log it properly if needed though :)

xcoulon: doh! (and done in 198b954)

alexeykazakov:
BTW. Great test coverage @xcoulon !

xcoulon:
> BTW. Great test coverage @xcoulon !

thanks @alexeykazakov ! (and kudos to @dipak-pawar as well for the initial effort)

pkg/template/processor_test.go
xcoulon: my bad. Done in 18dcc52

PR #55, opened 2019-08-22
Update k8s to 1.13.4

alexeykazakov:
This PR updates and pins k8s to v.1.13.4 which is currently used by operator-sdk v.0.1.10


alexeykazakov:
Yes. I have already done go mod tidy. I had to re-add Azure/autorest thing back to required though to fix the build. The rest should be clean.

PR #62, opened 2019-09-10
check unused params with golangci

xcoulon:
Unable the `unparam` linter on golangci and modify the code accordingly. 
For the tests, just use the `// nolint:unparam` comment to avoid reporting a problem
 

xcoulon:
also, making some changes in some function's returned values by moving the `error` at the last position (see https://github.com/codeready-toolchain/member-operator/pull/62/files#diff-25293d17133b9d89712a15c8e56795a3R276)

pkg/controller/useraccountstatus/useraccount_status_controller_test.go
```diff
@@ -207,7 +208,7 @@ func newUserAccount(userName, resourceVersion string) *toolchainv1alpha1.UserAcc
 	return userAcc
 }
 
-func newMasterUserRecord(userName, syncIndex string) *toolchainv1alpha1.MasterUserRecord {
+func newMasterUserRecord(userName, syncIndex string) *toolchainv1alpha1.MasterUserRecord { //nolint:unparam
```
MatousJobanek: may I know why do we ignore linter here?

xcoulon: because the linter says that it's always the same value that is used when this func is called, but since it's in a test, I guess it's ok and we can ignore the problem. I wanted to focus on the "production code" errors in this PR.

xcoulon:
> Looks good. I have only one question.
> Not sure why the analyzis test failed :-/ https://golangci.com/r/github.com/codeready-toolchain/member-operator/pulls/62

yeah, that's what I'm trying to understand as well :/

xcoulon:
@MatousJobanek I was able to fix the problem with golang-ci (see https://github.com/codeready-toolchain/member-operator/pull/62/commits/2bf8a304910a9362a1d67d61146ceec7b32fb3fc)

pkg/controller/useraccountstatus/useraccount_status_controller_test.go
MatousJobanek: ah, sure. because it's been always called with the userName = "johny" so it says that the param is not necessary. 
I understand that you want to focus on the "production code" but I wouldn't consider the tests as something less important. At the end of the day, we could set the linters to ignore all test files, but this is something that we really don't want to do I believe. 
My point is that it looks weird that it has the ignore clause when there is no problem with the function. I would rather change one of the callers instead of ignoring it.

xcoulon: ok, fair enough, I'll update the test func, then. Or use another value in some tests 😉

xcoulon:
thanks for the review @MatousJobanek

PR #63, opened 2019-09-17
add func to retrieve templates from NSTemplateTier on host

xcoulon:
returns a map of revisioned templates

fixes CRT-249


pkg/apis/apis.go
```diff
@@ -11,6 +13,7 @@ import (
 
 // AddToScheme adds all Resources to the Scheme
 func AddToScheme(s *runtime.Scheme) error {
+	fmt.Println("adding OpenShift specific resources...")
```
MatousJobanek: zombie :-)

pkg/template/nstemplatetier.go
```diff
@@ -0,0 +1,47 @@
...
+// RevisionedTemplate a template along with its revision number
+type RevisionedTemplate struct {
+	Revision string
+	Template string
+}
+
+// GetNSTemplates gets the templates configured in the NSTemplateTier resource
+// which is fetched from the host cluster.
+func GetNSTemplates(tierName string) (NSTemplates, error) {
+	// retrieve the address of the host cluster in the
```
MatousJobanek:
```suggestion
	// retrieve the FedCluster instance representing the host cluster
```

pkg/template/nstemplatetier.go
```diff
@@ -0,0 +1,47 @@
...
+	host, ok := cluster.GetHostCluster()
+	if !ok {
+		return nil, fmt.Errorf("unable to connect to the Host cluster: unknown cluster")
+	}
```
MatousJobanek: I think that you should also verify that the cluster is ready - see here: 
https://github.com/codeready-toolchain/member-operator/blob/master/pkg/controller/useraccountstatus/useraccount_status_controller.go#L101-L106

pkg/template/nstemplatetier_test.go
```diff
@@ -0,0 +1,214 @@
...
+	})
+
+}
+
+func newHostCluster(name string) (*v1beta1.KubeFedCluster, *corev1.Secret) {
+	// make sure that the client gets a valid 200 OK response
+	// when trying to ping the `/api` endpoint on the fake host cluster
+	gock.Observe(gock.DumpRequest)
+
+	gock.New("https://" + name).
```
MatousJobanek: May I ask you why did you choose this approach of using gock mocks instead of the fake client?

pkg/template/nstemplatetier_test.go
```diff
@@ -0,0 +1,214 @@
...
+			Kind:       "NSTemplateTier",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "basic",
+			Namespace: "toolchain-host-operator",
+		},
+	}
+	cl := fake.NewFakeClient(secret, basicTier)
+	service := cluster.KubeFedClusterService{Log: logf.Log, Client: cl}
+	service.AddKubeFedCluster(kubeFedCluster)
```
MatousJobanek: There is no cleanup. Couldn't that affect other tests?

pkg/template/nstemplatetier_test.go
```diff
@@ -0,0 +1,214 @@
...
+		Reply(200).
+		BodyString(`{
+			"apiVersion": "toolchain.dev.openshift.com/v1alpha1",
+			"kind": "NSTemplateTier",
+			"metadata": {
+				"name": "basic",
+				"namespace": "host-operator-1568102314"
+			},
+			"spec": {
+				"namespaces": [
```
MatousJobanek: The biggest drawback of this approach that I see is that it will be hard to maintain in case we change anything in the structure of the `NSTemplateTier` type

pkg/template/nstemplatetier_test.go
```diff
@@ -0,0 +1,214 @@
...
+	service := cluster.KubeFedClusterService{Log: logf.Log, Client: cl}
+	service.AddKubeFedCluster(kubeFedCluster)
+
+	t.Run("success", func(t *testing.T) {
+		// when
+		tmpls, err := template.GetNSTemplates("basic")
+
+		// then
+		require.NoError(t, err)
+		require.Len(t, tmpls, 3)
```
MatousJobanek: could we also verify that the returned templates contain the correct content?

pkg/apis/apis.go
xcoulon: 2hrs ago, I was thinking about having a linter for such statements! 😬

pkg/template/nstemplatetier_test.go
xcoulon: sure, let me add that

pkg/template/nstemplatetier_test.go
```diff
@@ -0,0 +1,95 @@
...
+)
+
+func TestGetNSTemplateTier(t *testing.T) {
+
+	// given
+	// Setup Scheme for all resources (required before adding objects in the fake client)
+	err := apis.AddToScheme(scheme.Scheme)
+	require.NoError(t, err)
+	logf.SetLogger(zap.Logger())
+	basicTier := &toolchainv1alpha1.NSTemplateTier{
```
alexeykazakov: Let's have multiple tiers in the client available. Not the only one. And the NS types should use the same names in different tiers.
Like `basicTier` with `stage`, `cicd`, `ide`. And `advanced` with its own `stage`, `cicd`, `ide`. So we can make sure that the templates loaded from the correct tier.

pkg/apis/apis.go
xcoulon: removed in 21bb803

pkg/template/nstemplatetier.go
xcoulon: thanks for the pointer! I added that check in edc3e2f

xcoulon: removed in https://github.com/codeready-toolchain/member-operator/commit/21bb803d4a1d43875aabaf0f26d516f94a817ad2 after our fruitful discussion ;)

xcoulon: (responded to wrong comment)

pkg/template/nstemplatetier_test.go
xcoulon: removed in https://github.com/codeready-toolchain/member-operator/commit/21bb803d4a1d43875aabaf0f26d516f94a817ad2 after our fruitful discussion ;)

pkg/template/nstemplatetier_test.go
xcoulon: outdated since it was removed in https://github.com/codeready-toolchain/member-operator/commit/21bb803d4a1d43875aabaf0f26d516f94a817ad2

pkg/template/nstemplatetier_test.go
xcoulon: outdated since it was removed in https://github.com/codeready-toolchain/member-operator/commit/21bb803d4a1d43875aabaf0f26d516f94a817ad2

pkg/template/nstemplatetier_test.go
xcoulon: done in 888297c

pkg/template/nstemplatetier_test.go
xcoulon: ok, let me do that

xcoulon: done in 66d2d6d

pkg/template/nstemplatetier_test.go
```diff
@@ -0,0 +1,169 @@
...
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "toolchain.dev.openshift.com/v1alpha1",
+			Kind:       "NSTemplateTier",
+		},
```
MatousJobanek: I don't really think that we need to set the `APIVersion` and `Kind` as you are using the `NSTemplateTier` object directly, so it should have been set automatically.

pkg/template/nstemplatetier.go
```diff
@@ -0,0 +1,52 @@
...
+
+// RevisionedTemplate a template along with its revision number
+type RevisionedTemplate struct {
+	Revision string
+	Template string
+}
+
+// GetNSTemplates gets the templates configured in the NSTemplateTier resource
+// which is fetched from the host cluster.
+func GetNSTemplates(hostClusterFunc cluster.GetHostClusterFunc, tierName string) (NSTemplates, error) {
```
MatousJobanek: the `hostClusterFunc` could have been a struct param, but this is fine as well.

pkg/template/nstemplatetier_test.go
```diff
@@ -0,0 +1,169 @@
...
+			require.Error(t, err)
+			assert.Contains(t, err.Error(), "unable to retrieve the NSTemplateTier 'unknown' from 'Host' cluster")
+		})
+	})
+
+}
+
+func newHostCluster(cl client.Client, condition fedv1b1.ClusterCondition) cluster.GetHostClusterFunc {
+	return func() (*cluster.FedCluster, bool) {
+		return &cluster.FedCluster{
```
MatousJobanek: You should have also set the `OperatorNamespace`, otherwise, when you make the call https://github.com/codeready-toolchain/member-operator/pull/63/files#diff-88f2951b69e27daa0728c455506a942aR38 the client will look for it cluster-wide, but you should look for it only inside of the namespace

pkg/template/nstemplatetier_test.go
xcoulon: not sure if those values will be set automatically, but I can remove them from my test, indeed

PR #66, opened 2019-09-25
chore: clone the whole toolchain-e2e repo to make merging easier

MatousJobanek:
We need to clone the whole history so we can merge PR with unrelated history into master


make/test.mk
```diff
@@ -94,7 +94,7 @@ ifeq ($(E2E_REPO_PATH),"")
 	# delete to have clear environment
 	rm -rf ${E2E_REPO_PATH}
 	# clone
-	git clone https://github.com/codeready-toolchain/toolchain-e2e.git --depth 1 ${E2E_REPO_PATH}
+	git clone https://github.com/codeready-toolchain/toolchain-e2e.git ${E2E_REPO_PATH}
```
dipak-pawar: Just wondering, instead of taking a whole clone, can we fetch required commits of master till the point where we can merge and There might not be a ref that we're looking for.
```
while [ -z $( git merge-base master feature ) ]; do     
    git fetch -q --deepen=1 origin master feature;
done
```

we can start from `--deepgen=10`, some big number. if it's saving some time for us, then maybe worth trying this wdyt @MatousJobanek

MatousJobanek: I was thinking about it too, but for the sake of keeping it simple (the logic is complex enough), I decided to clone the whole master... The reason was that I thought that the clone part is just a minor part of the whole e2e execution and doesn't take to much time...
anyway, as you are already asking to do that, let me try it ;-)

MatousJobanek: @dipak-pawar check this 3cba5cc

alexeykazakov: It seems to get too complicated :) I would rather stick with the simplest clone/fetch the whole history for the sake of simplicity.

MatousJobanek: Yeah, I figured out that maintaining the lazy fetch and the loop is really painful and very complicates the code. I'm gonna stick with the simple version.

PR #68, opened 2019-09-26
refactor processor after NSTemplateTier changes

xcoulon:
No need to decode a `[]byte` into a `templatev1.Template` object in the `Process` func since the `GetNSTemplateTier` func will now return a `templatev1.Template` object.

Also, refactor some test templates to remove the unnecessary `PROJECT_NAME` parameter (won't exist in production templates)

See CRT-273


go.mod
```diff
@@ -1,43 +1,81 @@
 module github.com/codeready-toolchain/member-operator
 
 require (
```
MatousJobanek: run `go mod tidy` please

pkg/template/processor_test.go
```diff
@@ -497,6 +482,20 @@ func assertRoleBindingExists(t *testing.T, c client.Client, ns string) authv1.Ro
 	return rb
 }
 
+type decodeFunc func([]byte, *schema.GroupVersionKind, runtime.Object) (runtime.Object, *schema.GroupVersionKind, error)
```
MatousJobanek: this is not used anywhere

go.mod
xcoulon: done in e165b0b

pkg/template/processor_test.go
xcoulon: ah right, that's a left-over :/ I removed it in e165b0b

xcoulon:
thanks for the review @alexeykazakov 🙌

xcoulon:
thanks for the review @nurali-techie !

PR #70, opened 2019-10-02
Fix make deploy-crd

alexeykazakov:


alexeykazakov:
> Looks good, but I have the same comment as for the host-operator repo: wouldn't be better to run oc apply -f deploy/crds/ (for the whole directory) so we avoid such problems when we rename the CRD again?

Done in 0bcca0703d4291970ecd92da667b85901f45066e

PR #73, opened 2019-10-08
add commit and buildtime vars

xcoulon:
fixes CRT-282

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -14,7 +14,7 @@ import (
 	"github.com/codeready-toolchain/member-operator/pkg/controller"
 
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
-	"github.com/operator-framework/operator-sdk/pkg/kube-metrics"
+	kubemetrics "github.com/operator-framework/operator-sdk/pkg/kube-metrics"
 	"github.com/operator-framework/operator-sdk/pkg/leader"
 	"github.com/operator-framework/operator-sdk/pkg/log/zap"
 	"github.com/operator-framework/operator-sdk/pkg/metrics"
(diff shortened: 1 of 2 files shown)


xcoulon:
> Looks good. But we can add at least a Unit test for it, right? e2e tests would be trickier...

yes, I can add a test to verify that these vars exist so we don't remove them by mistake, but I can't assert their values (or well, it wouldn't help much IMO)

alexeykazakov:
> yes, I can add a test to verify that these vars exist so we don't remove them by mistake, but I can't assert their values (or well, it wouldn't help much IMO)

For values you could check the format of the build date. So, it's a date and not some random string :)

xcoulon:
> > yes, I can add a test to verify that these vars exist so we don't remove them by mistake, but I can't assert their values (or well, it wouldn't help much IMO)
> 
> For values you could check the format of the build date. So, it's a date and not some random string :)

I don't think so: the values are set by the `-ldflags="-X ..."` arg during the `go build`, not the `go test`, so the values will still be `unknown` at this point.

alexeykazakov:
> I don't think so: the values are set by the -ldflags="-X ..." arg during the go build, not the go test, so the values will still be unknown at this point.

Ah, right. I keep forgetting it. So, we can only test the presents of the fields itself.

xcoulon:
> > I don't think so: the values are set by the -ldflags="-X ..." arg during the go build, not the go test, so the values will still be unknown at this point.
> 
> Ah, right. I keep forgetting it. So, we can only test the presents of the fields itself.

no worries ;) I added a (very basic) test in https://github.com/codeready-toolchain/member-operator/pull/73/commits/d57100670e2cfe61eea71361b1d32fbf856c21e3 so we can at least assert that the 2 vars exist.

PR #83, opened 2019-10-16
update Commit and BuildTime ldflags to new pkg path

xcoulon:
forgot to do that in 34aca69 :/

Fixes CRT-282


xcoulon:
Needed to update the path to the new pkg: https://github.com/codeready-toolchain/member-operator/blob/master/version/version.go#L10-L12

PR #95, opened 2019-11-14
set resourceversion before updating

xcoulon:
if the object to update can be cast to
'unstructured.Unstructured`, then use the client
to get the existing version before replacing it with
the given one.

Fixes CRT-288

### DIFF
--- a/pkg/template/processor_test.go
+++ b/pkg/template/processor_test.go
@@ -287,6 +287,22 @@ func TestProcessAndApply(t *testing.T) {
 	t.Run("should update existing role binding", func(t *testing.T) {
 		// given
 		cl := test.NewFakeClient(t)
+		cl.MockCreate = func(ctx context.Context, obj runtime.Object, opts ...client.CreateOption) error {
+			if rb, ok := obj.(*unstructured.Unstructured); ok {
+				rb.SetResourceVersion("1")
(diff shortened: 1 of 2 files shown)


pkg/template/processor_test.go
```diff
@@ -287,6 +287,22 @@ func TestProcessAndApply(t *testing.T) {
...
+				rb.SetResourceVersion("1")
+				return cl.Client.Create(ctx, obj, opts...)
+			}
+			return cl.Client.Create(ctx, obj, opts...)
+		}
+		cl.MockUpdate = func(ctx context.Context, obj runtime.Object, opts ...client.UpdateOption) error {
+			if u, ok := obj.(*unstructured.Unstructured); ok {
+				fmt.Printf("updating resource of kind %s with version %s\n", u.GetKind(), u.GetResourceVersion())
+				if u.GetKind() == "RoleBinding" && u.GetResourceVersion() != "1" {
+					return errors.New(fmt.Sprintf("invalid resource version: %q", u.GetResourceVersion()))
```
golangcibot: S1028: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (from `gosimple`)

xcoulon: fixed in da5f892

PR #104, opened 2019-12-11
Add Banned property to MasterUserRecord

sbryzak:
https://jira.coreos.com/browse/CRT-373


sbryzak:
That's what I thought, however it is the latest master from the api module that generated this code.

alexeykazakov:
Yeah. but API uses some source from member to generate stuff. I guess you need to update your member repo.

PR #105, opened 2019-12-12
Disabling useraccount

tinakurian:
Currently e2e-tests are not possible - but they will be added with changes made in the host-operator


pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -776,6 +776,76 @@ func TestUpdateStatus(t *testing.T) {
...
+	err := apis.AddToScheme(s)
+	require.NoError(t, err)
+
+	t.Run("disabled useraccount", func(t *testing.T) {
+		// given
+		userAcc := newUserAccount(username, userID)
+		r, req, _ := prepareReconcile(t, username, userAcc)
+
+		//when
+		res, err := r.Reconcile(req)
```
golangcibot: ineffectual assignment to `res` (from `ineffassign`)

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -786,6 +856,7 @@ func newUserAccount(userName, userID string) *toolchainv1alpha1.UserAccount {
 		Spec: toolchainv1alpha1.UserAccountSpec{
 			UserID:        userID,
 			NSTemplateSet: newNSTmplSetSpec(),
+			Disabled:false,
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
			Disabled:      false,
```

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -41,6 +41,7 @@ const (
 
 	// Finalizers
 	userAccFinalizerName = "finalizer.toolchain.dev.openshift.com"
+	userAccFinalizerDisabled = "finalizer.toolchain.dev.openshift.com.disabled"
```
golangcibot: `userAccFinalizerDisabled` is unused (from `varcheck`)

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -125,25 +125,44 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+			identityErr := r.client.Get(context.TODO(), types.NamespacedName{Name: name}, identity)
+			if errors.IsNotFound(identityErr) {
+				user := &userv1.User{}
+				userErr := r.client.Get(context.TODO(), types.NamespacedName{Name: userAcc.Name}, user)
+				if errors.IsNotFound(userErr) {
+					return reconcile.Result{}, r.setStatusDisabled(userAcc)
+				} else {
+					// log error and return it
+				}
+			} else {
```
golangcibot: SA9003: empty branch (from `staticcheck`)

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -125,25 +125,44 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+		if userAcc.Spec.Disabled {
+			name := ToIdentityName(userAcc.Spec.UserID)
+			identity := &userv1.Identity{}
+			identityErr := r.client.Get(context.TODO(), types.NamespacedName{Name: name}, identity)
+			if errors.IsNotFound(identityErr) {
+				user := &userv1.User{}
+				userErr := r.client.Get(context.TODO(), types.NamespacedName{Name: userAcc.Name}, user)
+				if errors.IsNotFound(userErr) {
+					return reconcile.Result{}, r.setStatusDisabled(userAcc)
+				} else {
```
golangcibot: SA9003: empty branch (from `staticcheck`)

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -125,28 +125,52 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
-	} else if util.HasFinalizer(userAcc, userAccFinalizerName) {
+	}
+	if util.HasFinalizer(userAcc, userAccFinalizerName) || userAcc.Spec.Disabled {
 		if err = r.manageCleanUp(userAcc); err != nil {
 			return reconcile.Result{}, err
 		}
+
+		if userAcc.Spec.Disabled {
+			disabled, err := r.setDisabledStatus(userAcc)
+			if disabled == true {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `disabled` (from `gosimple`)

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -125,28 +125,52 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
 		if _, createdOrUpdated, err = r.ensureIdentity(reqLogger, userAcc, user); err != nil || createdOrUpdated {
 			return reconcile.Result{}, err
 		}
-
 		if _, createdOrUpdated, err = r.ensureNSTemplateSet(reqLogger, userAcc); err != nil || createdOrUpdated {
 			return reconcile.Result{}, err
 		}
-	} else if util.HasFinalizer(userAcc, userAccFinalizerName) {
+	}
+	if util.HasFinalizer(userAcc, userAccFinalizerName) || userAcc.Spec.Disabled {
```
sbryzak: I think that the logic to handle the disabled case should go in a different `if` block, and that the status shouldn't be updated in the same reconcile loop as the identity and user deletions.  Once the identity and user resources are deleted, it will trigger another reconcile loop for the owning `UserAccount` resource, correct? If that's the case, then it should be this reconcile that updates the status.

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -125,28 +125,52 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+			} else {
+				return reconcile.Result{}, err
+			}
+		}
 	}
+
 	return reconcile.Result{}, r.setStatusReady(userAcc)
 }
 
+func (r *ReconcileUserAccount) setDisabledStatus(userAcc *toolchainv1alpha1.UserAccount) (bool, error) {
```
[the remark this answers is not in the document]
tinakurian: done here: e0c55d8218eff7e1c406842e0d44110caebaca49

pkg/controller/useraccount/useraccount_controller.go
tinakurian: done here: e0c55d8218eff7e1c406842e0d44110caebaca49

tinakurian: @sbryzak i have moved the the disabled part into another if block but i'm not sure it should stay this way. the reason we're waiting to set the status to disabled until the next reconcile is to ensure that they have both been deleted prior to setting the status. the spec is the desired state and the status is the actual state - hence we should wait until all conditions are met prior to setting the status.

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -120,33 +120,59 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+				return reconcile.Result{}, r.setStatusDisabled(userAcc)
+			} else {
+				return reconcile.Result{}, err
+			}
+		}
 	}
 	return reconcile.Result{}, r.setStatusReady(userAcc)
 }
 
+func (r *ReconcileUserAccount) checkAccountDisabled(userAcc *toolchainv1alpha1.UserAccount) (bool, error) {
```
alexeykazakov: We don't really need `checkAccountDisabled`. Just return a bool in `deleteUserIdentity()`. It should return `false` if both User and Identity do not exist (so there is nothing to delete). Similar to what `deleteUser()`/`deleteIdentity()` functions do.

tinakurian: done here: c0106baf040cff9a90957c721626bf27fb5e97bf

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -121,28 +121,35 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
 		}
-	} else if util.HasFinalizer(userAcc, userAccFinalizerName) {
-		if err = r.manageCleanUp(userAcc); err != nil {
+	} else if util.HasFinalizer(userAcc, userAccFinalizerName) || userAcc.Spec.Disabled {
+		reqLogger.Info("Deleting user and identity associated with UserAccount")
+		if err = r.deleteUserIdentity(userAcc); err != nil {
 			return reconcile.Result{}, err
 		}
+
+		if userAcc.Spec.Disabled {
```
alexeykazakov: As Shane and me already pointed we should not set "disabled" status in the same reconcile loop as the user/identity deletion. `deleteUserIdentity(userAcc)` should return a bool (see https://github.com/codeready-toolchain/member-operator/pull/105#discussion_r368707702). Let's say it returns `true` if the user OR the identity were actually deleted. If both User and Identity do not exist (so nothing to delete) then it returns `false` (nothing deleted). Similar to what we do with ensure methods.
Then we should update the status to "disabled" only if `deleteUserIdentity()` returns false. If it returns `true` (and spec.disabled == true) we could even set the status to "disabling". It will guarantee us that we update the status after the deletion. Not when we just requested it.

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -776,6 +776,118 @@ func TestUpdateStatus(t *testing.T) {
...
 
+func TestDisabledUserAccount(t *testing.T) {
+	logf.SetLogger(logf.ZapLogger(true))
+	username := "johndoe"
+	userID := uuid.NewV4().String()
+	s := scheme.Scheme
+	err := apis.AddToScheme(s)
+	require.NoError(t, err)
+
+	t.Run("disabled useraccount", func(t *testing.T) {
```
alexeykazakov: This test currently looks like an e2e test. The entire flow will be covered by the e2e tests.
In unit test we need to pre-configure everything then do just one reconcile and check the result.

So it would be something like this:

Test 1. Testing initiating disabling.
- Pre-configuration: create User Account; create User; create Identity, create NSTemplateSet. See https://github.com/codeready-toolchain/member-operator/blob/master/pkg/controller/useraccount/useraccount_controller_test.go#L347 for example
- Set UserAccount.Disabled = true
- Reconcile
- Check: User&Identity is gone. NSTemplateSet is still there. Status is set to "disabling".

Test 2. Testing "disabled" status.
- Pre-configuration: create User Account (disabled == true); create NSTemplateSet. No User or Identity.
- Reconcile
- Check: User&Identity is still missing. NSTemplateSet is still there. Status is set to "disabled".

Tests 3&4. Testing initiating disabling when only User or Identity exist. Almost the same as 1 but covering the use cases when there is only User or Identity exist (not both of them).

Test 5. Testing deletion of the disabled user account.
- pre-configuration is almost the same as in Test 2 but add finalizer to UserAccount + set UserAccount deletion timestamp to something (== being deleted). It imitates deletion of the disabled user account.
- Reconcile
- Check that User&Identity&NSTemplateSet do not exist.

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -301,8 +308,8 @@ func (r *ReconcileUserAccount) addFinalizer(userAcc *toolchainv1alpha1.UserAccou
 	return nil
 }
 
-// manageCleanUp deletes the identity, user and finalizer when the UserAccount is being deleted
-func (r *ReconcileUserAccount) manageCleanUp(userAcc *toolchainv1alpha1.UserAccount) error {
+// deleteUserIdentity deletes the identity, user and finalizer when the UserAccount is being deleted
+func (r *ReconcileUserAccount) deleteUserIdentity(userAcc *toolchainv1alpha1.UserAccount) error {
```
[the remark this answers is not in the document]
tinakurian: fixed here: e03b49a2833f31d72dfe8c54d9e9cba0159c7e0d

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -301,21 +326,15 @@ func (r *ReconcileUserAccount) addFinalizer(userAcc *toolchainv1alpha1.UserAccou
...
-	if deleted, err := r.deleteUser(userAcc); err != nil || deleted {
-		return err
-	}
-	// Remove finalizer from UserAccount
-	util.RemoveFinalizer(userAcc, userAccFinalizerName)
-	if err := r.client.Update(context.Background(), userAcc); err != nil {
-		return err
+// deleteUserAndIdentity deletes the identity, user and finalizer when the UserAccount is being deleted
+func (r *ReconcileUserAccount) deleteUserAndIdentity(logger logr.Logger, userAcc *toolchainv1alpha1.UserAccount) (bool, error) {
+	deletedIdentity, err := r.deleteIdentity(userAcc)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -121,28 +121,53 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
 		if _, createdOrUpdated, err = r.ensureIdentity(reqLogger, userAcc, user); err != nil || createdOrUpdated {
 			return reconcile.Result{}, err
 		}
-
 		if _, createdOrUpdated, err = r.ensureNSTemplateSet(reqLogger, userAcc); err != nil || createdOrUpdated {
 			return reconcile.Result{}, err
 		}
-	} else if util.HasFinalizer(userAcc, userAccFinalizerName) {
-		if err = r.manageCleanUp(userAcc); err != nil {
+	} else if util.HasFinalizer(userAcc, userAccFinalizerName) && util.IsBeingDeleted(userAcc)  {
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	} else if util.HasFinalizer(userAcc, userAccFinalizerName) && util.IsBeingDeleted(userAcc) {
```

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -301,21 +326,15 @@ func (r *ReconcileUserAccount) addFinalizer(userAcc *toolchainv1alpha1.UserAccou
...
-	}
-	if deleted, err := r.deleteUser(userAcc); err != nil || deleted {
-		return err
-	}
-	// Remove finalizer from UserAccount
-	util.RemoveFinalizer(userAcc, userAccFinalizerName)
-	if err := r.client.Update(context.Background(), userAcc); err != nil {
-		return err
+// deleteUserAndIdentity deletes the identity, user and finalizer when the UserAccount is being deleted
+func (r *ReconcileUserAccount) deleteUserAndIdentity(logger logr.Logger, userAcc *toolchainv1alpha1.UserAccount) (bool, error) {
```
golangcibot: `(*ReconcileUserAccount).deleteUserAndIdentity` - `logger` is unused (from `unparam`)

pkg/controller/useraccount/useraccount_controller_test.go
tinakurian: fixed!

pkg/controller/useraccount/useraccount_controller.go
tinakurian: fixed!

pkg/controller/useraccount/useraccount_controller.go
tinakurian: fixed!

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -121,28 +121,55 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
 		if _, createdOrUpdated, err = r.ensureNSTemplateSet(reqLogger, userAcc); err != nil || createdOrUpdated {
 			return reconcile.Result{}, err
 		}
-	} else if util.HasFinalizer(userAcc, userAccFinalizerName) {
-		if err = r.manageCleanUp(userAcc); err != nil {
+	} else if util.HasFinalizer(userAcc, userAccFinalizerName) && util.IsBeingDeleted(userAcc) {
+		reqLogger.Info("Deleting user and identity associated with UserAccount")
+		deleted, err := r.deleteUserAndIdentity(userAcc)
+		if err != nil {
 			return reconcile.Result{}, err
```
sbryzak: Should there be a status update here?

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -301,21 +328,18 @@ func (r *ReconcileUserAccount) addFinalizer(userAcc *toolchainv1alpha1.UserAccou
 	return nil
 }
 
-// manageCleanUp deletes the identity, user and finalizer when the UserAccount is being deleted
-func (r *ReconcileUserAccount) manageCleanUp(userAcc *toolchainv1alpha1.UserAccount) error {
+// deleteUserAndIdentity deletes the identity, user and finalizer when the UserAccount is being deleted
+func (r *ReconcileUserAccount) deleteUserAndIdentity(userAcc *toolchainv1alpha1.UserAccount) (bool, error) {
+
 	if deleted, err := r.deleteIdentity(userAcc); err != nil || deleted {
```
sbryzak: If the deletion is successful does that mean `deleted` == true ? If so, doesn't that mean you will return a result of `false` for a successful deletion?  I may be missing something here.

tinakurian: @sbryzak yeah, i'm returning false because to set the status to disabled i'd expect that both the user/identity are already deleted (in other words the deletion should have failed due to "is not found" error). If i had to delete the user/identity then the status should be set to disabling. hope that makes sense!

pkg/controller/useraccount/useraccount_controller.go
tinakurian: hmm, i'm not too sure, what would we set the status update to? i figured we wouldn't need to update the status here because something went wrong - we either couldn't get the user/identity (for reasons other than "is not found") or the deletion failed. In which case, i'd like to just retry because this has to be completed before the useraccount can be deleted..

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -776,6 +771,248 @@ func TestUpdateStatus(t *testing.T) {
...
+		assert.True(t, apierros.IsNotFound(err))
+
+		// Check that the associated user has been deleted
+		// since disabled has been set to true
+		user := &userv1.User{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: userAcc.Name}, user)
+		require.Error(t, err)
+		assert.True(t, apierros.IsNotFound(err))
+
+		userAcc = &toolchainv1alpha1.UserAccount{}
```
alexeykazakov: Check that User/Identity is deleted after the first reconcile. Also check the status (Terminating).
And check that the finalizer is deleted (and User/Identity are still deleted too) after the second reconcile.

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -448,6 +471,37 @@ func (r *ReconcileUserAccount) setStatusReady(userAcc *toolchainv1alpha1.UserAcc
...
+func (r *ReconcileUserAccount) setStatusTerminating(userAcc *toolchainv1alpha1.UserAccount, message string) error {
+	return r.updateStatusConditions(
+		userAcc,
+		toolchainv1alpha1.Condition{
+			Type:    toolchainv1alpha1.ConditionReady,
+			Status:  corev1.ConditionFalse,
+			Reason:  toolchainv1alpha1.UserAccountTerminatingReason,
+			Message: message,
+		})
+}
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
}

```

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -550,37 +550,52 @@ func TestReconcile(t *testing.T) {
...
 		assert.Equal(t, reconcile.Result{}, res)
+		require.EqualError(t, err, fmt.Sprintf("failed to remove finalizer: unable to remove finalizer for user account %s", userAcc.Name))
+
+		userAcc = &toolchainv1alpha1.UserAccount{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: userAcc.Name, Namespace: "toolchain-member"}, userAcc)
 		require.NoError(t, err)
 
+		test.AssertConditionsMatch(t, userAcc.Status.Conditions,
+			toolchainv1alpha1.Condition{
+				Type:   toolchainv1alpha1.ConditionReady,
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

tinakurian:
> Looks good overall! Just a few suggestions.

Thanks @alexeykazakov. I think i have addressed all your comments in this commit: 854ef11f2560ded5644197dd5a1054fea9dd3dfb

pkg/controller/useraccount/useraccount_controller.go
tinakurian: @alexeykazakov @sbryzak since we agreed to do this in separate reconcile loops i'm going to mark this as resolved - please re-open if there is something i missed.

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -776,6 +771,248 @@ func TestUpdateStatus(t *testing.T) {
...
+		// Set disabled to true
+		userAcc.Spec.Disabled = true
+		err = r.client.Update(context.TODO(), userAcc)
+		require.NoError(t, err)
+
+		res, err = r.Reconcile(req)
+		assert.Equal(t, reconcile.Result{}, res)
+		require.NoError(t, err)
+
+		res, err = r.Reconcile(req)
```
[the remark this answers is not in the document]
tinakurian: @alexeykazakov resolving this due to our discussion in slack - please re-open if i missed something!

pkg/controller/useraccount/useraccount_controller_test.go
tinakurian: @alexeykazakov marking as resolved. the status will not be "terminating" here. Since there will be no error returned. Also since our discussion in slack both user/identity will be deleted in separate reconciles. Please re-open if i misunderstood though! 

however, failing to remove the finalizer is tested in another test and i've ensured that the status is checked and that it is "terminating"

tinakurian: test for removing finalizer failing (state terminating): https://github.com/codeready-toolchain/member-operator/pull/105/files#diff-fc74d9d211758dbe5bd817320a8fe2a1R530

test for removing user/identity failing (state terminating): https://github.com/codeready-toolchain/member-operator/pull/105/files#diff-fc74d9d211758dbe5bd817320a8fe2a1R607

test for deleting both user and identity failing (state terminating): https://github.com/codeready-toolchain/member-operator/pull/105/files#diff-fc74d9d211758dbe5bd817320a8fe2a1R658

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -301,21 +326,19 @@ func (r *ReconcileUserAccount) addFinalizer(userAcc *toolchainv1alpha1.UserAccou
 	return nil
 }
 
-// manageCleanUp deletes the identity, user and finalizer when the UserAccount is being deleted
-func (r *ReconcileUserAccount) manageCleanUp(userAcc *toolchainv1alpha1.UserAccount) error {
+// deleteUserAndIdentity deletes the identity and user. Returns bool and error indicating that
+// whether the user/identity were deleted.
+func (r *ReconcileUserAccount) deleteUserAndIdentity(userAcc *toolchainv1alpha1.UserAccount) (bool, error) {
+
 	if deleted, err := r.deleteIdentity(userAcc); err != nil || deleted {
```
alexeykazakov: It's confusing. You return `deleted:false` if the identity was deleted and there is no error?

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -121,28 +121,53 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+	} else if util.HasFinalizer(userAcc, userAccFinalizerName) && util.IsBeingDeleted(userAcc) {
+		reqLogger.Info("Deleting user and identity associated with UserAccount")
+		deleted, err := r.deleteUserAndIdentity(userAcc)
+		if err != nil {
+			return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, userAcc, r.setStatusTerminating, err, "failed to delete user/identity")
+		}
+
+		// Remove finalizer from UserAccount
+		if deleted {
+			util.RemoveFinalizer(userAcc, userAccFinalizerName)
```
alexeykazakov: So we don't remove the finalizer if there is nothing to delete?

We need:
1. If User OR Identity was deleted then just exit (the status is Terminating). Wait for the next reconcile.
2. If there is nothing to delete (no User or Identity exist). Then remove the finalizer. We are done.

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -550,37 +550,52 @@ func TestReconcile(t *testing.T) {
...
+			userAcc := obj.(*toolchainv1alpha1.UserAccount)
+			userAcc.Finalizers = finalizers
+			return fmt.Errorf("unable to remove finalizer for user account %s", userAcc.Name)
+		}
+
+		res, err = r.Reconcile(req)
+		require.NoError(t, err)
+		assert.Equal(t, reconcile.Result{}, res)
+
+		res, err = r.Reconcile(req)
```
alexeykazakov: OK if you still want to use multiple reconciles in the same test then at least check the status and relevant resources between them (in all tests).
For example the status should be "Terminating" and finalizer is not removed until the last reconcile.

pkg/controller/useraccount/useraccount_controller.go
tinakurian: no, we remove the finalizer only after the user and identity have both been deleted. if we don't wait until after we will not enter into the same else-if condition. 

1. it does this, but the status is not set to "terminating" yet.. i will update this. 
2. yes it does this.

alexeykazakov: what if there is nothing to delete? Identity & User do not exist. Then you won't delete the finalizer.

As I said we need to remove the finalizer when there is nothing to delete (so deleted==false). Why there is nothing to delete? Because it was just deleted in the previous reconcile loop or because it was never created we do not care. We can't remove the finalizer until we make sure user/identity are gone for good (vs. we just requested kube to delete them).

tinakurian: no, if there is nothing to delete (everything has already been deleted) the finalizer will get removed. There is a test for this: https://github.com/codeready-toolchain/member-operator/pull/105/files#diff-fc74d9d211758dbe5bd817320a8fe2a1R904

pkg/controller/useraccount/useraccount_controller_test.go
tinakurian: fixed

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -121,28 +121,55 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
 			return reconcile.Result{}, err
 		}
-	} else if util.HasFinalizer(userAcc, userAccFinalizerName) {
-		if err = r.manageCleanUp(userAcc); err != nil {
-			return reconcile.Result{}, err
+	} else if util.HasFinalizer(userAcc, userAccFinalizerName) && util.IsBeingDeleted(userAcc) {
+		reqLogger.Info("Deleting user and identity associated with UserAccount")
+		deleted, err := r.deleteUserAndIdentity(userAcc)
+		if err != nil {
+			return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, userAcc, r.setStatusTerminating, err, "failed to delete user/identity")
```
alexeykazakov: I think we should set the status first then try to delete:
```
reqLogger.Info("Terminating UserAccount")
if err := r.setStatusTerminating(userAcc); err != nil {
	reqLogger.Error(err, "error updating status")
}
deleted, err := r.deleteUserAndIdentity(userAcc)
if err != nil {
	return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, userAcc, r.setStatusTerminating, err, "failed to delete user/identity")
}

// Remove finalizer from UserAccount
if !deleted {
	util.RemoveFinalizer(userAcc, userAccFinalizerName)
}
...
```

pkg/controller/useraccount/useraccount_controller.go
alexeykazakov: yeah looks good now

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -121,28 +121,55 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+				return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, userAcc, r.setStatusTerminating, err, "failed to remove finalizer")
+			}
+
+			return reconcile.Result{}, nil
+		}
+
+		return reconcile.Result{}, r.setStatusTerminating(userAcc, "deleting user/identity")
+	} else if userAcc.Spec.Disabled {
+		reqLogger.Info("Deleting user and identity associated with UserAccount")
+		deleted, err := r.deleteUserAndIdentity(userAcc)
```
alexeykazakov: Same here. Let's set the disabling status first:
```
reqLogger.Info("Disabling UserAccount")
if err := r.setStatusDisabling(userAcc, "deleting user/identity"); err != nil {
	reqLogger.Error(err, "error updating status")
}
deleted, err := r.deleteUserAndIdentity(userAcc)
if err != nil {
	return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, userAcc, r.setStatusDisabling, err, "failed to delete user/identity")
}

if !deleted {
	return reconcile.Result{}, r.setStatusDisabled(userAcc)
}
return reconcile.Result{}, r.setStatusDisabling(userAcc, "deleting user/identity")
```

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -776,6 +752,235 @@ func TestUpdateStatus(t *testing.T) {
...
+		assertNSTemplateFound(t, r, userAcc)
+
+		userAcc = &toolchainv1alpha1.UserAccount{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: userAcc.Name, Namespace: "toolchain-member"}, userAcc)
+		require.NoError(t, err)
+		require.False(t, util.HasFinalizer(userAcc, userAccFinalizerName))
+	})
+}
+
+func assertStatus(t *testing.T, r *ReconcileUserAccount, account *toolchainv1alpha1.UserAccount, status string, msg string) {
```
alexeykazakov:
```suggestion
func assertNotReadyStatus(t *testing.T, r *ReconcileUserAccount, account *toolchainv1alpha1.UserAccount, reason, msg string) {
```

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -776,6 +752,235 @@ func TestUpdateStatus(t *testing.T) {
...
+			},
+		},
+	}
+
+	t.Run("disabling useraccount", func(t *testing.T) {
+		// given
+		r, req, _ := prepareReconcile(t, username, userAcc, preexistingUser, preexistingIdentity, preexistingNsTmplSet)
+
+		//when
+		res, err := r.Reconcile(req)
```
alexeykazakov: What do you need this reconcile for?

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -776,6 +752,235 @@ func TestUpdateStatus(t *testing.T) {
...
+		assert.Equal(t, reconcile.Result{}, res)
+		require.NoError(t, err)
+
+		//then
```
alexeykazakov: It's rather "when", right?
```
// given the pre-configured reconcile
prepareReconcile()...
...
// when disabling the user account
userAcc.Spec.Disabled = true
...
update...
reconcile...
// then status should be Disabling...
assertNotReadyStatus(...)

```

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -776,6 +752,235 @@ func TestUpdateStatus(t *testing.T) {
...
+		// Set disabled to true
+		userAcc.Spec.Disabled = true
+		err = r.client.Update(context.TODO(), userAcc)
+		require.NoError(t, err)
+
+		res, err = r.Reconcile(req)
+		assert.Equal(t, reconcile.Result{}, res)
+		require.NoError(t, err)
+
+		assertStatus(t, r, userAcc, "Disabling", "deleting user and identity resources")
```
alexeykazakov: We could also `assertIdentityNotFound(t, r, ToIdentityName(userAcc.Spec.UserID))` but User is there (or whatever we delete first).

pkg/controller/useraccount/useraccount_controller_test.go
tinakurian: this reconcile enters us into this condition: if !util.IsBeingDeleted(userAcc) && !userAcc.Spec.Disabled {...} prior to setting disabled to true

alexeykazakov: This condition should be setup as the pre-configuration in prepareReconcile(), shouldn't it?

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -121,28 +121,61 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
 		}
+		deleted, err := r.deleteIdentityAndUser(userAcc)
+		if err != nil {
+			return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, userAcc, r.setStatusDisabling, err, "failed to delete user/identity")
+		}
+
+		if !deleted {
+			return reconcile.Result{}, r.setStatusDisabled(userAcc)
+		}
+		return reconcile.Result{}, r.setStatusDisabling(userAcc, "deleting user/identity")
```
alexeykazakov: Same here.
```suggestion
		return reconcile.Result{}, nil
```

pkg/controller/useraccount/useraccount_controller_test.go
tinakurian: @alexeykazakov yeah, that's right! forgot! sorry, fixed!

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -121,28 +121,61 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
...
+		// Remove finalizer from UserAccount
+		if !deleted {
+			util.RemoveFinalizer(userAcc, userAccFinalizerName)
+			if err := r.client.Update(context.Background(), userAcc); err != nil {
+				return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, userAcc, r.setStatusTerminating, err, "failed to remove finalizer")
+			}
+
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, r.setStatusTerminating(userAcc, "deleting user/identity")
```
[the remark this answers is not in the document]
tinakurian: Fixed!

pkg/controller/useraccount/useraccount_controller.go
tinakurian: Fixed!

PR #111, opened 2020-01-03
 Fix 'cannot list resource bindings' error

alexeykazakov:
See https://github.com/codeready-toolchain/host-operator/pull/128


alexeykazakov:
The issue should be fixed by now. Closing.

PR #116, opened 2020-01-07
chore: remove unneeded permissons

MatousJobanek:
Since the RBAC issue with `reflector.go:95: Failed to list` errors should be fixed I decided to remove unneeded permissions from our roles. I didn't spend much time investigating which particular permissions are used and which are not, I just removed those that were apparently added just as a workaround for making our logs clean and happy.

### DIFF
--- a/deploy/cluster_role.yaml
+++ b/deploy/cluster_role.yaml
@@ -15,7 +15,7 @@ rules:
   resources:
   - nodes
   verbs:
-  - list
+  - "list"
 - apiGroups:
   - ""
   resources:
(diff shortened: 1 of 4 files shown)


MatousJobanek:
> Can we make sure that we don't have any permission issues in the logs in e2e tests for this PR?

I checked that locally but can try to add something in the e2e test setup

MatousJobanek:
> You mean you checked it when run e2e tests locally? It's good enough (if it was 4.* cluster).

yes, exactly - I ran it against 48h cluster at least 5 times and no `reflector.go:95: Failed to list` error was present in the logs

PR #118, opened 2020-01-10
Using reasons in api

tinakurian:
Updates the codeready-toolchain/api version and the reasons from codeready-toolchain/api in both controller and tests. relates to: https://github.com/codeready-toolchain/api/pull/86

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ require (
 
 // Pinned to kubernetes-1.14.1
 replace (
+	github.com/codeready-toolchain/api v0.0.0-20200110043643-8b83f07304f7 => github.com/tinakurian/api v0.0.0-20200110162517-96ec111a649e
 	// using 'github.com/openshift/api@release-4.2'
 	github.com/openshift/api => github.com/openshift/api v0.0.0-20190927182313-d4a64ec2cbd8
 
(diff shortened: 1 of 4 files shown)


pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -110,7 +110,7 @@ func TestReconcile(t *testing.T) {
 				toolchainv1alpha1.Condition{
 					Type:   toolchainv1alpha1.ConditionReady,
 					Status: corev1.ConditionFalse,
-					Reason: "Provisioning",
+					Reason: toolchainv1alpha1.ProvisioningReason,
```
alexeykazakov: We use strings instead of consts in tests intentionally. If we mistakenly use wrong const in the runtime code then it would be caught by the test.
So please revoke your changes in the test.

alexeykazakov:
Yeah, the reasons are moved because they will be used in other controllers too. More specifically in MUR controller. It will check the specific UserAccount status when synchronizing it with MUR status.

> the variable names should contain UserAccount name to make it clear what CR type they are used for

It's a good point. +1

go.mod
```diff
@@ -24,6 +24,7 @@ require (
 
 // Pinned to kubernetes-1.14.1
 replace (
+	github.com/codeready-toolchain/api v0.0.0-20200110043643-8b83f07304f7 => github.com/tinakurian/api v0.0.0-20200110162517-96ec111a649e
```
alexeykazakov: Just a reminder. Don’t forget to delete it before merging :)

PR #121, opened 2020-01-31
refactor NSTemplateSet/Namespace ownership

xcoulon:
add a finalizer on the NSTemplateSet resources
so that the controller is called when an NSTemplateSet
is deleted. When this happens, the controller sets the
status of the NSTemplateSet resource to `ready=false/reason=terminating`
and deletes the "child" namespaces one by one.
Once all namespaces have been deleted (or marked for deletion, ie
they have a deletion timestamp), the controller removes the
finalizer on the NSTemplateSet resource so that the GC
can delete it for good.

also:
- when creating the NSTemplateSet in the UserAccount controller,
add the finalizer aforementioned
- do not set the ownership to the NSTemplateSet on the user namespaces
- use a custom Event Handler to map the event on a namespace to its
associated NSTemplateSet (if a specific label is found)
- add permission to the member-operator service account
to delete namespaces

Fixes https://issues.redhat.com/browse/CRT-429


xcoulon:
marking as "draft" for now, until https://github.com/codeready-toolchain/toolchain-common/pull/73 is merged and https://github.com/codeready-toolchain/member-operator/pull/121/files#diff-37aff102a57d3d7b797f152915a6dc16R63 is removed

No specific e2e tests added for this PR: we already have https://github.com/codeready-toolchain/toolchain-e2e/blob/master/test/e2e/e2e_test.go#L117 and it should still work.

xcoulon:
> marking as "draft" for now, until [codeready-toolchain/toolchain-common#73](https://github.com/codeready-toolchain/toolchain-common/pull/73) is merged and https://github.com/codeready-toolchain/member-operator/pull/121/files#diff-37aff102a57d3d7b797f152915a6dc16R63 is removed

codeready-toolchain/toolchain-common#73 was merged and this PR is now ready for review.

go.mod
```diff
@@ -45,4 +60,6 @@ replace (
 	github.com/coreos/prometheus-operator => github.com/coreos/prometheus-operator v0.29.0
 )
 
+replace github.com/codeready-toolchain/toolchain-common => github.com/xcoulon/toolchain-common v0.0.0-20200131154201-fea8485ab790
```
alexeykazakov: You forgot to delete this replace.

xcoulon: no, I pushed without committing the removal, first! 🤦‍♂

xcoulon: done in https://github.com/codeready-toolchain/member-operator/pull/121/commits/8398463cd4b37ca57718e879cf8110ce059cd84c

pkg/controller/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -527,6 +528,137 @@ func TestUpdateStatusToProvisionedWhenPreviouslyWasSetToFailed(t *testing.T) {
...
+				require.NoError(t, err)
+				// get the second namespace and check its deletion timestamp
+				secondNS := corev1.Namespace{}
+				secondtNSName := fmt.Sprintf("%s-%s", username, nsTmplSet.Spec.Namespaces[1].Type)
+				err = r.client.Get(context.TODO(), types.NamespacedName{
+					Name: secondtNSName,
+				}, &secondNS)
+				require.NoError(t, err)
+				assert.NotNil(t, secondNS.GetDeletionTimestamp(), "expected a deletion timestamp on '%s' namespace", secondtNSName)
+
```
alexeykazakov: Let's also check the status, so it's still terminating. And the finalizer is still present.

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -840,8 +841,9 @@ func newNSTmplSetSpec() toolchainv1alpha1.NSTemplateSetSpec {
 func newNSTmplSetWithStatus(username, reason, meessage string) *toolchainv1alpha1.NSTemplateSet {
 	return &toolchainv1alpha1.NSTemplateSet{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      username,
-			Namespace: "toolchain-member",
+			Name:       username,
+			Namespace:  "toolchain-member",
+			Finalizers: []string{toolchainv1alpha1.FinalizerName},
```
alexeykazakov: If I understand it correctly the finalizers should be added by the primary resource controller. Not by the "parent" controller. In this case it should be done by NSTemplateSet controller. As we do it for UserAccounts: https://github.com/codeready-toolchain/member-operator/

Finalizer is an implementation detail of the recourse reconciler. The clients should not care about that. In other words it's only the NSTemplateSet Controller knows how to terminate the nstemplateset resource.

alexeykazakov:
And also worth mentioning that this PR will break the owner relationship for the existing namespaces. Which is fine. We can afford it. But we should update the existing resources manually in our crt clusters before closing the JIRA.
We will need to introduce some kind of migration mechanism for such breaking changes in the future when we run real production clusters.

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -57,35 +51,55 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+// 			return []reconcile.Request{
+// 				{
+// 					NamespacedName: types.NamespacedName{Namespace: "toolchain-member-operator", Name: ownerLbl},
+// 				},
+// 			}
+// 		}
+// 	}
+// 	// the obj was not a namespace or it did not have the required label.
+// 	return []reconcile.Request{}
+// }
```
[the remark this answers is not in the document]
xcoulon: yes, absolutely! 🤦‍♂ 
Let me remove it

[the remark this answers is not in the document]
xcoulon: done in bb02e64

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -106,6 +120,38 @@ func (r *ReconcileNSTemplateSet) Reconcile(request reconcile.Request) (reconcile
 		reqLogger.Error(err, "failed to get NSTemplateSet")
 		return reconcile.Result{}, err
 	}
+	if util.IsBeingDeleted(nsTmplSet) {
+		// if the NSTemplateSet has no finalizer, then we don't have anything to do
+		if len(nsTmplSet.Finalizers) == 0 {
+			reqLogger.Info("NSTemplateSet resource is terminated")
+			return reconcile.Result{}, nil
+		}
+		// since the NSTmplSet resource is being deleted, we need to delete all "child" namespaces explicitly
```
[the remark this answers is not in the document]
xcoulon: sure! done in d954af2

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -106,6 +120,38 @@ func (r *ReconcileNSTemplateSet) Reconcile(request reconcile.Request) (reconcile
...
+		// delete the first namespace which (still) exists and is not in a terminating state
+		reqLogger.Info("checking user namepaces associated with the deleted NSTemplateSet...")
+		for _, userNS := range userNamespaces {
+			if !util.IsBeingDeleted(&userNS) {
+				reqLogger.Info("deleting a user namepace associated with the deleted NSTemplateSet", "namespace", userNS.Name)
+				if err := r.client.Delete(context.TODO(), &userNS); err != nil {
+					return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, nsTmplSet, r.setStatusDeletingFailed, err, "failed to delete user namespace '%s'", userNS.Name)
+				}
+				// set the NSTemplateSet status to `ready=false / reason=terminating` while the user namespaces are being terminated...
+				return reconcile.Result{}, r.setStatusTerminating(nsTmplSet)
```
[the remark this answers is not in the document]
xcoulon: done in d954af2

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -57,35 +51,55 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 	}
 
-	// Watch for changes to secondary resource
-	enqueueRequestForOwner := &handler.EnqueueRequestForOwner{
-		IsController: true,
-		OwnerType:    &toolchainv1alpha1.NSTemplateSet{},
-	}
-	if err := c.Watch(&source.Kind{Type: &corev1.Namespace{}}, enqueueRequestForOwner); err != nil {
+	// Watch for changes to secondary resources: Namespaces associated with an NSTemplateSet (not owned, though - see https://issues.redhat.com/browse/CRT-429)
+	if err := c.Watch(&source.Kind{Type: &corev1.Namespace{}}, commoncontroller.MapToOwnerByLabel("", toolchainv1alpha1.LabelKeyPrefix+"owner")); err != nil {
```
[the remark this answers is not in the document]
xcoulon: 🤦‍♂ 
done in c3a47a6

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -117,18 +163,24 @@ func (r *ReconcileNSTemplateSet) Reconcile(request reconcile.Request) (reconcile
 	return reconcile.Result{}, r.setStatusReady(nsTmplSet)
 }
 
-func (r *ReconcileNSTemplateSet) ensureUserNamespaces(logger logr.Logger, nsTmplSet *toolchainv1alpha1.NSTemplateSet) (bool, error) {
-	username := nsTmplSet.GetName()
-
+func (r *NSTemplateSetReconciler) fetchUserNamespaces(username string) ([]corev1.Namespace, error) {
```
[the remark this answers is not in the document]
xcoulon: good point!

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -296,44 +349,65 @@ func (r *ReconcileNSTemplateSet) updateStatusConditions(nsTmplSet *toolchainv1al
...
+}
+
+func (r *NSTemplateSetReconciler) setStatusDeletingFailed(nsTmplSet *toolchainv1alpha1.NSTemplateSet, message string) error {
 	return r.updateStatusConditions(
 		nsTmplSet,
 		toolchainv1alpha1.Condition{
 			Type:    toolchainv1alpha1.ConditionReady,
 			Status:  corev1.ConditionFalse,
-			Reason:  unableToProvisionReason,
+			Reason:  toolchainv1alpha1.NSTemplateSetUnableToProvisionReason,
```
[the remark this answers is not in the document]
xcoulon: good point. Let me do that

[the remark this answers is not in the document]
xcoulon: done in 225e152 (and also renamed func and reorganize/regroup other funcs)

pkg/controller/useraccount/useraccount_controller_test.go
xcoulon: ah right. Let me refactor this part of the PR.

xcoulon:
> And also worth mentioning that this PR will break the owner relationship for the existing namespaces. Which is fine. We can afford it. But we should update the existing resources manually in our crt clusters before closing the JIRA.
> We will need to introduce some kind of migration mechanism for such breaking changes in the future when we run real production clusters.

agreed on both parts of this comment: existing CRs on our dev cluster and later on production. Somehow, this is more or less similar to what we used to do in the past with our migration scripts.

pkg/controller/useraccount/useraccount_controller_test.go
xcoulon: done in e3dbfff

pkg/controller/nstemplateset/nstemplateset_controller_test.go
xcoulon: done in 1cc75b7

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -106,6 +97,42 @@ func (r *ReconcileNSTemplateSet) Reconcile(request reconcile.Request) (reconcile
 		reqLogger.Error(err, "failed to get NSTemplateSet")
 		return reconcile.Result{}, err
 	}
+	if util.IsBeingDeleted(nsTmplSet) {
+		// if the NSTemplateSet has no finalizer, then we don't have anything to do
+		if len(nsTmplSet.Finalizers) == 0 {
```
MatousJobanek: you could use github.com/redhat-cop/operator-utils for manipulating with finalizers - see how it is done for UserAccount: https://github.com/codeready-toolchain/member-operator/blob/master/pkg/controller/useraccount/useraccount_controller.go#L292-L302 or MUR: https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/masteruserrecord/masteruserrecord_controller.go#L118-L152
The same repo can be used also for removal of the finalizer - use `RemoveFinalizer` function

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -106,6 +97,46 @@ func (r *ReconcileNSTemplateSet) Reconcile(request reconcile.Request) (reconcile
...
+		reqLogger.Info("checking user namepaces associated with the deleted NSTemplateSet...")
+		for _, userNS := range userNamespaces {
+			if !util.IsBeingDeleted(&userNS) {
+				reqLogger.Info("deleting a user namepace associated with the deleted NSTemplateSet", "namespace", userNS.Name)
+				if err := r.client.Delete(context.TODO(), &userNS); err != nil {
+					return reconcile.Result{}, r.wrapErrorWithStatusUpdate(reqLogger, nsTmplSet, r.setStatusTerminatingFailed, err, "failed to delete user namespace '%s'", userNS.Name)
+				}
+				return reconcile.Result{}, nil
+			}
+		}
```
MatousJobanek: could you please make the deletion in a separate method to make the `Reconcile` method shorter and easier to read? Thanks :-)

pkg/controller/nstemplateset/nstemplateset_controller.go
MatousJobanek: We should not check just if there is any finalizer, but if there is our specific finalizer `toolchainv1alpha1.FinalizerName`

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -106,6 +97,46 @@ func (r *ReconcileNSTemplateSet) Reconcile(request reconcile.Request) (reconcile
...
+		// if no namespace was to be deleted, then we can remove the finalizer and we're done
+		reqLogger.Info("NSTemplateSet resource is ready to be terminated: all related user namespaces have been marked for deletion")
+		nsTmplSet.SetFinalizers([]string{})
```
MatousJobanek: similarly here - we should remove only our specific finalizer, not all of them - the functions from operator-utils should help you ;-)

pkg/controller/nstemplateset/nstemplateset_controller.go
xcoulon: sure! I wanted to do it but for some reason it slipped out of my mind 🤦‍♂

pkg/controller/nstemplateset/nstemplateset_controller.go
xcoulon: good point! Let me fix that

MatousJobanek:
> And also worth mentioning that this PR will break the owner relationship for the existing namespaces. 

I believe that this is one of the reasons why the reconcile function is called for all present CR that are being watched as primary resource after the start of the controller/operator. 
I have observed this behavior every time when I started the operator and some instances of the CR were present, but I haven't found any note about that. If that's true, then it would make sense to incorporate such a migration mechanism into our controller. In this case, we would need to remove the owner-reference if it is present.

pkg/controller/nstemplateset/nstemplateset_controller.go
xcoulon: hmm, that's already done: https://github.com/codeready-toolchain/member-operator/pull/121/commits/e3dbfff3555c5e08d381398fb9a1994e59faf9d3

alexeykazakov:
> I believe that this is one of the reasons why the reconcile function is called for all present CR that are being watched as primary resource after the start of the controller/operator.

Yes, but I thought that we are introducing a new owner label in namespaces. In that case the new controller wouldn't start watching the existing namespaces. So, auto-reconciling all the NSTemplateSet after re-deploying the operator wouldn't help. But this PR actually uses the existing owner label. So all existing namespaces already have this label and the new watcher will work correctly for all existing and new namespaces. So, it's not an issue for this PR. We will just need to delete the owner reference from meta of the existing namespaces after re-deploying the operator. We don't have to do any migration before deploying it.

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -106,6 +97,46 @@ func (r *ReconcileNSTemplateSet) Reconcile(request reconcile.Request) (reconcile
 		reqLogger.Error(err, "failed to get NSTemplateSet")
 		return reconcile.Result{}, err
 	}
+	if util.IsBeingDeleted(nsTmplSet) {
+		// if the NSTemplateSet has no finalizer, then we don't have anything to do
+		if len(nsTmplSet.Finalizers) == 0 {
```
alexeykazakov:
```suggestion
		if !util.HasFinalizer(nsTmplSet, toolchainv1alpha1.FinalizerName) {
```

pkg/controller/nstemplateset/nstemplateset_controller.go
alexeykazakov: I think this is what @MatousJobanek means: https://github.com/codeready-toolchain/member-operator/pull/121#discussion_r374371849

xcoulon: ok, got it now! thanks for the clarification

pkg/controller/nstemplateset/nstemplateset_controller.go
xcoulon: will incorporate this change during the refactoring (moving the deletion part in a separate func)

xcoulon:
also, see https://github.com/codeready-toolchain/toolchain-e2e/pull/72 for e2e tests

pkg/controller/nstemplateset/nstemplateset_controller.go
xcoulon: done in b1f9439

pkg/controller/nstemplateset/nstemplateset_controller.go
xcoulon: done in b1f9439

pkg/controller/nstemplateset/nstemplateset_controller.go
xcoulon: done in b1f9439

pkg/controller/nstemplateset/nstemplateset_controller.go
xcoulon: done in b1f9439

xcoulon:
thanks for the reviews and feedback, @alexeykazakov and @MatousJobanek 🙌

PR #122, opened 2020-02-05
Set mur spec

tinakurian:
set mur.Spec.UserAccounts[x].Spec.Disabled field


pkg/controller/useraccountstatus/useraccount_status_controller.go
```diff
@@ -114,6 +114,7 @@ func (r *ReconcileUserAccountStatus) updateMasterUserRecord(userAcc *toolchainv1
 	for i, account := range mur.Spec.UserAccounts {
 		if account.TargetCluster == fedCluster.OwnerClusterName {
 			mur.Spec.UserAccounts[i].SyncIndex = userAcc.ResourceVersion
+			mur.Spec.UserAccounts[i].Spec.Disabled = userAcc.Spec.Disabled
```
alexeykazakov: Why would we need that? We are not supposed to change MUR Spec from member/UserAccount. It works the other way around.
`SyncIndex` is a special case. It's just a way for us to trigger a MUR reconcile when UserAccount has changed.

MatousJobanek:
@tinakurian May I know why the PR is in member-operator repo? The Disabled field should be set in MUR controller - that controller is the one who should set the required spec to `UserAccount` (including the `disabled` field) - see this [part](https://github.com/codeready-toolchain/host-operator/blob/master/pkg/controller/masteruserrecord/masteruserrecord_controller.go#L154-L200). 
As is said by Alexey in his [comment](https://github.com/codeready-toolchain/member-operator/pull/122#discussion_r375459130) we set `SyncIndex` from UserAccount controller just for the sake of ensuring that the statuses of these two resources are in sync. The `Disabled` param in `MUR` should be set from `UserSignup`

PR #129, opened 2020-03-02
feat(CRT-474): support promotion to other tiers

MatousJobanek:
Add support to promote from one tier to another and back. When there is a change in the tier then it:
* checks if the user's provisioned namespaces contain the same list of namespaces as are set in NSTemplateSet, if not then it deletes the redundant ones
* takes the template of the current tier and compares all objects from that template if they all also exist in the template of the new tier, if not then it deletes the redundant objects

The changes in this PR also guarantee that when there is an update of the namespace revisions in the NSTemplateSet then the user's namespaces will be updated using the new template. However, there is no comparison of the objects from the new/old version of the template - that will be part of other PR.

uses https://github.com/codeready-toolchain/api/pull/111
paired PR https://github.com/codeready-toolchain/toolchain-e2e/pull/76


pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -286,15 +312,56 @@ func (r *NSTemplateSetReconciler) ensureInnerNamespaceResources(logger logr.Logg
 	return nil
 }
 
-// nextNamespaceToProvision returns first namespace (from given namespaces) with
-// namespace status is active and revision not set
-// or namespace present in tcNamespaces but not found in given namespaces
-func nextNamespaceToProvision(tcNamespaces []toolchainv1alpha1.NSTemplateSetNamespace, namespaces []corev1.Namespace) (*toolchainv1alpha1.NSTemplateSetNamespace, *corev1.Namespace, bool) {
-	for _, tcNamespace := range tcNamespaces {
+func (r *NSTemplateSetReconciler) deleteRedundantObjects(logger logr.Logger, nsType string, params map[string]string, namespace *corev1.Namespace, newObjects []runtime.RawExtension) error {
```
golangcibot: `(*NSTemplateSetReconciler).deleteRedundantObjects` - `logger` is unused (from `unparam`)

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -414,6 +414,92 @@ func TestReconcile(t *testing.T) {
...
+		r, req, _ := prepareReconcile(t, username, userAcc, preexistingUser, preexistingIdentity, preexistingNsTmplSet)
+
+		//when
+		_, err := r.Reconcile(req)
+
+		//then
+		require.NoError(t, err)
+
+		updatedNSTmplSet := &toolchainv1alpha1.NSTemplateSet{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: req.Namespace, Name: username}, updatedNSTmplSet)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -414,6 +414,92 @@ func TestReconcile(t *testing.T) {
...
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -303,6 +370,19 @@ func nextNamespaceToProvision(tcNamespaces []toolchainv1alpha1.NSTemplateSetName
 	return nil, nil, false
 }
 
+func nextNamespaceToDeprovision(tcNamespaces []toolchainv1alpha1.NSTemplateSetNamespace, namespaces []corev1.Namespace) (*corev1.Namespace, bool) {
```
alexeykazakov: Let's add a description/comment to this method. It can be useful to have a comment describing why and what we want to deprovison here.

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -286,15 +312,56 @@ func (r *NSTemplateSetReconciler) ensureInnerNamespaceResources(logger logr.Logg
 	return nil
 }
 
-// nextNamespaceToProvision returns first namespace (from given namespaces) with
-// namespace status is active and revision not set
-// or namespace present in tcNamespaces but not found in given namespaces
-func nextNamespaceToProvision(tcNamespaces []toolchainv1alpha1.NSTemplateSetNamespace, namespaces []corev1.Namespace) (*toolchainv1alpha1.NSTemplateSetNamespace, *corev1.Namespace, bool) {
-	for _, tcNamespace := range tcNamespaces {
+func (r *NSTemplateSetReconciler) deleteRedundantObjects(nsType string, params map[string]string, namespace *corev1.Namespace, newObjects []runtime.RawExtension) error {
```
alexeykazakov: Let's add a comment to this method too.

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -286,15 +312,56 @@ func (r *NSTemplateSetReconciler) ensureInnerNamespaceResources(logger logr.Logg
...
+	currentTier := namespace.Labels[toolchainv1alpha1.TierLabelKey]
+	currentTmpl, err := r.getTemplateContent(currentTier, nsType)
+	if err != nil {
+		return errs.Wrapf(err, "failed to to retrieve template of the current tier '%s' for namespace '%s'", currentTier, namespace.Name)
+	}
+
+	tmplProcessor := template.NewProcessor(r.client, r.scheme)
+	currentObjs, err := tmplProcessor.Process(currentTmpl, params, template.RetainAllButNamespaces)
```
alexeykazakov: This will work for the usecase when we change the tier. But I wonder if we could just load all the objects in the namespace with `labels["provider"] == "codeready-toolchain"`, so, we get the actual objects and not just objects which are defined in the used tier. In this case we could re-use it later for template updates. WDYT?
But I'm fine with keeping it as is for now and address the template update case separately later.

MatousJobanek: Good idea, but I'm not sure if there is some generic list-object that we could use for that operation. I'm afraid that we would need to know which kinds of objects we want to list and list them separately. I tried to use the `UnstructuredList` to list all objects but it failed exactly with the error that I expected: `Object 'Kind' is missing in 'unstructured object has no kind'`.
We cannot make the "list request" because without knowing the GVK we cannot contsruct the API url for that.

MatousJobanek: I know that this will work only for changes of the tier; for the updates of the templates itself we have two options:
* store all revisions of the tiers in OS
* download the older revision from GH

I would go with the second option - downloading it from GH

pkg/controller/nstemplateset/nstemplateset_controller.go
MatousJobanek: done 29f23da

pkg/controller/nstemplateset/nstemplateset_controller.go
MatousJobanek: done done 29f23da

MatousJobanek:
> Can we add some tests for error handling too?

added for those cases when it makes sense a3426fa

go.mod
```diff
@@ -29,6 +29,8 @@ require (
 	sigs.k8s.io/kubefed v0.1.0-rc6.0.20200224204536-6207193c49f7
 )
 
+replace github.com/codeready-toolchain/api => github.com/MatousJobanek/api v0.0.0-20200302133147-9f81d9ad1330
```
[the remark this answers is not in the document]
MatousJobanek: updated 40411a5

pkg/controller/nstemplateset/nstemplateset_controller.go
alexeykazakov: Yeah... I was thinking about using the dynamic client but we would still need to know the GVK to list objects... We could get all GVKs available in the API Server but it would be very time/resource consuming to get the list of all GVKs and then list the corresponding resources in the namespace.

> - store all revisions of the tiers in OS
> - download the older revision from GH

I would avoid depending on GitHub in runtime. Storing all the revisions in etcd should be pretty easy to implement and use.

PR #132, opened 2020-03-18
add limit ranges on namespace templates

xcoulon:
Added cluster role so the service account
is allowed to create LimitRanges on the user
namespaces.

Updates https://issues.redhat.com/browse/CRT-460

### DIFF
--- a/hack/deploy_csv.yaml
+++ b/hack/deploy_csv.yaml
@@ -54,20 +54,6 @@ data:
               spec:
                 description: NSTemplateSetSpec defines the desired state of NSTemplateSet
                 properties:
-                  clusterResources:
-                    description: the cluster resources template (for cluster-wide quotas,
-                      etc.)
-                    properties:
-                      revision:
-                        description: The revision of the corresponding template
-                        type: string
-                      template:
-                        description: Template contains an OpenShift Template to be used
-                          for provisioning of cluster-scoped resources
-                        type: string
-                    required:
-                    - revision
-                    type: object
                   namespaces:
                     description: The namespace templates
                     items:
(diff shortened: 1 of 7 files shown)


hack/deploy_csv.yaml
```diff
@@ -54,20 +54,6 @@ data:
               spec:
                 description: NSTemplateSetSpec defines the desired state of NSTemplateSet
                 properties:
-                  clusterResources:
```
xcoulon: I need to verify why this was removed.

MatousJobanek: did you generate the files while having api repo in master branch and with the latest commits pulled from the origin?

xcoulon: yes, but I regenerated these files using the `make generate-olm-files` command in the `member-operator` repository. But can this explain the problem?

xcoulon: hmm, running the `make generate` from the `codereadytoolchain/api` repo restored these `clusterResources` properties.

xcoulon: fixed in https://github.com/codeready-toolchain/member-operator/pull/132/commits/4ff21fe290068148ce95301340bde952057fd08b

MatousJobanek: hmm... weird. The `make generate-olm-files` should not take anything from the `api` repo yet :man_shrugging: (except for the script)

I tried it locally it it works fine.

xcoulon: indeed, I re-ran the `make generate-olm-files` and I did not get any change, which means that the `clusterResources` properties were not removed again. I have no idea what happened before 🤨

PR #142, opened 2020-04-03
adding provider label in nstemplateset_controller

tinakurian:
Adding the provider label to nstemplateset as well as to the namespace and innernamespace. 

They will be removed from basic, advanced and teams (https://github.com/codeready-toolchain/host-operator/blob/master/deploy/templates/nstemplatetiers/.... )

Jira story: https://issues.redhat.com/browse/CRT-495
Related e2e-test PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/99


pkg/controller/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -334,6 +334,23 @@ func TestReconcileProvisionOK(t *testing.T) {
 				HasLabel("toolchain.dev.openshift.com/type", "code")
 		})
 
+		t.Run("set_label_for_nstemplateset", func(t *testing.T) {
```
xcoulon, line 337: you don't need to use underscores in the test name

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -128,6 +132,24 @@ func (r *NSTemplateSetReconciler) addFinalizer(nsTmplSet *toolchainv1alpha1.NSTe
 	return nil
 }
 
+// addLabels sets the labels for NSTemplateSet
+func (r *NSTemplateSetReconciler) addLabels(nsTmplSet *toolchainv1alpha1.NSTemplateSet) error {
```
alexeykazakov, line 136: We don't need the provider key in the NSTemplateSet resources itself. We only need to add this label to all resources created by the NSTemplateSet.

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -302,6 +326,8 @@ func (r *NSTemplateSetReconciler) ensureInnerNamespaceResources(logger logr.Logg
 	}
 	namespace.Labels[toolchainv1alpha1.RevisionLabelKey] = tcNamespace.Revision
 	namespace.Labels[toolchainv1alpha1.TierLabelKey] = nsTmplSet.Spec.TierName
+	namespace.Labels[toolchainv1alpha1.ProviderLabelKey] = toolchainv1alpha1.ProviderLabelValue
```
alexeykazakov, line 329: You don't have to add this label here. It's already added in line 274: https://github.com/codeready-toolchain/member-operator/pull/142/files#diff-cffea82f7fdc7427b8c9f0d0fdd55aedR274

What you need to do instead is to add the label to all objects which are created inside the namespace. They are created in line 319: https://github.com/codeready-toolchain/member-operator/pull/142/files#diff-cffea82f7fdc7427b8c9f0d0fdd55aedL295
```
err = tmplProcessor.Apply(objs)
```

pkg/controller/nstemplateset/nstemplateset_controller_test.go
tinakurian: @xcoulon sorry! copied it from the tests below: https://github.com/codeready-toolchain/member-operator/pull/142/files/9a97ba0348d4e9206102f847c600c50e748a2a7b#diff-9b35dfc60ea38cb3e546642f4f9cedb4R583 

i've removed it from all the tests though - hope that's okay.

fixed here: 5cc8815b8c2ff4320c17a69bf55d5e7353b91b3f

pkg/controller/nstemplateset/nstemplateset_controller.go
tinakurian: @alexeykazakov i had another look at this code and the labels here actually belong to the objects prior to the apply. So in this case, the labels should be added just before the objects are created. `ensureNamespaceResource` actually ranges over the objects just the same way apply does so we can add labels to it. Maybe i misunderstood the code though?

tinakurian: @alexeykazakov, i think you are right that they don't need to be added in both places though. will start up a temp cluster and test the idea to see if the labels still get set!

tinakurian: @alexeykazakov i think it may be needed in both places - it looks like one func is called because namespace is nil and the other is called when namespace is not nil - is this correct? or did i miss something?

MatousJobanek: @tinakurian Please, try to read the code one more time and check in which cases the `ensureNamespaceResource` or `ensureInnerNamespaceResources` are called. What do we check there - as you said, we check if the namespace is nil or not - but what does it refer to? 
Take a look also at the logic inside of these methods - what they do (their names should give you a hint). 
Check what kind of methods of the client is probably used in these methods (if `client.Update` method is called or rather `client.Create`)

If you focus on these parts of the code, then you should be able to clearly understand the purpose. This will help you with answering the question if the label has to be added here as well or not.

tinakurian: @MatousJobanek @alexeykazakov fixed.. i understand... need to add a test for inner resources though..

test/namespace_assertion.go
```diff
@@ -65,11 +68,28 @@ func (a *NamespaceAssertion) HasNoLabel(key string) *NamespaceAssertion {
...
 
-func (a *NamespaceAssertion) HasResource(name string, obj runtime.Object) *NamespaceAssertion {
+func InnerResourceHasLabels(key, value string) ResourceCriterion {
+	return func(obj runtime.Object) bool {
+		// TODO: check the obj labels using the accessor
+		metaObj, _ := meta.Accessor(obj)
+
+		labels := metaObj.GetLabels()
+		v := labels[key]
+		if v == value {
```
golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

test/namespace_assertion.go
```diff
@@ -65,11 +68,25 @@ func (a *NamespaceAssertion) HasNoLabel(key string) *NamespaceAssertion {
 	return a
 }
 
-func (a *NamespaceAssertion) HasResource(name string, obj runtime.Object) *NamespaceAssertion {
+func InnerResourceHasLabels(key, value string) ResourceCriterion {
+	return func(obj runtime.Object) bool {
+		metaObj, _ := meta.Accessor(obj)
+		labels := metaObj.GetLabels()
+		if labels[key] == value {
```
golangcibot, line 75: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

golangcibot: S1008: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (from `gosimple`)

pkg/controller/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -1071,6 +1097,9 @@ func newNSTmplSet(namespaceName, name, tier string, options ...nsTmplSetOption)
 			Namespace:  namespaceName,
 			Name:       name,
 			Finalizers: []string{toolchainv1alpha1.FinalizerName},
+			Labels: map[string]string{
```
alexeykazakov, line 1100: You don't need this label in the NSTemplateSet resource, do you?

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -591,6 +591,9 @@ func newNSTemplateSet(userAcc *toolchainv1alpha1.UserAccount) *toolchainv1alpha1
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      userAcc.Name,
 			Namespace: userAcc.Namespace,
+			Labels: map[string]string{
```
alexeykazakov, line 594: As I mentioned before, the provider label is for objects created by the NSTempalateSet controller only. Not for the NSTemplateSet itself.

test/namespace_assertion.go
```diff
@@ -65,11 +68,22 @@ func (a *NamespaceAssertion) HasNoLabel(key string) *NamespaceAssertion {
 	return a
 }
 
-func (a *NamespaceAssertion) HasResource(name string, obj runtime.Object) *NamespaceAssertion {
+func InnerResourceHasLabels(key, value string) ResourceCriterion {
```
alexeykazakov, line 71:
```suggestion
func InnerResourceHasLabel(key, value string) ResourceCriterion {
```

test/nstemplateset_assertion.go
```diff
@@ -126,6 +126,14 @@ func (a *NSTemplateSetAssertion) HasFinalizer() *NSTemplateSetAssertion {
 	return a
 }
 
+func (a *NSTemplateSetAssertion) HasLabel(key, value string) *NSTemplateSetAssertion {
```
alexeykazakov, line 129: Not needed. See my comments above.

alexeykazakov:
BTW, please make sure our e2e tests already check the provider label. If not then please add such tests.

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -292,6 +295,22 @@ func (r *NSTemplateSetReconciler) ensureInnerNamespaceResources(logger logr.Logg
 		}
 	}
 
+	for _, rawObj := range objs {
+		acc, err := meta.Accessor(rawObj.Object)
+		if err != nil {
+			return r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusNamespaceProvisionFailed, err, "invalid element in template for namespace type '%s'", tcNamespace.Type)
```
MatousJobanek, line 301:
```suggestion
			return r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusNamespaceProvisionFailed, err, "unable to get meta.Interface of the object '%s' in the namespace '%s'", rawObj.Raw, namespace.Name)
```

pkg/controller/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -306,6 +306,7 @@ func TestReconcileProvisionOK(t *testing.T) {
 			assert.Equal(t, reconcile.Result{}, res)
 			AssertThatNSTemplateSet(t, namespaceName, username, fakeClient).
 				HasFinalizer().
+				HasLabel(toolchainv1alpha1.ProviderLabelKey, toolchainv1alpha1.ProviderLabelValue).
```
MatousJobanek, line 309: The NSTemplateSet is not supposed to have the label. You should verify it in the namespace.
```suggestion
```

pkg/controller/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -330,6 +331,7 @@ func TestReconcileProvisionOK(t *testing.T) {
...
```
MatousJobanek, line 334: same here: 
```suggestion
```

pkg/controller/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -354,14 +356,15 @@ func TestReconcileProvisionOK(t *testing.T) {
...
```
MatousJobanek, line 359: and here
```suggestion
```
and in all following cases

test/namespace_assertion.go
```diff
@@ -65,11 +68,22 @@ func (a *NamespaceAssertion) HasNoLabel(key string) *NamespaceAssertion {
...
+	return func(obj runtime.Object) bool {
+		metaObj, _ := meta.Accessor(obj)
+		labels := metaObj.GetLabels()
+		return labels[key] == value
+	}
+}
+func (a *NamespaceAssertion) HasResource(name string, obj runtime.Object, criteria ...ResourceCriterion) *NamespaceAssertion {
```
MatousJobanek, line 78: The label is expected to be present at all inner resources - I guess that we don't need to have optional criteria for that and we can make it as a mandatary check for all resources. 
In other words, we can drop the parameter `criteria` and move the check inside of this `HasResource` method. With that, it won't happen that you could miss to heck it in some cases - eg here: https://github.com/codeready-toolchain/member-operator/blob/f0052f4968415f8b942659b034b5e28018343c99/pkg/controller/nstemplateset/nstemplateset_controller_test.go#L480

tinakurian:
> BTW, please make sure our e2e tests already check the provider label. If not then please add such tests.

@alexeykazakov it looks like this may already be covered in our e2e-tests. 

in ```e2e_test.go``` 
verifyResourcesProvisionedForSignup-->VerifyNsTemplateSet-->tierChecks.GetInnerObjectChecks

MatousJobanek:
> it looks like this may already be covered in our e2e-tests.

You're partially correct - we check the expected number of resource types listed using the label, but we don't check it for the specific resources found by name - this is what is missing there.

test/namespace_assertion.go
```diff
@@ -3,17 +3,22 @@ package test
...
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type ResourceCriterion func(obj runtime.Object) bool
+
```
MatousJobanek, line 21: This is not used if I'm not mistaken
```suggestion
```

pkg/controller/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -1139,6 +1139,9 @@ func newRoleBinding(namespace, name string) *v1.RoleBinding {
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: namespace,
 			Name:      name,
+			Labels: map[string]string{
+				toolchainv1alpha1.ProviderLabelKey: toolchainv1alpha1.ProviderLabelValue,
```
MatousJobanek, line 1143: May I know why did you add the label here, and not in the `newRole` function? In both cases, the resource is created as part of the test setup.

test/namespace_assertion.go
tinakurian: good catch, fixed as part of: ad15c624078af2b4d102df7ed1304e6483044976

pkg/controller/nstemplateset/nstemplateset_controller_test.go
tinakurian: fixed as part of: ad15c624078af2b4d102df7ed1304e6483044976

tinakurian:
> @tinakurian thanks for removing the checks from the NSTemplateSet, but you forgot to add the check of the label for `Namespace` resource itself

yeah, forgot to add checks for this. i've added it as part of this commit: ad15c624078af2b4d102df7ed1304e6483044976

alexeykazakov:
> it looks like this may already be covered in our e2e-tests.

@tinakurian Couple days ago I added the provider label assertion to the edit-role check in the e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/blob/35378bf0a53d25ce3d73dfbd0121e0fb810207cb/tiers/checks.go#L160
We need to add it to all other checks for all resources such as 
https://github.com/codeready-toolchain/toolchain-e2e/blob/35378bf0a53d25ce3d73dfbd0121e0fb810207cb/tiers/checks.go#L131
https://github.com/codeready-toolchain/toolchain-e2e/blob/35378bf0a53d25ce3d73dfbd0121e0fb810207cb/tiers/checks.go#L118
https://github.com/codeready-toolchain/toolchain-e2e/blob/35378bf0a53d25ce3d73dfbd0121e0fb810207cb/tiers/checks.go#L164
https://github.com/codeready-toolchain/toolchain-e2e/blob/35378bf0a53d25ce3d73dfbd0121e0fb810207cb/tiers/checks.go#L194
https://github.com/codeready-toolchain/toolchain-e2e/blob/35378bf0a53d25ce3d73dfbd0121e0fb810207cb/tiers/checks.go#L225

tinakurian:
> > it looks like this may already be covered in our e2e-tests.
> 
> @tinakurian Couple days ago I added the provider label assertion to the edit-role check in the e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/blob/35378bf0a53d25ce3d73dfbd0121e0fb810207cb/tiers/checks.go#L160
> We need to add it to all other checks for all resources such as
> https://github.com/codeready-toolchain/toolchain-e2e/blob/35378bf0a53d25ce3d73dfbd0121e0fb810207cb/tiers/checks.go#L131
> https://github.com/codeready-toolchain/toolchain-e2e/blob/35378bf0a53d25ce3d73dfbd0121e0fb810207cb/tiers/checks.go#L118
> https://github.com/codeready-toolchain/toolchain-e2e/blob/35378bf0a53d25ce3d73dfbd0121e0fb810207cb/tiers/checks.go#L164
> https://github.com/codeready-toolchain/toolchain-e2e/blob/35378bf0a53d25ce3d73dfbd0121e0fb810207cb/tiers/checks.go#L194
> https://github.com/codeready-toolchain/toolchain-e2e/blob/35378bf0a53d25ce3d73dfbd0121e0fb810207cb/tiers/checks.go#L225

thanks @alexeykazakov, yeah, just noticed when i pulled the latest master for e2e-tests. will add the rest!

tinakurian:
> Looks good! Just waiting for the corresponding e2e PR before adding the final approval.

related e2e-test PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/99

PR #147, opened 2020-04-20
fix: list ClusterResourceQuotas by owner label

MatousJobanek:
While listing the existing ClusterResourceQuotas it was forgotten to use the ListOption for filtering by `owner` label. It shouldn't cause any issues, but it's better to use it and it makes the code more efficient.


xcoulon:
> While listing the existing ClusterResourceQuotas it was forgotten to use the ListOption for filtering by `owner` label. It shouldn't cause any issues, but it's better to use it and it makes the code more efficient.

Good catch indeed! 
It would have caused an issue if we had users with different tiers. Listing ClusterResourceQuotas without the `owner` label would have returned all quotas, included other users'. If the first quota returned by the client was set to another tier, we may have deleted the quotas and then recreated them again (to the expected level, no matter what we got before). At the end of the day, the quotas would be correct, but we may have deleted them and recreated them again for nothing.

PR #149, opened 2020-04-29
fix: NSTemplateSet tests should test a CRQ in basic tier

MatousJobanek:
this PR fixes a few things in the test:
* there was a wrong indentation in the ClusterResourceQuotas yaml snipets
* when using fake client it should rely on changes of the generation - based on this information it decides what to do next
* ~the basic tier should contain the ClusterResourceQuotas resource - it's more or less prerequisite of all tiers~
* when creating the ClusterResourceQuotas resource during the test setup, we need to parse the quota from yaml file, otherwise, the object contains slightly different formating that the one retrieved from the template - this causes that it looks like the resource was modified, thus generation incremented

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ require (
 	github.com/Azure/go-autorest/autorest v0.9.2 // indirect
 	github.com/Azure/go-autorest/autorest/adal v0.8.0 // indirect
 	github.com/codeready-toolchain/api v0.0.0-20200416071738-c52e81ffb85d
-	github.com/codeready-toolchain/toolchain-common v0.0.0-20200430074824-fb63cc12194b
+	github.com/codeready-toolchain/toolchain-common v0.0.0-20200504150510-6252228f3527
 	github.com/go-logr/logr v0.1.0
 	github.com/go-openapi/swag v0.19.9 // indirect
 	github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible
@@ -26,8 +26,6 @@ require (
 	sigs.k8s.io/controller-runtime v0.5.0
 	sigs.k8s.io/kubefed v0.1.0-rc6.0.20200224204536-6207193c49f7
 )
-
-replace github.com/codeready-toolchain/toolchain-common => github.com/matousjobanek/toolchain-common v0.0.0-20200504102205-05095b3cf41b
 
 // Pinned to kubernetes-1.16.2
 replace (
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/stretchr/testify v1.4.0
 	golang.org/x/net v0.0.0-20200324143707-d3edc9973b7e // indirect
 	golang.org/x/sys v0.0.0-20200413165638-669c56c373c4 // indirect
+	gopkg.in/yaml.v2 v2.2.8
 	k8s.io/api v0.18.1
 	k8s.io/apiextensions-apiserver v0.18.1
 	k8s.io/apimachinery v0.18.1
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ require (
 	github.com/Azure/go-autorest/autorest v0.9.2 // indirect
 	github.com/Azure/go-autorest/autorest/adal v0.8.0 // indirect
 	github.com/codeready-toolchain/api v0.0.0-20200416071738-c52e81ffb85d
-	github.com/codeready-toolchain/toolchain-common v0.0.0-20200427080755-4d31a88fd82c
+	github.com/codeready-toolchain/toolchain-common v0.0.0-20200430074824-fb63cc12194b
 	github.com/go-logr/logr v0.1.0
 	github.com/go-openapi/swag v0.19.9 // indirect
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -27,6 +27,8 @@ require (
 	sigs.k8s.io/kubefed v0.1.0-rc6.0.20200224204536-6207193c49f7
 )
 
+replace github.com/codeready-toolchain/toolchain-common => github.com/matousjobanek/toolchain-common v0.0.0-20200429153037-a2a22771697a
```
MatousJobanek, line 30: a note for myself !!!

pkg/controller/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -1912,8 +1939,8 @@ func getTemplateContent(decoder runtime.Decoder) func(tierName, typeName string)
 			}
 		case "basic":
 			switch typeName {
-			case ClusterResources: // assume that this tier has no "cluster resources" template
-				return nil, nil
+			case ClusterResources:
```
alexeykazakov, line 1942: We explicitly made ClusterResource optional: https://github.com/codeready-toolchain/api/blob/master/pkg/apis/toolchain/v1alpha1/nstemplatetier_types.go#L22
So, we needed a test which would make sure a tier without ClusterResource would work.

But I think we can actually make ClusterResource required but then let's fix the API.

MatousJobanek: ah, I totally forgot that it's optional.

> But I think we can actually make ClusterResource required but then let's fix the API.

sure, but until it's required, we should test also the optionality - let me fix that

MatousJobanek: done 5129524
I had to change the logic for creating the ClusterResourceQuotas resource during the test setup - we need to parse the quota from yaml file, otherwise, the object contains slightly different formating that the one retrieved from the template - this causes that it looks like the resource was modified, thus generation incremented

go.mod
MatousJobanek: done 99c4b85

pkg/controller/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -1722,19 +1723,42 @@ func prepareController(t *testing.T, initObjs ...runtime.Object) (*NSTemplateSet
 		if err != nil {
 			return err
 		}
-		return fakeClient.Client.Create(ctx, o, opts...)
+		innerFakeClient := test.NewFakeClient(t)
```
xcoulon, line 1726: what's the benefit of such an `innerFakeClient`? Why can't we just rely on the `fakeClient.Client`?

pkg/controller/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -1869,7 +1893,17 @@ func newRole(namespace, name string) *rbacv1.Role { //nolint: unparam
...
-func newClusterResourceQuota(username, tier string) *quotav1.ClusterResourceQuota {
+func newClusterResourceQuota(t *testing.T, username, tier string) *quotav1.ClusterResourceQuota {
+	quota := corev1.ResourceQuotaSpec{}
+	quotaContent := `
+quota:
+  hard:
+    limits.cpu: 1750m
+    limits.memory: 7Gi`
+
+	err := yaml.Unmarshal([]byte(quotaContent), &quota)
```
xcoulon, line 1904: could we try to see what's the result of this unmarshalling so we can keep the code as it was before, but maybe with adding the missing entries in the quota specs? (in line 1919 on the updated file)

pkg/controller/nstemplateset/nstemplateset_controller_test.go
MatousJobanek: no, because I want to use the logic that manages the generation number. Since this instance adds its own mocked function for `Create` and `Update` methods so the logic implemented in the fake client is not used and skipped

MatousJobanek: but I could move the logic out of the `Create` and `Update` methods of the fake client and make it available/usable via an exported function - that would also help you in your PR.

pkg/controller/nstemplateset/nstemplateset_controller_test.go
MatousJobanek: I made a wrong assumption while debugging failures in the test. Sorry for that - after another investigation, I figured out that the problems were somewhere else.
Brought back the original way of creating it in e49bb97
Thanks a lot for raising this concern :+1:

pkg/controller/nstemplateset/nstemplateset_controller_test.go
MatousJobanek: PR created: https://github.com/codeready-toolchain/toolchain-common/pull/102
changed & used in: 0ff927b

xcoulon: thanks, it's actually easier to understand without having to rely on a second client.

pkg/controller/nstemplateset/nstemplateset_controller_test.go
xcoulon: ok

PR #150, opened 2020-04-29
Allow member operator to be configured

rajivnathan:
### Related Issue(s)
https://issues.redhat.com/browse/CRT-216

### What does this PR do?
Allows the member operator to be configured via config map or environment variables in the same way that the host operator already does.

### Acceptance Criteria
- [x] Use the Viper library for loading configuration
- [x] The configuration can load (when needed)
    - [x] IdP (replace [current solution](https://github.com/codeready-toolchain/member-operator/blob/c681c6e208ddd7282b621dcf63809927e59e4cbb/pkg/config/env.go))
    - [x] KubeFedCluster healthcheck periods and timeout
- [x] As part of verification (testing), there is changed the wait timeout of KubeFedCluster e2e test.
- [x] Add unit tests

e2e tests PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/105


cmd/manager/main.go
```diff
@@ -171,6 +179,40 @@ func main() {
 	}
 }
 
+func loadConfig() (*configuration.Registry, error) {
```
xcoulon, line 182: what about moving this func in the `configuration` pkg?

cmd/manager/main.go
```diff
@@ -171,6 +179,40 @@ func main() {
...
+	var configFilePath string
+	flag.StringVar(&configFilePath, "config", "", "path to the config file to read (if none is given, defaults will be used)")
```
xcoulon, line 184: I don't think we need support for command line flags here

pkg/configuration/configuration.go
```diff
@@ -0,0 +1,94 @@
...
+	c.setConfigDefaults()
+	return &c
+}
+
+// New creates a configuration reader object using a configurable configuration
+// file path. If the provided config file path is empty, a default configuration
+// will be created.
+func New(configFilePath string) (*Registry, error) {
+	c := CreateEmptyRegistry()
+	if configFilePath != "" {
```
xcoulon, line 54: same here, I believe we can skip the support for an external config file. We'll just use config maps and env vars, AFAIK.

pkg/configuration/configuration.go
```diff
@@ -0,0 +1,94 @@
...
+	// IdentityProvider specifies an identity provider (IdP) for newly created users
+	IdentityProvider = "identity.provider"
+
+	// DefaultIdentityProvider the default value used for the identity provider (IdP) for newly created users
+	DefaultIdentityProvider = "rhd"
+)
+
+// Registry encapsulates the Viper configuration registry which stores the
+// configuration data in-memory.
+type Registry struct {
```
xcoulon, line 33: I'm the one who introduced this `Registry` type name a while ago (on OSIO) but I believe that `configuration.Configuration` would be more explicit, honestly (despite the repetition) 😬

pkg/configuration/configuration.go
```diff
@@ -0,0 +1,94 @@
...
+	member *viper.Viper
+}
+
+// CreateEmptyRegistry creates an initial, empty registry.
+func CreateEmptyRegistry() *Registry {
```
xcoulon, line 38: this func does not need to be public AFAICS

pkg/configuration/kubefed.go
```diff
@@ -0,0 +1,58 @@
+package configuration
```
xcoulon, line 1: why do you use a separate file? It's a bit confusing having methods on the `Registry` type in another file

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -598,6 +599,6 @@ func newNSTemplateSet(userAcc *toolchainv1alpha1.UserAccount) *toolchainv1alpha1
 }
 
 // ToIdentityName converts the given `userID` into an identity
-func ToIdentityName(userID string) string {
+func ToIdentityName(userID string, config *configuration.Registry) string {
```
xcoulon, line 602: here you could introduce a specific interface which has a single `GetIdP()` func, so it's more flexible in term of usage if we want to replace the `config` with a custom/mock object. (That's just a suggestion)

cmd/manager/main.go
rajivnathan: This was also taken from the host operator. I had the same thought but assumed it was there for a reason so was hesitant to remove it. Any idea why it's in the host operator?

pkg/configuration/configuration.go
rajivnathan: Appears so, good catch! 👍

cmd/manager/main.go
xcoulon: yes, I saw it's the same in the host operator, but I could not find any use-case with such a config file. So I guess we could remove it in that repo, too. Unless I'm missing something @alexeykazakov and @MatousJobanek?

alexeykazakov: Yeah.. I don't think we need to support loading the configuration from a file. But maybe @MatousJobanek had something in mind when introduced it.

pkg/configuration/configuration.go
rajivnathan: This is also something taken from the host operator. Are we sure we want to remove it? I can make the same change to the host operator when we've confirmed.

pkg/configuration/kubefed.go
rajivnathan: I thought it would be better to have the kubefed configuration separate and it seemed like a logical way to divide the configuration into pieces but not if it makes it more confusing. There isn't a ton of configuration right now so I moved it back and we can always revisit later if it does grow too large.

pkg/controller/useraccount/useraccount_controller.go
rajivnathan: I think that's a very good suggestion as it would allow us to reuse that function for any number of applications beyond just configuration but it looked a bit funny when I created a small interface that only returns a single string and the interface is only used in this one place at the moment. Do we foresee it being reused elsewhere? If not, should wait until we need to use it somewhere else or want to expand on it?

pkg/configuration/configuration.go
rajivnathan: Haha, agree. I renamed it from `Registry` to `Config` so we can keep it short.

pkg/controller/useraccount/useraccount_controller.go
xcoulon: it's just that having a simple interface here makes it easier to test. You could provide a mock that returns a specific value for your test case, without having to deal with the initialization of a full-size configuration object.

cmd/manager/main.go
MatousJobanek: Yeah, we can remove it... To be honest, I'm not sure why I added the support for loading from files - it was probably just a result of a blind "copy-paste driven development" ¯\_(ツ)_/¯

cmd/manager/main.go
MatousJobanek: nope, we don't use it so feel free to drop it from the host-operator as well ;-)

pkg/configuration/configuration.go
MatousJobanek: yes, we don't need to load the config from the files - you can drop it. 
Thanks and sorry for bringing the confusions :-)

pkg/controller/useraccount/useraccount_controller.go
rajivnathan: Good point, I'll make the change

cmd/manager/main.go
rajivnathan: Removed as part of 6d417df27d5ddb043b10285770c7a37c5b959286

pkg/configuration/configuration.go
rajivnathan: Removed as part of a0acc7e66f8b3bca14fd5cad1631b966e28b399d

pkg/controller/useraccount/useraccount_controller.go
rajivnathan: Done in 69cd855580034f247065ba24ca5e4ffc673acfeb

rajivnathan:
I've addressed the review comments, added unit tests and configured the wait timeout for the KubeFed cluster period for the e2e tests environment. @xcoulon please take a look at your convenience. Thanks again for the review! 🙂 

The corresponding change for the e2e tests is in https://github.com/codeready-toolchain/toolchain-e2e/pull/105

pkg/configuration/configuration.go
```diff
@@ -0,0 +1,135 @@
...
+	return vars
+}
+
+// GetIdP returns the configured Identity Provider (IdP) for the member operator
+// Openshift clusters can be configured with multiple IdPs. This config option allows admins to specify which IdP should be used by the toolchain operator.
+func (c *Config) GetIdP() string {
+	return c.member.GetString(IdentityProvider)
+}
+
+// GetClusterHealthCheckPeriod returns the configured member cluster health check period
```
MatousJobanek, line 107: It's just a detail, so it's more FYI - the comment that you wrote may be understood as it is a period for member cluster health check in the sense that we want to check the member cluster, but it's vice versa. The period is used for checking a host cluster from the member cluster - in other words, the period is giving duration after which the Kubefed should check if the host cluster is ready or not. Similarly for the following methods.

pkg/configuration/configuration_test.go
```diff
@@ -0,0 +1,211 @@
...
+		params := config.GetAllMemberParameters()
+		require.Empty(t, params)
+	})
+	t.Run("IdP environment variable", func(t *testing.T) {
+		key := "MEMBER_OPERATOR_IDENTITY_PROVIDER"
+		resetFunc := test.UnsetEnvVarAndRestore(t, key)
+		defer resetFunc()
+		u, err := uuid.NewV4()
+		require.NoError(t, err)
+		os.Setenv(key, u.String())
```
MatousJobanek, line 43: the `os.Setenv` function returns an error, so you should handle it. But it would be maybe better to use `test.SetEnvVarAndRestore` function that gives you a function that will restore the original value back. you could than also drop the `test.UnsetEnvVarAndRestore` from the line 39

pkg/configuration/configuration_test.go
```diff
@@ -0,0 +1,211 @@
...
+		config := getDefaultConfiguration(t)
+		assert.Equal(t, "rhd", config.GetIdP())
+	})
+
+	t.Run("env overwrite", func(t *testing.T) {
+		restore := test.SetEnvVarAndRestore(t, key, "testingIdP")
+		defer restore()
+
+		restore = test.SetEnvVarAndRestore(t, MemberEnvPrefix+"_"+"ANY_CONFIG", "20s")
+		defer restore()
```
MatousJobanek, line 69: you could use:
```suggestion
		restore := test.SetEnvVarsAndRestore(t,
			test.Env(key, "testingIdP"),
			test.Env(MemberEnvPrefix+"_"+"ANY_CONFIG", "20s"))
		defer restore()
```

pkg/configuration/configuration_test.go
```diff
@@ -0,0 +1,211 @@
...
+		config := getDefaultConfiguration(t)
+		assert.Equal(t, cast.ToDuration("10s"), config.GetClusterHealthCheckPeriod())
+	})
+
+	t.Run("env overwrite", func(t *testing.T) {
+		restore := test.SetEnvVarAndRestore(t, key, "30s")
+		defer restore()
+
+		restore = test.SetEnvVarAndRestore(t, MemberEnvPrefix+"_"+"ANY_CONFIG", "20s")
+		defer restore()
```
MatousJobanek, line 92: same here
```suggestion
		restore := test.SetEnvVarsAndRestore(t,
			test.Env(key, "30s"),
			test.Env(MemberEnvPrefix+"_"+"ANY_CONFIG", "20s"))
		defer restore()
```

pkg/configuration/configuration_test.go
```diff
@@ -0,0 +1,211 @@
...
+		config := getDefaultConfiguration(t)
+		assert.Equal(t, cast.ToDuration("3s"), config.GetClusterHealthCheckTimeout())
+	})
+
+	t.Run("env overwrite", func(t *testing.T) {
+		restore := test.SetEnvVarAndRestore(t, key, "30s")
+		defer restore()
+
+		restore = test.SetEnvVarAndRestore(t, MemberEnvPrefix+"_"+"ANY_CONFIG", "20s")
+		defer restore()
```
MatousJobanek, line 115: and here
```suggestion
		restore := test.SetEnvVarsAndRestore(t,
			test.Env(key, "30s"),
			test.Env(MemberEnvPrefix+"_"+"ANY_CONFIG", "20s"))
		defer restore()
```

pkg/configuration/configuration_test.go
```diff
@@ -0,0 +1,211 @@
...
+		config := getDefaultConfiguration(t)
+		assert.Equal(t, int64(3), config.GetClusterHealthCheckFailureThreshold())
+	})
+
+	t.Run("env overwrite", func(t *testing.T) {
+		restore := test.SetEnvVarAndRestore(t, key, "5")
+		defer restore()
+
+		restore = test.SetEnvVarAndRestore(t, MemberEnvPrefix+"_"+"ANY_CONFIG", "20")
+		defer restore()
```
MatousJobanek, line 138: and here:
```suggestion
		restore := test.SetEnvVarsAndRestore(t,
			test.Env(key, "5"),
			test.Env(MemberEnvPrefix+"_"+"ANY_CONFIG", "20"))
		defer restore()
```

pkg/configuration/configuration_test.go
```diff
@@ -0,0 +1,211 @@
...
+		config := getDefaultConfiguration(t)
+		assert.Equal(t, int64(1), config.GetClusterHealthCheckSuccessThreshold())
+	})
+
+	t.Run("env overwrite", func(t *testing.T) {
+		restore := test.SetEnvVarAndRestore(t, key, "3")
+		defer restore()
+
+		restore = test.SetEnvVarAndRestore(t, MemberEnvPrefix+"_"+"ANY_CONFIG", "20")
+		defer restore()
```
MatousJobanek, line 161:
```suggestion
		restore := test.SetEnvVarsAndRestore(t,
			test.Env(key, "3"),
			test.Env(MemberEnvPrefix+"_"+"ANY_CONFIG", "20s"))
		defer restore()
```

pkg/configuration/configuration_test.go
```diff
@@ -0,0 +1,211 @@
...
+		config := getDefaultConfiguration(t)
+		assert.Equal(t, cast.ToDuration("20s"), config.GetClusterAvailableDelay())
+	})
+
+	t.Run("env overwrite", func(t *testing.T) {
+		restore := test.SetEnvVarAndRestore(t, key, "30s")
+		defer restore()
+
+		restore = test.SetEnvVarAndRestore(t, MemberEnvPrefix+"_"+"ANY_CONFIG", "40s")
+		defer restore()
```
MatousJobanek, line 184:
```suggestion
		restore := test.SetEnvVarsAndRestore(t,
			test.Env(key, "30s"),
			test.Env(MemberEnvPrefix+"_"+"ANY_CONFIG", "40s"))
		defer restore()
```

pkg/configuration/configuration_test.go
```diff
@@ -0,0 +1,211 @@
...
+		config := getDefaultConfiguration(t)
+		assert.Equal(t, cast.ToDuration("60s"), config.GetClusterUnavailableDelay())
+	})
+
+	t.Run("env overwrite", func(t *testing.T) {
+		restore := test.SetEnvVarAndRestore(t, key, "30s")
+		defer restore()
+
+		restore = test.SetEnvVarAndRestore(t, MemberEnvPrefix+"_"+"ANY_CONFIG", "20s")
+		defer restore()
```
MatousJobanek, line 207:
```suggestion
		restore := test.SetEnvVarsAndRestore(t,
			test.Env(key, "30s"),
			test.Env(MemberEnvPrefix+"_"+"ANY_CONFIG", "20s"))
		defer restore()
```

pkg/controller/useraccount/useraccount_controller.go
MatousJobanek: just one thing that I noticed - we have this fancy new interface but is not used anywhere, so do we really need it? The tests still rely on and use the config without any problem, so I'm missing the benefit of the interface. 

I'm not saying that you should remove it, I'm just sharing my observation :wink: You can keep it if you want :-)

cmd/manager/main.go
```diff
@@ -171,6 +175,19 @@ func main() {
 	}
 }
 
+func printConfig(cfg *configuration.Config) {
+	logWithValuesMemberOperator := log
+	for key, value := range cfg.GetAllMemberParameters() {
+		logWithValuesMemberOperator = logWithValuesMemberOperator.WithValues("key", key, "value", value)
```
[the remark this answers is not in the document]
rajivnathan: Actually, `getMemberEnvVarKey` was copied by accident. I meant to log the values in the same way that the registration config is logged here: https://github.com/codeready-toolchain/host-operator/blob/d764a1c431ef2ad964ca374dbb34100f8d094c0b/cmd/manager/main.go#L233

So I will just remove the `getMemberEnvVarKey` function since it's not needed.

pkg/configuration/configuration_test.go
rajivnathan: Good suggestion, thanks!

pkg/configuration/configuration_test.go
rajivnathan: Cool! Didn't know about that. Thanks!

pkg/configuration/configuration_test.go
rajivnathan: Thanks!

pkg/configuration/configuration_test.go
rajivnathan: Thanks!

pkg/configuration/configuration_test.go
rajivnathan: Thanks!

pkg/configuration/configuration_test.go
rajivnathan: Thanks!

pkg/configuration/configuration_test.go
rajivnathan: Thanks!

pkg/configuration/configuration_test.go
rajivnathan: Thanks!

cmd/manager/main.go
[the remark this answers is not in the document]
rajivnathan: Done as part of ddbbefc0a4adaa34d70562e367889177f2cb83a3

pkg/configuration/configuration.go
```diff
@@ -0,0 +1,135 @@
...
+// prefixes
+const (
+	// MemberEnvPrefix will be used for member environment variable name prefixing.
+	MemberEnvPrefix = "MEMBER_OPERATOR"
+)
+
+// Configuration constants
+const (
+	// ToolchainConfigMapName specifies a name of a ConfigMap that keeps toolchain configuration
+	ToolchainConfigMapName = "toolchain-saas-config"
```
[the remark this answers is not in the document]
rajivnathan: Removed as part of ddbbefc0a4adaa34d70562e367889177f2cb83a3

pkg/configuration/configuration.go
rajivnathan: Done as part of ddbbefc0a4adaa34d70562e367889177f2cb83a3

pkg/configuration/configuration_test.go
rajivnathan: Done as part of ddbbefc0a4adaa34d70562e367889177f2cb83a3

rajivnathan:
@MatousJobanek I believe I've addressed the previous review comments, please take another look at your convenience. Thanks! 🙂

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -846,11 +847,13 @@ func TestDisabledUserAccount(t *testing.T) {
 	s := scheme.Scheme
 	err := apis.AddToScheme(s)
 	require.NoError(t, err)
+	config := configuration.LoadConfig()
+	require.NoError(t, err)
```
MatousJobanek, line 851: this line is redundant - the error was already checked and `configuration.LoadConfig()` doesn't return an error
```suggestion
```

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -1217,10 +1220,13 @@ func prepareReconcile(t *testing.T, username string, initObjs ...runtime.Object)
 	err := apis.AddToScheme(s)
 	require.NoError(t, err)
 	fakeClient := test.NewFakeClient(t, initObjs...)
+	config := configuration.LoadConfig()
+	require.NoError(t, err)
```
MatousJobanek, line 1224: same here
```suggestion
```

pkg/controller/useraccount/useraccount_controller_test.go
rajivnathan: Oops 😱, thanks for catching that

PR #155, opened 2020-05-13
feat: support other cluster scoped resources than CRQ

MatousJobanek:
Jira issue: https://issues.redhat.com/browse/CRT-461

This PR changes the logic that creates/updates/deletes cluster resources so it is able to create/update/delete also other kinds of cluster scoped resources than only `ClusterResourceQuota` - for now, it supports a new kind `ClusterRoleBinding`.

The main difference is that we keep a list of such cluster-scoped kinds and the code generically:
* watches all kinds
* creates/updates/deletes one resource of the kind per one reconcile

This is achieved with a help of a new type called `clusterResourceKind`. The type contains/provides:
* GVK of the kind
* an instance of the object
* a function that returns a list of existing resources (of the kind) for the given user

The logic of the code is simple - it goes through all kinds, lists the existing ones, and compares them with objects taken from the template. If any has to be created/updated or deleted, then it executes the action for one resource per one reconcile.


alexeykazakov:
> The type mainClusterResource represents all cluster-scoped resources that should exist in all "cluster" templates.

So, maybe `commonClusterResource` would be a better name then? WDYT?

alexeykazakov:
> So, maybe commonClusterResource would be a better name then? WDYT?

Actually, maybe it would be better to name them `watchedClusterResources`?

alexeykazakov:
Also, shouldn't we actually watch all the cluster resources regardless of the fact if they are in all templates or not?
I like the idea of having the explicit list of "watched" kinds. And we have to update that list when adding "common" resources to the templates anyway. But why just not make all the cluster resources watched? Even if they are defined on some tiers/templates only.

MatousJobanek:
> Actually, maybe it would be better to name them watchedClusterResources?

great idea - rename in da6bb95

MatousJobanek:
> Also, shouldn't we actually watch all the cluster resources regardless of the fact if they are in all templates or not?

There are a few reasons why I decided to support also non-watched resources:
* the templates are placed in host operator, but this logic is in member, so we cannot dynamically watch for all resource types from the cluster-resources template (as we do for the registration-service template), because we are not able to catch the changes in the template
* in case we want to watch for some new cluster-scoped resource types, we would need to first adjust the member operator and then add the resource into the template - this can be sometimes irritating
* I need to rely on specific resource types that exist in all templates - based on the existence of these resources and their labels I can detect what is going on - if the resources are being provisioned, updated, or upgraded. 
* for me it looks to be more reliable to have one or two resource types that are main/watched and based on which I can decide what to do next  - some of them exist in all templates and some not
* when we provision resources in the namespaces, we don't watch them either - we watch only the namespace resource - so I believe that we can use the same approach here
* and still, if we introduce some resource that exists in all templates and we want to watch it, we can easily add it to the list of the watched resources. But it still gives us the flexibility of adding new ones without changing the list.

MatousJobanek:
@alexeykazakov @xcoulon I refactored the code and simplified the logic - now it supports only cluster-scoped resources that are being watched by the NSTemplateSet controller. I also changed some names there  - I modified the description of the PR accordingly.
Could you please take a look when you find some time? Thanks a ton :+1:

MatousJobanek:
Thanks for your review @alexeykazakov and for your suggestion. :raised_hands: 

I agree with the proposal for having a  function that would help us with comparing the resources of the same resource kind. :+1:  That was kind of the same idea that I had when we were discussing the approaches. But I'd like to at first check how and if the KubeFed does comparison - they need to have some generic function as well.

> `kindType                 runtime.Object // Rename object to type to avoid confusion!`

good point - I'll rename it :+1: 

> When we need to add a new resource kind to our templates (namespace or cluster) we just add a new templateResourceKind to the corresponding list. No other code changes needed.

that is not true - you need to create `getResourceList` function for that resource kind as well - there is no generic object that you could use for listing the existing resources.

> Introduce templateObject. Can be used for both cluster and namespace resources to abstract actual objects and all kind specific functions.

I'm slightly missing the purpose of the `templateObject` type. I understand the idea behind the abstraction of the objects retrieved from the template, but I don't understand why adding the specific functions to it. I thought that the set of created/watched resource kinds should be given by a  hard-coded list in the code, not by the objects retrieved from the template. Maybe I don't understand how you want to use it. 
The flow that I use now is:
* We have a set of the watched resource kinds (that the controller can manage) hard-coded in the 
array
* For each GVK we do:
1. We list all existing resources
2. we get (by filtering for the same GVK) all objects from the template
3. we compare them 
4. we created/update/delete any of the resources if necessary

here you can see that the function for listing existing resources shouldn't be retrieved from the objects in the template (for example if the template doesn't contain any object of the GVK).

But it's a great idea for having the abstraction - I would maybe use something like this:
```go
type namedObject struct {
	object runtime.Object
	name   string
	gvk    schema.GroupVersionKind
}
```
this type would be returned by both:
* the `Process` method that returns template objects
* the `listExistingResources` function that returns existing objects

with that type we won't have to get the name of the object by using the `meta.Accessor` over and over again - it will make the code much shorter and cleaner :slightly_smiling_face:

alexeykazakov:
>> When we need to add a new resource kind to our templates (namespace or cluster) we just add a new templateResourceKind to the corresponding list. No other code changes needed.

>that is not true - you need to create getResourceList function for that resource kind as well - there is no generic object that you could use for listing the existing resources.

Sure. This is exactly what I meant. We would need to add a new `templateResourceKind` with a new `tempalateClient` implementation with `getResourceList`, `compare`, etc function specific to that kind. My point was that is just one place where we add it. So, it's pretty straight forward.

> I'm slightly missing the purpose of the templateObject type

Maybe `templateObject` is not the best name for that wrapper... The idea is to work with some abstraction/wrapper instead of the runtime objects when working with both current objects (which we currently have in cluster/namespace and which are managed via templates) and the new one which are currently listed in the template. The main purpose of the abstraction to keep the kind specific client (templateClient with ListResources, compare, etc) in that object.

So, that would work in the same why as now but with one slight modification:

- We have a set of the watched resource kinds (that the controller can manage) hard-coded in the
array
- For each GVK we do:
1. We list all existing resources (using `templateObject.kind.client.getResourceList() which returns a list of current/existing objects as types of `templateObject`/abstractions)
2. we get (by filtering for the same GVK) all objects from the template
3. we compare them (using `currentTemplateObject.kind.client.compare(templateObjectFromTemplate, currentTemplateObjectFromClusterOrNamespace)`)
4. we created/update/delete any of the resources if necessary (using `templateObject.kind.client.update/create/delete` if kind specific function is needed)

Basically we just wrap the client function with gvk specific implementations (via the corresponding `templateClient`) and always delegate list/compare/create/etc to that client instead of working with raw runtime objects in the main reconcile loop functions.

We could talk about that tomorrow on the sync call if it's still unclear what I mean here.

MatousJobanek:
@alexeykazakov so I added some abstraction there 87f0db7 
The main types are added here: https://github.com/codeready-toolchain/toolchain-common/pull/109 - I had to move it to toolchain-common because I need to make `ApplyClient` and the `Processor` compatible with it. 

I introduced `ToolchainObject` and `ComparableToolchainObject`. 
* `ToolchainObjects` are returned by the `Processor` that processes templates. 
* `ComparableToolchainObjects` are returned by the function that lists all existing resources in the cluster - it can be compared with any instance of `ToolchainObject`.

Apart from that, there is also `toolchainObjectKind` - it's more or less just renamed the original `clusterResourceKind`. I kept it here in host-operator cluster and didn't move it to toolchain-common as it didn't seem to make any sense to have it there. If you think that it would be better to move it there as well, I can do it ;-). 

I couldn't use your proposed structure of types because:
* I couldn't assign `templateResourceKind` to the objects retrieved from the processed templates. The type of the object is `Unstructured`, but  GVK would contain the actual values taken from the template, so it wouldn't match with the `kindType`. In addition, keeping `kindType` for the `Unstructured` wouldn't make any sense
* I couldn't assign any `templateResourceClient` to the objects retrieved from the processed templates either. As I said, the type is `Unstructured`, so listing the existing objects doesn't make any sense. Nor comparing.
* keeping the `listExistingResources` function for the actual instances of the object wouldn't make any sense. we need to go through all kinds we are interested in, not through all objects taken from the template. Related is:
> We list all existing resources (using templateObject.kind.client.getResourceList() which returns a list of current/existing objects as types of templateObject`/abstractions)

don't forget that in your design `templateObject` already represents an existing resource - but we should go through the lists of kinds instead.


But, as for your comment:
>  The main purpose of the abstraction to keep the kind specific client (templateClient with ListResources, compare, etc) in that object.

I understand your point. But instead of storing everything in one type, I created `ComparableToolchainObject` that contains the function that will be used for comparisons. The function is provided during the initialization of the object inside of the `listExistingResources` from `toolchainObjectKind` - that means that the kind specific logic is kept in one place - inside of `toolchainObjectKind` type

MatousJobanek:
and yeah, I didn't introduce the new types in separated PR because most of the features as well as of the changes are placed in this PR. Doing another PR just for these new types would require a lot of unnecessary work and tons of conflicts.

pkg/controller/nstemplateset/cluster_resources.go
```diff
@@ -90,73 +168,138 @@ func (r *clusterResourcesManager) ensure(logger logr.Logger, nsTmplSet *toolchai
...
+		if !isUpToDateAndProvisioned(currentObject, tierTemplate) {
+			for _, newObject := range newObjs {
+				if newObject.HasSameName(currentObject) {
+					// is found then let's update it
+					if err := r.setStatusUpdatingIfNotProvisioning(nsTmplSet); err != nil {
+						return false, err
+					}
+					return r.apply(logger, nsTmplSet, tierTemplate, newObject)
+				}
 			}
```
alexeykazakov, line 205: WDYT about something like this:
```suggestion
		for _, newObject := range newObjs {
			if newObject.HasSameName(currentObject) {
				// is found then let's check if we need to update it
				if !isUpToDate(currentObject, newObject, tierTemplate) {
					// let's update it
					if err := r.setStatusUpdatingIfNotProvisioning(nsTmplSet); err != nil {
						return false, err
					}
					return r.apply(logger, nsTmplSet, tierTemplate, newObject)
				}
			}
		}

...
// isUpToDate returns true if the currentObject uses the corresponding templateRef and equals to the new object
func isUpToDate(currentObject, newObject applycl.ToolchainObject, tierTemplate *tierTemplate) bool {
	return currentObject.GetLabels()[toolchainv1alpha1.TemplateRefLabelKey] != "" &&
		currentObject.GetLabels()[toolchainv1alpha1.TemplateRefLabelKey] == tierTemplate.templateRef
		// && currentObject.IsSame(newObject)  <-- TODO Uncomment wheh IsSame is implemented for all ToolchainObjects!
}

```

pkg/controller/nstemplateset/cluster_resources.go
```diff
@@ -2,82 +2,160 @@ package nstemplateset
...
+type listExistingResources func(cl client.Client, username string) ([]applycl.ComparableToolchainObject, error)
+
+// toolchainObjectKind represents a resource kind that should be present in templates containing cluster resources.
+// Such a kind should be watched by NSTempalateSet controller which means that every change of the
+// resources of that kind triggers a new reconcile.
+// It is expected that the template contains only the kinds that are being watched and there is an instance of
+// toolchainObjectKind type created in clusterResourceKinds list
+type toolchainObjectKind struct {
+	gvk                   schema.GroupVersionKind
+	object                runtime.Object
```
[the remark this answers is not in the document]
MatousJobanek: I kept the original `object` because I didn't combine the type with other types/objects, so I thought that there shouldn't be any confusion.
Anyway, I renamed it in: 317e5b0

pkg/controller/nstemplateset/cluster_resources.go
MatousJobanek: changed in ef2a0df

PR #156, opened 2020-05-14
 feat(CRT-501): migrate NSTemplateSet to templateRef field 

MatousJobanek:
This PR migrates the member-operator logic to the templateRef ield in NSTemplateSet. It stops using the legacy ones. 
In additon, it replaces both labels `/tier` and `/revision` with a new one `/templateref`

api PR: https://github.com/codeready-toolchain/api/pull/136
paired PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/110

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,6 +31,8 @@ require (
 	sigs.k8s.io/kubefed v0.1.0-rc6.0.20200224204536-6207193c49f7
 )
 
+replace github.com/codeready-toolchain/api => ../api
+
 // Pinned to kubernetes-1.16.2
 replace (
 	k8s.io/api => k8s.io/api v0.0.0-20191016110408-35e52d86657a
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	sigs.k8s.io/kubefed v0.1.0-rc6.0.20200224204536-6207193c49f7
 )
 
-replace github.com/codeready-toolchain/api => github.com/MatousJobanek/api v0.0.0-20200514221709-6828f4a4090a
+replace github.com/codeready-toolchain/api => github.com/MatousJobanek/api v0.0.0-20200515110108-4f5f9ccea1e9
 
 // Pinned to kubernetes-1.16.2
 replace (
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	sigs.k8s.io/kubefed v0.1.0-rc6.0.20200224204536-6207193c49f7
 )
 
-replace github.com/codeready-toolchain/api => ../api
+replace github.com/codeready-toolchain/api => github.com/MatousJobanek/api v0.0.0-20200514221709-6828f4a4090a
 
 // Pinned to kubernetes-1.16.2
 replace (
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/codeready-toolchain/member-operator
 require (
 	github.com/Azure/go-autorest/autorest v0.9.2 // indirect
 	github.com/Azure/go-autorest/autorest/adal v0.8.0 // indirect
-	github.com/codeready-toolchain/api v0.0.0-20200506130228-a821d9c227f1
+	github.com/codeready-toolchain/api v0.0.0-20200519085052-34d2a9c4a8e9
 	github.com/codeready-toolchain/toolchain-common v0.0.0-20200504150510-6252228f3527
 	github.com/go-logr/logr v0.1.0
 	github.com/go-openapi/swag v0.19.9 // indirect
@@ -30,8 +30,6 @@ require (
 	sigs.k8s.io/controller-runtime v0.5.0
 	sigs.k8s.io/kubefed v0.1.0-rc6.0.20200224204536-6207193c49f7
 )
-
-replace github.com/codeready-toolchain/api => github.com/MatousJobanek/api v0.0.0-20200515110108-4f5f9ccea1e9
 
 // Pinned to kubernetes-1.16.2
 replace (
--- a/pkg/controller/nstemplateset/cluster_resources_test.go
+++ b/pkg/controller/nstemplateset/cluster_resources_test.go
@@ -312,6 +312,7 @@ func TestPromoteClusterResources(t *testing.T) {
 			AssertThatCluster(t, cl).
 				HasResource("for-"+username, &quotav1.ClusterResourceQuota{},
 					WithLabel("toolchain.dev.openshift.com/templateref", "team-cluster-12345bb"),
+					WithLabel("toolchain.dev.openshift.com/tier", "team"),
 					Containing(`"limits.cpu":"4","limits.memory":"15Gi"`))
 		})
 
@@ -371,6 +372,7 @@ func TestPromoteClusterResources(t *testing.T) {
 			AssertThatCluster(t, cl).
 				HasResource("for-"+username, &quotav1.ClusterResourceQuota{},
 					WithLabel("toolchain.dev.openshift.com/templateref", "advanced-cluster-12345bb"),
+					WithLabel("toolchain.dev.openshift.com/tier", "advanced"),
 					Containing(`"limits.cpu":"2","limits.memory":"10Gi"`)) // upgraded
 		})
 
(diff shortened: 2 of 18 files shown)


pkg/controller/nstemplateset/cluster_resources_test.go
```diff
@@ -370,13 +371,14 @@ func TestPromoteClusterResources(t *testing.T) {
...
 				HasResource("for-"+username, &quotav1.ClusterResourceQuota{},
+					WithLabel("toolchain.dev.openshift.com/templateref", "advanced-cluster-12345bb"),
 					WithLabel("toolchain.dev.openshift.com/tier", "advanced"),
 					Containing(`"limits.cpu":"2","limits.memory":"10Gi"`)) // upgraded
 		})
 
 		t.Run("no redundant cluster resource quota to be deleted for the given user", func(t *testing.T) {
 			// given
-			nsTmplSet := newNSTmplSet(namespaceName, username, "advanced", withConditions(Provisioned())) // no cluster resources, so the "advancedCRQ" should be deleted
+			nsTmplSet := newNSTmplSet(namespaceName, username, "advanced", withConditions(Provisioned()), withClusterResources())
```
xcoulon, line 381: you're changing the initial state of the test here. In this case, we are supposed to deal with an `NSTemplateSet` in which no `clusterResources` are defined.  So should we add a test where there is no `clusterResources` at all, or are we going to drop this case?

pkg/controller/nstemplateset/nstemplatetier.go
```diff
@@ -56,26 +46,30 @@ func getTemplatesFromHost(hostClusterFunc cluster.GetHostClusterFunc, tierName s
...
+	return tierTemplate, nil
+}
+
+// tierTemplate contains all data from TierTemplate including its name
+type tierTemplate struct {
+	templateRef string
+	tierName    string
+	typeName    string
+	revision    string
+	template    templatev1.Template
```
xcoulon, line 66: It looks like you don't need the `revision` field (other fields are used, AFAICS).

pkg/controller/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -628,29 +633,33 @@ func withConditions(conditions ...toolchainv1alpha1.Condition) nsTmplSetOption {
...
 	return ns
 }
 
-type namespaceOption func(*corev1.Namespace)
+type namespaceOption func(ns *corev1.Namespace, tier string, typeName string)
 
-func withRevision(revision string) namespaceOption { // nolint: unparam
-	return func(ns *corev1.Namespace) {
-		ns.ObjectMeta.Labels["toolchain.dev.openshift.com/revision"] = revision
+func withTemplateRef(revision string) namespaceOption { // nolint: unparam
```
xcoulon, line 660: I believe the name of this func is a bit confusing. I was reviewing the test and commenting on all occurrences to say that the value was incorrect, until I reached here 🤦 

May I suggest that you rename this func, so we know that the argument is a revision and not a full templateRef value? Although, the result is an added `toolchain.dev.openshift.com/templateref` label on the namespace... 🤔 

So what about `withTemplateRevision(string)` instead?

pkg/controller/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -706,37 +716,40 @@ func newClusterResourceQuota(username, tier string) *quotav1.ClusterResourceQuot
...
+		typeName := refParts[1]
 		if typeName == "fail" || tierName == "fail" {
 			return nil, fmt.Errorf("failed to retrieve template")
 		}
 		var tmplContent string
 		switch tierName {
 		case "advanced": // assume that this tier has a "cluster resources" template
 			switch typeName {
-			case ClusterResources:
+			case "cluster":
```
xcoulon, line 731: we use `clusterresources` for the templateRef value. Should we use the same value here for the sake of consistency?

MatousJobanek: wait, but in host operator there are files named `cluster.yaml` so I thought that the name should be the same - it would be good to unify it so we avoid such confusion.
What is the proper naming then?

pkg/controller/nstemplateset/nstemplateset_controller_test.go
MatousJobanek: but the function doesn't add template `/revision` label to the namespace, but `/templateRef` label - then the name `withTemplateRevision` doesn't make sense

pkg/controller/nstemplateset/nstemplatetier.go
MatousJobanek: yeah, I copied all values from the `TierTemplate`

pkg/controller/nstemplateset/cluster_resources_test.go
MatousJobanek: well, the original test was wrong - look at the flow - it doesn't expect that anything should be deleted, but there is a cluster resource owned by the user and with `advanced` so it should actually expect that the resources should be deleted.

> So should we add a test where there is no clusterResources at all, or are we going to drop this case?

There are is a test where  `clusterResources` template is  not set, but I can add one more for this specific case

pkg/controller/nstemplateset/nstemplateset_controller_test.go
xcoulon: > but the function doesn't add template /revision label to the namespace, but /templateRef label - then the name withTemplateRevision doesn't make sense

ok, but having `withTemplateRef("123456a")` looks confusing IMO. And I suggested `withTemplateRevision` and not `withRevision` because I know that it's not going to be a `/revision` label which is added on the namespace.

so maybe simplifying the func and pass the actual `/templateref` value as-is?

pkg/controller/nstemplateset/nstemplateset_controller_test.go
xcoulon: > in host operator there are files named cluster.yaml so I thought that the name should be the same

that's a fair point. But we also don't have strictly the same file names for the namespace templates, AFAICS: https://github.com/codeready-toolchain/host-operator/tree/master/deploy/templates/nstemplatetiers/advanced

MatousJobanek: the namespace files look to have pretty strict naming:
`ns_<type>.yaml`
thus I would expect that the resource name of the TierTemplate containing cluster resources should use the file name `cluster`, not something else. 

To avoid confusion in the future I would propose to rename the files (or change the TierTemplate resource naming but this seems to be more complicated.)

pkg/controller/nstemplateset/nstemplateset_controller_test.go
MatousJobanek: I found pretty easily solution - the revision that is passed is always the same `abcde11` so I can easily remove the parameter and just keep `withTemplateRef()`
changed in 9ebe6f5

pkg/controller/nstemplateset/nstemplatetier.go
MatousJobanek: removed 758cca5

pkg/controller/nstemplateset/nstemplateset_controller_test.go
MatousJobanek: changed in e61459d

pkg/controller/nstemplateset/cluster_resources_test.go
MatousJobanek: added here: eba87b7

pkg/controller/nstemplateset/nstemplateset_controller_test.go
xcoulon: that's simpler, indeed!

go.mod
```diff
@@ -31,6 +31,8 @@ require (
 	sigs.k8s.io/kubefed v0.1.0-rc6.0.20200224204536-6207193c49f7
 )
 
+replace github.com/codeready-toolchain/api => github.com/MatousJobanek/api v0.0.0-20200514221709-6828f4a4090a
```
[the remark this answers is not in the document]
MatousJobanek: removed and updated 495f5b7

PR #162, opened 2020-05-26
feat: cache TierTemplates retrieved from the host cluster

MatousJobanek:
This PR adds a cache for TierTemplates retrieved from the host cluster. When a TierTemplate is retrieved from the host cluster, then it's stored in a map indexed by its name and for all future calls it uses only the cache instead of calling the host cluster again.

### DIFF
--- a/pkg/controller/nstemplateset/nstemplatetier.go
+++ b/pkg/controller/nstemplateset/nstemplatetier.go
@@ -3,6 +3,7 @@ package nstemplateset
 import (
 	"context"
 	"fmt"
+	"sync"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
 	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
@@ -15,7 +16,7 @@ import (
 	"sigs.k8s.io/kubefed/pkg/controller/util"
 )
 
-var cachedTierTemplates = map[string]*tierTemplate{}
+var tierTemplatesCache = newTierTemplateCache()
 
 // getTierTemplate retrieves the TierTemplate resource with the given name from the host cluster
 // and returns an instance of the tierTemplate type for it whose template content can be parsable.
@@ -24,7 +25,7 @@ func getTierTemplate(hostClusterFunc cluster.GetHostClusterFunc, templateRef str
 	if templateRef == "" {
 		return nil, fmt.Errorf("templateRef is not provided - it's not possible to fetch related TierTemplate resource")
 	}
-	if tierTmpl, ok := cachedTierTemplates[templateRef]; ok && tierTmpl != nil {
+	if tierTmpl, ok := tierTemplatesCache.get(templateRef); ok && tierTmpl != nil {
 		return tierTmpl, nil
 	}
 	tmpl, err := getToolchainTierTemplate(hostClusterFunc, templateRef)
@@ -37,7 +38,7 @@ func getTierTemplate(hostClusterFunc cluster.GetHostClusterFunc, templateRef str
 		typeName:    tmpl.Spec.Type,
 		template:    tmpl.Spec.Template,
 	}
-	cachedTierTemplates[templateRef] = tierTmpl
+	tierTemplatesCache.add(tierTmpl)
 
 	return tierTmpl, nil
 }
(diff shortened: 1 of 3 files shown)


pkg/controller/nstemplateset/nstemplatetier.go
```diff
@@ -15,23 +15,31 @@ import (
 	"sigs.k8s.io/kubefed/pkg/controller/util"
 )
 
+var cachedTierTemplates = map[string]*tierTemplate{}
```
alexeykazakov, line 18: Let’s add a comment saying that the key here is a templateref name.
Or simply rename it to cachedTierTemplatesByTemplateRef (it’s probably too long name though)

pkg/controller/nstemplateset/nstemplatetier.go
```diff
@@ -15,23 +15,31 @@ import (
...
 	}
-	return &tierTemplate{
+	tierTmpl := &tierTemplate{
 		templateRef: templateRef,
 		tierName:    tmpl.Spec.TierName,
 		typeName:    tmpl.Spec.Type,
 		template:    tmpl.Spec.Template,
-	}, nil
+	}
+	cachedTierTemplates[templateRef] = tierTmpl
```
alexeykazakov, line 40: We can’t do that, can we? Multiple go routines can mess up with the cache. We have to use rwmutex.

MatousJobanek: I was thinking about the need of using synchronization, but there is always only one reconcile running of the NSTemplateSet controller, so there won't be multiple goroutines accessing the cache. There will be always only one.
Just to be clear, I'm fine with adding the rwmutex to synchronize the access - after all, it can be considered as a best practise for designing the cache. I didn't add it because it's another complexity that is not really needed for now.

MatousJobanek: add synchronization here: 0e220e5
originally, I was thinking about more changes (moving to a separated package to isolate the cache, etc...) but I believe that this simple version is fine :-)

pkg/controller/nstemplateset/nstemplatetier.go
MatousJobanek: done 0e719f3

pkg/controller/nstemplateset/nstemplatetier.go
alexeykazakov: Are you sure only one reconcile is run at a time? I remember multiple cases when multiple reconciles were executed simultaneously (different controller though).

MatousJobanek: I was talking about one reconcile in one controller. This cache is used only in one controller (the vars and functions are not exported ;-))

alexeykazakov: Still.. what is so special about that reconcile? Not trying to argue :) I'm apparently missing something. Trying to understand what exactly I'm missing.

MatousJobanek: * This cache is used only in NSTempalteSet controller
* The whole flow of the reconcile logic is sequential - there is no parallelism. 
* There is always only one reconcile of the NSTempalteSet controller running at one point in time

=> Thus, there is always only one goroutine accessing the cache, so it shouldn't happen that there would be multiple get/add operations in parallel.

alexeykazakov: > There is always only one reconcile of the NSTempalteSet controller running at one point in time

This is what I was questioning. I did some research and found out that by default concurrent reconciling is disabled: https://github.com/kubernetes-sigs/controller-runtime/blob/81842d0e78f7111f0566156189806e2801e3adf1/pkg/controller/controller.go#L33-L34
But we can set `MaxConcurrentReconciles` to something >1 if we want in our controllers to enable concurrent reconciling for distinct objects (the same object can't be reconciled in parallel anyway).
So, yeah.. in the current NSTemplateSet (and controller) implementation cache synchronization is not necessary. But I would still keep it because of:
- We might want to enable concurrent reconciling for NSTemplateSet in the future.
- The default value of `MaxConcurrentReconciles` might be changed in the controllers in the future.
So I think it would be less error prone to make sure our reconciles loops support concurrent reconciling.

PR #165, opened 2020-06-02
chore: make sure that changes of template revisions trigger update

MatousJobanek:
There is no need to implement anything in the controllers - the update of both namespaces and cluster resources is supported out-of-the-box. 
This PR just adds unit tests covering specifically the use case when tier stays the same and the only thing that is changed is the revision in the TemplateRef.


pkg/controller/nstemplateset/namespaces_test.go
```diff
@@ -324,8 +323,8 @@ func TestEnsureNamespacesOK(t *testing.T) {
 
 	t.Run("should create the second namespace when the first one already exists", func(t *testing.T) {
 		// given
-		nsTmplSet := newNSTmplSet(namespaceName, username, "basic", withNamespaces("dev", "code"), withConditions(Provisioning()))
-		devNS := newNamespace("basic", username, "dev", withTemplateRef())
+		nsTmplSet := newNSTmplSet(namespaceName, username, "basic", withNamespaces("abcde11", "dev", "code"), withConditions(Provisioning()))
+		devNS := newNamespace("basic", username, "dev", withTempalteRefUsingRevision("abcde11"))
```
xcoulon, line 327: there's a small typo in the func name: `withTempalteRefUsingRevision` -> `withTemplateRefUsingRevision`

MatousJobanek:
README changed in 94f0fd4

pkg/controller/nstemplateset/namespaces_test.go
MatousJobanek: fixed in 7ff9a06
Thanks :+1:

PR #167, opened 2020-06-09
chore: add acon to CSV

MatousJobanek:
add icon to CSV so the formatting in OLM page in OpenShift console is better


MatousJobanek:
image
It's the same icon as is used in toolchain-operator :-) 
And thanks for a reminder regarding the typo

PR #174, opened 2020-06-23
Add member controller

rajivnathan:
Related issue: https://issues.redhat.com/browse/CRT-667

Link to PR tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/128

This PR adds a controller for the new MemberStatus resource. Its responsibility is to read the state of toolchain member cluster components and update the MemberStatus resource with information useful for observation or troubleshooting. For more details about the MemberStatus resource see the related issue.

Example:
```
apiVersion: toolchain.dev.openshift.com/v1alpha1
kind: MemberStatus
metadata:
  annotations:
    toolchain.dev.openshift.com/last-applied-configuration: '{"metadata":{"name":"toolchain-member-status","namespace":"rsenthil-member-operator","creationTimestamp":null},"spec":{},"status":{"memberOperator":{"version":"","revision":"","buildTimestamp":"","deployment":{"name":""}},"hostConnection":{"conditions":null}}}'
  creationTimestamp: "2020-06-30T02:36:55Z"
  generation: 1
  name: toolchain-member-status
  namespace: rsenthil-member-operator
  resourceVersion: "1840440"
  selfLink: /apis/toolchain.dev.openshift.com/v1alpha1/namespaces/rsenthil-member-operator/memberstatuses/toolchain-member-status
  uid: d19e0e89-5eb8-4cd6-abd1-7e1bc96ca4d4
spec: {}
status:
  conditions:
  - lastTransitionTime: "2020-06-30T09:48:29Z"
    lastUpdatedTime: "2020-06-30T09:48:29Z"
    message: 'components not ready: [memberOperator hostConnection]'
    reason: ComponentsNotReady
    status: "False"
    type: Ready
  hostConnection:
    conditions:
    - lastProbeTime: "2020-06-30T09:48:19Z"
      lastTransitionTime: "2020-06-30T09:46:59Z"
      message: /healthz responded with ok
      reason: ClusterReady
      status: "True"
      type: Ready
    - lastProbeTime: "2020-06-30T09:48:19Z"
      message: 'exceeded the maximum duration since the last probe: 39s'
      reason: KubefedLastProbeTimeExceeded
      status: "False"
      type: Ready
    region: ""
  memberOperator:
    buildTimestamp: "2020-06-30T19:58:59Z"
    conditions:
    - lastTransitionTime: "2020-06-30T09:48:29Z"
      lastUpdatedTime: "2020-06-30T09:48:29Z"
      message: the member operator deployment has unavailable replicas
      reason: DeploymentUnavailableReplicas
      status: "False"
      type: Ready
    deployment:
      name: member-operator
    revision: e38ad2d0883e3295ae32a26d96c04dfc7a9cb57c-dirty
    version: 0.0.1
```

### DIFF
--- a/pkg/controller/memberstatus/creation.go
+++ b/pkg/controller/memberstatus/creation.go
@@ -9,15 +9,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func CreateOrUpdateResources(client client.Client, s *runtime.Scheme, namespace string) error {
+func CreateOrUpdateResources(client client.Client, s *runtime.Scheme, namespace, memberStatusName string) error {
 	memberStatus := &v1alpha1.MemberStatus{
 		ObjectMeta: v1.ObjectMeta{
 			Namespace: namespace,
-			Name:      defaultMemberStatusName,
+			Name:      memberStatusName,
 		},
 		Spec: v1alpha1.MemberStatusSpec{},
 	}
-	commonclient := commonclient.NewApplyClient(client, s)
-	_, err := commonclient.CreateOrUpdateObject(memberStatus, false, nil)
+	cl := commonclient.NewApplyClient(client, s)
+	_, err := cl.CreateOrUpdateObject(memberStatus, false, nil)
 	return err
 }
--- a/pkg/controller/memberstatus/creation.go
+++ b/pkg/controller/memberstatus/creation.go
@@ -1,0 +1,23 @@
+package memberstatus
+
+import (
+	"github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+	commonclient "github.com/codeready-toolchain/toolchain-common/pkg/client"
+
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func CreateOrUpdateResources(client client.Client, s *runtime.Scheme, namespace string) error {
+	memberStatus := &v1alpha1.MemberStatus{
+		ObjectMeta: v1.ObjectMeta{
+			Namespace: namespace,
+			Name:      defaultMemberStatusName,
+		},
+		Spec: v1alpha1.MemberStatusSpec{},
+	}
+	commonclient := commonclient.NewApplyClient(client, s)
+	_, err := commonclient.CreateOrUpdateObject(memberStatus, false, nil)
+	return err
+}
--- a/pkg/controller/memberstatus/creation_test.go
+++ b/pkg/controller/memberstatus/creation_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 
 	"github.com/codeready-toolchain/member-operator/pkg/apis"
+	"github.com/codeready-toolchain/member-operator/pkg/configuration"
+	. "github.com/codeready-toolchain/member-operator/test"
 	. "github.com/codeready-toolchain/toolchain-common/pkg/test"
 	"github.com/stretchr/testify/require"
 
@@ -25,15 +27,13 @@ func TestCreateOrUpdateResources(t *testing.T) {
 		cl := NewFakeClient(t)
 
 		// when
-		err = CreateOrUpdateResources(cl, s, MemberOperatorNs)
+		err = CreateOrUpdateResources(cl, s, MemberOperatorNs, configuration.DefaultMemberStatusName)
 
 		// then
 		require.NoError(t, err)
 
 		// check that the MemberStatus exists now
-		memberStatus := newMemberStatus()
-		err = cl.Get(context.TODO(), NamespacedName(MemberOperatorNs, defaultMemberStatusName), memberStatus)
-		require.NoError(t, err)
+		AssertThatMemberStatus(t, MemberOperatorNs, "toolchain-member-status", cl).Exists()
 	})
 
 	t.Run("should return an error if creation fails ", func(t *testing.T) {
@@ -44,7 +44,7 @@ func TestCreateOrUpdateResources(t *testing.T) {
 		}
 
 		// when
-		err = CreateOrUpdateResources(cl, s, MemberOperatorNs)
+		err = CreateOrUpdateResources(cl, s, MemberOperatorNs, configuration.DefaultMemberStatusName)
 
 		// then
 		require.Error(t, err)
--- a/pkg/controller/memberstatus/creation_test.go
+++ b/pkg/controller/memberstatus/creation_test.go
@@ -1,0 +1,47 @@
+package memberstatus
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/codeready-toolchain/member-operator/pkg/apis"
+	. "github.com/codeready-toolchain/toolchain-common/pkg/test"
+	"github.com/stretchr/testify/require"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestCreateOrUpdateResources(t *testing.T) {
+	// given
+	s := scheme.Scheme
+	err := apis.AddToScheme(s)
+	require.NoError(t, err)
+
+	t.Run("creation", func(t *testing.T) {
+		// given
+		cl := NewFakeClient(t)
+
+		// when
+		err = CreateOrUpdateResources(cl, s, MemberOperatorNs)
+
+		// then
+		require.NoError(t, err)
+	})
+
+	t.Run("should return an error if creation fails ", func(t *testing.T) {
+		// given
+		cl := NewFakeClient(t)
+		cl.MockCreate = func(ctx context.Context, obj runtime.Object, opts ...client.CreateOption) error {
+			return fmt.Errorf("creation failed")
+		}
+
+		// when
+		err = CreateOrUpdateResources(cl, s, MemberOperatorNs)
+
+		// then
+		require.Error(t, err)
+	})
+}
--- a/pkg/controller/memberstatus/creation_test.go
+++ b/pkg/controller/memberstatus/creation_test.go
@@ -29,6 +29,11 @@ func TestCreateOrUpdateResources(t *testing.T) {
 
 		// then
 		require.NoError(t, err)
+
+		// check that the MemberStatus exists now
+		memberStatus := newMemberStatus()
+		err = cl.Get(context.TODO(), NamespacedName(MemberOperatorNs, defaultMemberStatusName), memberStatus)
+		require.NoError(t, err)
 	})
 
 	t.Run("should return an error if creation fails ", func(t *testing.T) {
@@ -43,5 +48,6 @@ func TestCreateOrUpdateResources(t *testing.T) {
 
 		// then
 		require.Error(t, err)
+		require.Contains(t, err.Error(), "unable to create resource")
 	})
 }
--- a/pkg/controller/memberstatus/memberstatus_assertion.go
+++ b/pkg/controller/memberstatus/memberstatus_assertion.go
@@ -1,0 +1,67 @@
+package memberstatus
+
+import (
+	"context"
+	"fmt"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
+	"github.com/codeready-toolchain/toolchain-common/pkg/test"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type MemberStatusAssertion struct {
+	memberStatus   *toolchainv1alpha1.MemberStatus
+	client         client.Client
+	namespacedName types.NamespacedName
+	t              test.T
+}
+
+func (a *MemberStatusAssertion) loadMemberStatus() error {
+	memberStatus := &toolchainv1alpha1.MemberStatus{}
+	err := a.client.Get(context.TODO(), a.namespacedName, memberStatus)
+	a.memberStatus = memberStatus
+	return err
+}
+
+func AssertThatMemberStatus(t test.T, namespace, name string, client client.Client) *MemberStatusAssertion {
+	return &MemberStatusAssertion{
+		client:         client,
+		namespacedName: test.NamespacedName(namespace, name),
+		t:              t,
+	}
+}
+
+func (a *MemberStatusAssertion) HasNoConditions() *MemberStatusAssertion {
+	err := a.loadMemberStatus()
+	require.NoError(a.t, err)
+	require.Empty(a.t, a.memberStatus.Status.Conditions)
+	return a
+}
+
+func (a *MemberStatusAssertion) HasCondition(expected toolchainv1alpha1.Condition) *MemberStatusAssertion {
+	err := a.loadMemberStatus()
+	require.NoError(a.t, err)
+	test.AssertConditionsMatch(a.t, a.memberStatus.Status.Conditions, expected)
+	return a
+}
+
+func ComponentsReady() toolchainv1alpha1.Condition {
+	return toolchainv1alpha1.Condition{
+		Type:   toolchainv1alpha1.ConditionReady,
+		Status: corev1.ConditionTrue,
+		Reason: toolchainv1alpha1.MemberStatusAllComponentsReady,
+	}
+}
+
+func ComponentsNotReady(components ...statusComponentTag) toolchainv1alpha1.Condition {
+	return toolchainv1alpha1.Condition{
+		Type:    toolchainv1alpha1.ConditionReady,
+		Status:  corev1.ConditionFalse,
+		Reason:  toolchainv1alpha1.MemberStatusComponentsNotReady,
+		Message: fmt.Sprintf("Components not ready: %v", components),
+	}
+}
--- a/pkg/controller/memberstatus/memberstatus_controller.go
+++ b/pkg/controller/memberstatus/memberstatus_controller.go
@@ -6,7 +6,7 @@ import (
 	"time"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
-	"github.com/codeready-toolchain/member-operator/pkg/configuration"
+	crtCfg "github.com/codeready-toolchain/member-operator/pkg/configuration"
 	"github.com/codeready-toolchain/member-operator/version"
 	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
 
@@ -42,14 +42,28 @@ const (
 
 // custom component condition error messages & reasons
 const (
+	// deployment ready
+	reasonDeploymentReady = "DeploymentReady"
+
 	// deployment not found
-	reasonNoDeployment         = "DeploymentNotFound"
-	errMsgNoMemberOperatorName = "unable to look up the member operator name, the environment variable OPERATOR_NAME is not set"
-	errMsgCannotGetDeployment  = "unable to get the member operator deployment"
+	reasonNoDeployment        = "DeploymentNotFound"
+	errMsgCannotGetDeployment = "unable to get the member operator deployment"
+
+	// deployment has unavailable replicas
+	reasonUnavailableReplicas = "DeploymentUnavailableReplicas"
+	errMsgUnavailableReplicas = "the member operator deployment has unavailable replicas"
+
+	// deployment not ready
+	reasonDeploymentConditionNotReady = "DeploymentNotReady"
+	errMsgDeploymentConditionNotReady = "member operator deployment has unready status conditions"
 
 	// kubefed not found
 	reasonHostConnectionNotFound = "KubefedNotFound"
 	errMsgHostConnectionNotFound = "the host connection was not found"
+
+	// kubefed last probe time exceeded
+	reasonHostConnectionLastProbeTimeExceeded = "KubefedLastProbeTimeExceeded"
+	errMsgHostConnectionLastProbeTimeExceeded = "exceeded the maximum duration since the last probe"
 )
 
 // statusComponentTags are used in the overall condition to point out which components are not ready
@@ -62,16 +76,17 @@ const (
 
 // Add creates a new MemberStatus Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
-func Add(mgr manager.Manager, _ *configuration.Config) error {
-	return add(mgr, newReconciler(mgr))
+func Add(mgr manager.Manager, crtConfig *crtCfg.Config) error {
+	return add(mgr, newReconciler(mgr, crtConfig))
 }
 
 // newReconciler returns a new reconcile.Reconciler
-func newReconciler(mgr manager.Manager) *ReconcileMemberStatus {
+func newReconciler(mgr manager.Manager, crtConfig *crtCfg.Config) *ReconcileMemberStatus {
 	return &ReconcileMemberStatus{
 		client:         mgr.GetClient(),
 		scheme:         mgr.GetScheme(),
 		getHostCluster: cluster.GetHostCluster,
+		config:         crtConfig,
 	}
 }
 
@@ -102,6 +117,7 @@ type ReconcileMemberStatus struct {
 	client         client.Client
 	scheme         *runtime.Scheme
 	getHostCluster func() (*cluster.FedCluster, bool)
+	config         *crtCfg.Config
 }
 
 // Reconcile reads the state of toolchain member cluster components and updates the MemberStatus resource with information useful for observation or troubleshooting
@@ -133,7 +149,7 @@ func (r *ReconcileMemberStatus) Reconcile(request reconcile.Request) (reconcile.
 }
 
 type statusHandler struct {
-	Name         statusComponentTag
+	name         statusComponentTag
 	handleStatus func(logger logr.Logger, memberStatus *toolchainv1alpha1.MemberStatus) error
 }
 
@@ -141,8 +157,8 @@ type statusHandler struct {
 // component status is not ready then it will set the condition of the top-level status of the MemberStatus resource to not ready.
 func (r *ReconcileMemberStatus) aggregateAndUpdateStatus(reqLogger logr.Logger, memberStatus *toolchainv1alpha1.MemberStatus) error {
 
-	memberOperatorStatusHandler := statusHandler{Name: memberOperator, handleStatus: r.memberOperatorHandleStatus}
-	hostConnectionStatusHandler := statusHandler{Name: hostConnection, handleStatus: r.hostConnectionHandleStatus}
+	memberOperatorStatusHandler := statusHandler{name: memberOperator, handleStatus: r.memberOperatorHandleStatus}
+	hostConnectionStatusHandler := statusHandler{name: hostConnection, handleStatus: r.hostConnectionHandleStatus}
 
 	statusHandlers := []statusHandler{memberOperatorStatusHandler, hostConnectionStatusHandler}
 
@@ -154,7 +170,7 @@ func (r *ReconcileMemberStatus) aggregateAndUpdateStatus(reqLogger logr.Logger, 
 		err := handler.handleStatus(reqLogger, memberStatus)
 		if err != nil {
 			reqLogger.Error(err, "status update problem")
-			unreadyComponents = append(unreadyComponents, string(handler.Name))
+			unreadyComponents = append(unreadyComponents, string(handler.name))
 		}
 	}
 
@@ -194,6 +210,44 @@ func (r *ReconcileMemberStatus) hostConnectionHandleStatus(reqLogger logr.Logger
 		return fmt.Errorf("the host connection is not ready")
 	}
 
+	var lastProbeTime metav1.Time
+	foundLastProbeTime := false
+	for _, condition := range fedCluster.ClusterStatus.Conditions {
+		if condition.Type == kubefed_common.ClusterReady {
+			lastProbeTime = condition.LastProbeTime
+			foundLastProbeTime = true
+		}
+	}
+	if !foundLastProbeTime {
+		return fmt.Errorf("the time of the last probe could not be determined")
+	}
+
+	// check that the last probe time is within limits. It should be less than (period + timeout) * threshold
+	period := r.config.GetClusterHealthCheckPeriod()
+	timeout := r.config.GetClusterHealthCheckTimeout()
+	threshold := r.config.GetClusterHealthCheckFailureThreshold()
+	totalf := (period.Seconds() + timeout.Seconds()) * float64(threshold)
+	maxDuration, err := time.ParseDuration(fmt.Sprintf("%fs", totalf))
+	if err != nil {
+		return fmt.Errorf("the maximum duration since the last probe could not be determined - check the configured %s %s and %s", crtCfg.ClusterHealthCheckPeriod, crtCfg.ClusterHealthCheckTimeout, crtCfg.ClusterHealthCheckFailureThreshold)
+	}
+
+	lastProbedTimePlusMaxDuration := lastProbeTime.Add(maxDuration)
+	currentTime := time.Now()
+	if currentTime.After(lastProbedTimePlusMaxDuration) {
+		errMsg := fmt.Sprintf("%s: %s", errMsgHostConnectionLastProbeTimeExceeded, maxDuration.String())
+		errReason := reasonHostConnectionLastProbeTimeExceeded
+		probeCondition := kubefed_v1beta1.ClusterCondition{
+			Type:          "ProbeWorking",
+			Status:        corev1.ConditionFalse,
+			Reason:        &errReason,
+			Message:       &errMsg,
+			LastProbeTime: lastProbeTime,
+		}
+		memberStatus.Status.HostConnection.Conditions = append(memberStatus.Status.HostConnection.Conditions, probeCondition)
+		return fmt.Errorf(errMsg)
+	}
+
 	return nil
 }
 
@@ -226,23 +280,31 @@ func (r *ReconcileMemberStatus) memberOperatorHandleStatus(reqLogger logr.Logger
 		memberStatus.Status.MemberOperator = operatorStatus
 		return err
 	}
-
-	// get and check conditions of member deployment
-	conditionsReady := true
-	deploymentConditions := []appsv1.DeploymentCondition{}
-	for _, condition := range memberDeployment.Status.Conditions {
-		deploymentConditions = append(deploymentConditions, condition)
-		conditionsReady = conditionsReady && condition.Status == corev1.ConditionTrue
-	}
-
-	// update member status
 	operatorStatus.Deployment.Name = memberDeployment.Name
-	operatorStatus.Deployment.DeploymentConditions = deploymentConditions
-	memberStatus.Status.MemberOperator = operatorStatus
 
-	if !conditionsReady {
-		return fmt.Errorf("the member operator deployment is not ready")
+	// check replicas
+	if memberDeployment.Status.UnavailableReplicas > 0 {
+		errCondition := newComponentErrorCondition(reasonUnavailableReplicas, errMsgUnavailableReplicas)
+		operatorStatus.Conditions = []toolchainv1alpha1.Condition{*errCondition}
+		memberStatus.Status.MemberOperator = operatorStatus
+		return fmt.Errorf(errMsgUnavailableReplicas)
 	}
+
+	// get and check conditions
+	for _, condition := range memberDeployment.Status.Conditions {
+		if condition.Status != corev1.ConditionTrue {
+			errMsg := fmt.Sprintf("%s: %s", errMsgDeploymentConditionNotReady, condition.Type)
+			errCondition := newComponentErrorCondition(reasonDeploymentConditionNotReady, errMsg)
+			operatorStatus.Conditions = []toolchainv1alpha1.Condition{*errCondition}
+			memberStatus.Status.MemberOperator = operatorStatus
+			return fmt.Errorf(errMsg)
+		}
+	}
+
+	// update member operator status
+	operatorReadyCondition := newComponentReadyCondition(reasonDeploymentReady)
+	operatorStatus.Conditions = []toolchainv1alpha1.Condition{*operatorReadyCondition}
+	memberStatus.Status.MemberOperator = operatorStatus
 
 	return nil
 }
@@ -280,6 +342,17 @@ func (r *ReconcileMemberStatus) setStatusNotReady(memberStatus *toolchainv1alpha
 			Reason:  toolchainv1alpha1.MemberStatusComponentsNotReady,
 			Message: message,
 		})
+}
+
+func newComponentReadyCondition(reason string) *toolchainv1alpha1.Condition {
+	currentTime := metav1.Now()
+	return &toolchainv1alpha1.Condition{
+		Type:               toolchainv1alpha1.ConditionReady,
+		Status:             corev1.ConditionTrue,
+		Reason:             reason,
+		LastTransitionTime: currentTime,
+		LastUpdatedTime:    &currentTime,
+	}
 }
 
 func newComponentErrorCondition(reason, msg string) *toolchainv1alpha1.Condition {
--- a/pkg/controller/memberstatus/memberstatus_controller.go
+++ b/pkg/controller/memberstatus/memberstatus_controller.go
@@ -3,7 +3,6 @@ package memberstatus
 import (
 	"context"
 	"fmt"
-	"os"
 	"time"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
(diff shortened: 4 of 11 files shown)


pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,236 @@
...
+
+	// track components that are not ready
+	unreadyComponents := []string{}
+
+	// retrieve component statuses eg. kubefed, member deployment
+	for _, handler := range statusHandlers {
+		err := handler.handleStatus(reqLogger, memberStatus)
+		if err != nil {
+			reqLogger.Error(err, "status update problem")
+			unreadyComponents = append(unreadyComponents, string(handler.Name))
```
alexeykazakov, line 140: The errors returned by the handlers seem to get lost (they are only logged). I don't see the error messages to be added to the MemberStatus.

It seems that there is a mix of different kind of errors in both handlers: 
errors which should go to the status when one of the component is not ready from an error when we can't really obtain the status (there is a problem with the client for example or there is no KubeFed cluster resource at all).

I would rather add any error we get when trying to update the component status to the corresponding component status. In this case a handler don't really need to return any error (all errors go to the component) but return a bool instead (component ready, or not). If any handler returns `false` then the overall condition is not ready.

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,236 @@
...
+		Revision:       version.Commit,
+		BuildTimestamp: version.BuildTime,
+	}
+
+	// look up status of member deployment
+	memberDeploymentName := types.NamespacedName{Namespace: memberStatus.Namespace, Name: memberOperatorDeploymentName}
+	memberDeployment := &appsv1.Deployment{}
+	err := r.client.Get(context.TODO(), memberDeploymentName, memberDeployment)
+	if err != nil {
+		return fmt.Errorf("the member operator deployment was not found")
```
alexeykazakov, line 180: It can be NotFound. Or it can be something else (a client error). Do not hide the original error.

pkg/controller/memberstatus/creation_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	t.Run("creation", func(t *testing.T) {
+		// given
+		cl := NewFakeClient(t)
+
+		// when
+		err = CreateOrUpdateResources(cl, s, MemberOperatorNs)
+
+		// then
+		require.NoError(t, err)
```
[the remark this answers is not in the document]
rajivnathan: Done: f617c6207d45e7f528f250ccae0ac2fe72fa17d6

pkg/controller/memberstatus/creation_test.go
```diff
@@ -0,0 +1,47 @@
...
+		cl := NewFakeClient(t)
+		cl.MockCreate = func(ctx context.Context, obj runtime.Object, opts ...client.CreateOption) error {
+			return fmt.Errorf("creation failed")
+		}
+
+		// when
+		err = CreateOrUpdateResources(cl, s, MemberOperatorNs)
+
+		// then
+		require.Error(t, err)
```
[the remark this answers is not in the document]
rajivnathan: Done: f617c6207d45e7f528f250ccae0ac2fe72fa17d6

pkg/controller/memberstatus/memberstatus_assertion.go
```diff
@@ -0,0 +1,67 @@
+package memberstatus
```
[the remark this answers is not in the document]
rajivnathan: Done: f617c6207d45e7f528f250ccae0ac2fe72fa17d6

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,236 @@
...
+)
+
+var log = logf.Log.WithName("controller_memberstatus")
+
+// general memberstatus constants
+const (
+	defaultMemberStatusName      = "toolchain-member-status"
+	memberOperatorDeploymentName = "member-operator"
+
+	defaultRequeueTime = time.Second * 10
```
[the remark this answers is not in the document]
rajivnathan: Sure, updated in f617c6207d45e7f528f250ccae0ac2fe72fa17d6

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,236 @@
...
+	memberStatus := &toolchainv1alpha1.MemberStatus{}
+	err := r.client.Get(context.TODO(), request.NamespacedName, memberStatus)
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// MemberStatus resource not found, log error and return
+			reqLogger.Error(err, "MemberStatus resource was not found")
+		}
+		return reconcile.Result{}, nil
+	}
```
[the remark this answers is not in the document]
rajivnathan: Sure, makes sense. Done: f617c6207d45e7f528f250ccae0ac2fe72fa17d6

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,236 @@
...
+
+	// if any components were not ready then set the overall status to not ready
+	if len(unreadyComponents) > 0 {
+		err := fmt.Errorf("Components not ready: %v", unreadyComponents)
+		return r.setStatusNotReady(memberStatus, err.Error())
+	}
+	return r.setStatusReady(memberStatus)
+}
+
+func (r *ReconcileMemberStatus) hostConnectionHandleStatus(reqLogger logr.Logger, memberStatus *toolchainv1alpha1.MemberStatus) error {
```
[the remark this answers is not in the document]
rajivnathan: Done: f617c6207d45e7f528f250ccae0ac2fe72fa17d6

pkg/controller/memberstatus/memberstatus_controller.go
rajivnathan: Good catch! I wrapped the error now: f617c6207d45e7f528f250ccae0ac2fe72fa17d6

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,236 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/source"
+	"sigs.k8s.io/kubefed/pkg/controller/util"
+)
+
+var log = logf.Log.WithName("controller_memberstatus")
+
+// general memberstatus constants
+const (
+	defaultMemberStatusName      = "toolchain-member-status"
+	memberOperatorDeploymentName = "member-operator"
```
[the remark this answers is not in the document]
rajivnathan: Nice, good point! Addressed and tested with f617c6207d45e7f528f250ccae0ac2fe72fa17d6

pkg/controller/memberstatus/memberstatus_controller.go
rajivnathan: Updated it so that now all the errors are visible in the component statuses. I kept the status handlers returning errors instead of bool because we're currently logging the error which I think is still useful in case some errors show up briefly and disappear? But I can change it if we'd rather handle it differently.

So with the change when the kubefed resource or deployment can't be found the error is now available in the component status:
```
status:
  conditions:
  - lastTransitionTime: "2020-06-24T20:37:47Z"
    lastUpdatedTime: "2020-06-24T20:37:47Z"
    message: 'Components not ready: [memberOperator hostConnection]'
    reason: ComponentsNotReady
    status: "False"
    type: Ready
  hostConnection:
    conditions:
    - lastProbeTime: "2020-06-24T20:37:47Z"
      message: the host connection was not found
      reason: KubefedNotFound
      status: "False"
      type: Ready
  memberOperator:
    buildTimestamp: "2020-06-24T20:35:11Z"
    conditions:
    - lastTransitionTime: "2020-06-24T20:37:47Z"
      lastUpdatedTime: "2020-06-24T20:37:47Z"
      message: unable to get the member operator deployment
      reason: DeploymentNotFound
      status: "False"
      type: Ready
    deployment:
      name: ""
    revision: f617c6207d45e7f528f250ccae0ac2fe72fa17d6-dirty
    version: 0.0.1
```

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,293 @@
...
+		Revision:       version.Commit,
+		BuildTimestamp: version.BuildTime,
+	}
+
+	// look up status of member deployment
+	cannotGetDeploymentMsg := "unable to get the member operator deployment"
+	noDeploymentReason := "DeploymentNotFound"
+	memberOperatorDeploymentName, err := getMemberOperatorDeploymentName()
+	if err != nil {
+		errCondition := newComponentErrorCondition(noDeploymentReason, cannotGetDeploymentMsg)
```
alexeykazakov, line 204: You lost the original `err` here. Why we can't get the deployment? Only `err` can tell you: "unable to look up the member operator name, the environment variable OPERATOR_NAME is not set".
So, I would rather do the following:
```suggestion
		errCondition := newComponentErrorCondition(noDeploymentReason, err.Error())
```
Or:
```suggestion
		err = errs.Wrap(err, cannotGetDeploymentMsg)
		errCondition := newComponentErrorCondition(noDeploymentReason, err.Error())
```

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,293 @@
...
+		operatorStatus.Conditions = []toolchainv1alpha1.Condition{*errCondition}
+		memberStatus.Status.MemberOperator = operatorStatus
+		return fmt.Errorf(cannotGetDeploymentMsg)
```
alexeykazakov, line 207: It doesn't matter much actually.. but just for the sake of consistency... If you wrap err above:
```suggestion
		return err
```

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	memberDeploymentName := types.NamespacedName{Namespace: memberStatus.Namespace, Name: memberOperatorDeploymentName}
+	memberDeployment := &appsv1.Deployment{}
+	err = r.client.Get(context.TODO(), memberDeploymentName, memberDeployment)
+	if err != nil {
+		errCondition := newComponentErrorCondition(noDeploymentReason, cannotGetDeploymentMsg)
```
alexeykazakov, line 213:
```suggestion
		err = errs.Wrap(err, cannotGetDeploymentMsg)
		errCondition := newComponentErrorCondition(noDeploymentReason, err.Error())
```

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,293 @@
...
+		operatorStatus.Conditions = []toolchainv1alpha1.Condition{*errCondition}
+		memberStatus.Status.MemberOperator = operatorStatus
+		return errs.Wrap(err, cannotGetDeploymentMsg)
```
alexeykazakov, line 216:
```suggestion
		return err
```

pkg/controller/memberstatus/memberstatus_controller_test.go
```diff
@@ -0,0 +1,241 @@
...
+		// when
+		res, err := reconciler.Reconcile(req)
+
+		// then - there should not be any error, the controller should only log that the resource was not found
+		require.NoError(t, err)
+		assert.Equal(t, reconcile.Result{}, res)
+	})
+}
+
+func TestOverallStatusCondition(t *testing.T) {
```
alexeykazakov, line 44: What about some tests to test the actual `hostConnectionHandleStatus()` and `memberOperatorHandleStatus()`? They are not covered by the tests if I'm not missing anything...
You can test the host one one by mocking `getHostCluster` function and the member one by using a fake client and overriding the `Get()` function to return the expected deployment or an error.

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,293 @@
...
+		memberStatus,
+		toolchainv1alpha1.Condition{
+			Type:    toolchainv1alpha1.ConditionReady,
+			Status:  corev1.ConditionFalse,
+			Reason:  toolchainv1alpha1.MemberStatusComponentsNotReady,
+			Message: message,
+		})
+}
+
+// gets the member operator deployment name via environment variable, returns an error if the value is empty or not set
```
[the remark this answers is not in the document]
rajivnathan: Fixed in ac86805b75b182c6030b9c827020aa3a58cb8624

pkg/controller/memberstatus/memberstatus_controller.go
rajivnathan: Sorry, I meant to wrap the error. Stupid mistake :disappointed:

Fixed in ac86805b75b182c6030b9c827020aa3a58cb8624

pkg/controller/memberstatus/memberstatus_controller.go
rajivnathan: Fixed in ac86805b75b182c6030b9c827020aa3a58cb8624

pkg/controller/memberstatus/memberstatus_controller.go
rajivnathan: Fixed in ac86805b75b182c6030b9c827020aa3a58cb8624

pkg/controller/memberstatus/memberstatus_controller.go
rajivnathan: Fixed in ac86805b75b182c6030b9c827020aa3a58cb8624

pkg/controller/memberstatus/memberstatus_controller_test.go
```diff
@@ -39,6 +42,25 @@ func TestNoMemberStatusFound(t *testing.T) {
...
+		getHostClusterFunc := newGetHostClusterReady
+		reconciler, req, fakeClient := prepareReconcile(t, requestName, getHostClusterFunc)
+		fakeClient.MockGet = func(ctx context.Context, key types.NamespacedName, obj runtime.Object) error {
+			return fmt.Errorf(expectedErrMsg)
+		}
+
+		// when
+		res, err := reconciler.Reconcile(req)
+
+		// then - there should not be any error, the controller should only log that the resource was not found
```
alexeykazakov, line 59: Need to fix the comment.

pkg/controller/memberstatus/memberstatus_controller_test.go
```diff
@@ -78,6 +100,7 @@ func TestOverallStatusCondition(t *testing.T) {
 		assert.Equal(t, requeueResult, res)
 		AssertThatMemberStatus(t, req.Namespace, requestName, fakeClient).
 			HasCondition(ComponentsNotReady(string(hostConnection)))
+		AssertThatMemberStatus(t, req.Namespace, requestName, fakeClient).HasHostConnectionConditionErrorMsg(errMsgHostConnectionNotFound)
```
alexeykazakov, line 103: We usually do not use const from runtime in the test. It's better to explicitly copy the strings.
```suggestion
		AssertThatMemberStatus(t, req.Namespace, requestName, fakeClient).HasHostConnectionConditionErrorMsg("the host connection was not found")
```
The reason for that is if we accidentally change the const to something wrong ("blah-blah" for example). Then the test will still pass. Using strings allows as to test the constants too.

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -42,6 +42,18 @@ const (
...
+// custom component condition error messages & reasons
+const (
+	// deployment not found
+	reasonNoDeployment         = "DeploymentNotFound"
+	errMsgNoMemberOperatorName = "unable to look up the member operator name, the environment variable OPERATOR_NAME is not set"
+	errMsgCannotGetDeployment  = "unable to get the member operator deployment"
+
+	// kubefed not found
+	reasonHostConnectionNotFound = "KubefedNotFound"
+	errMsgHostConnectionNotFound = "the host connection was not found"
```
alexeykazakov, line 54: As mentioned above, let's not use these constants in tests. It's OK to keep using them in the controller of course.

pkg/controller/memberstatus/memberstatus_controller_test.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+
+		// when
+		res, err := reconciler.Reconcile(req)
+
+		// then - there should not be any error, the controller should only log that the resource was not found
+		require.Error(t, err)
+		require.Equal(t, expectedErrMsg, err.Error())
+		assert.Equal(t, reconcile.Result{}, res)
+	})
```
alexeykazakov, line 63: Also check the status.

pkg/controller/memberstatus/memberstatus_controller_test.go
rajivnathan: Nice catch!

Fixed in c4d725ad046b4bf1e2f175f1bea36aa148d68ec3

pkg/controller/memberstatus/memberstatus_controller_test.go
rajivnathan: Makes sense. Updated in c4d725ad046b4bf1e2f175f1bea36aa148d68ec3

pkg/controller/memberstatus/memberstatus_controller_test.go
rajivnathan: In this case I am not checking the status because there is no `memberStatus` resource. Unless you mean to check the status of something else?

pkg/controller/memberstatus/memberstatus_controller_test.go
rajivnathan: The tests were covering most of `hostConnectionHandleStatus()` and `memberOperatorHandleStatus()` but I added a few more cases to get complete coverage. They now have complete coverage:

image

pkg/controller/memberstatus/memberstatus_controller_test.go
alexeykazakov: Never mind. I confused it with the No Deployment test.

pkg/controller/memberstatus/memberstatus_controller_test.go
alexeykazakov: Coverage is good now. BTW you can also check the coverage here: https://codecov.io/gh/codeready-toolchain/member-operator/pull/174/diff

rajivnathan: Ah, thanks! I didn't know it updates before the e2e test completes.

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,305 @@
...
+
+// custom component condition error messages & reasons
+const (
+	// deployment not found
+	reasonNoDeployment         = "DeploymentNotFound"
+	errMsgNoMemberOperatorName = "unable to look up the member operator name, the environment variable OPERATOR_NAME is not set"
+	errMsgCannotGetDeployment  = "unable to get the member operator deployment"
+
+	// kubefed not found
+	reasonHostConnectionNotFound = "KubefedNotFound"
```
MatousJobanek, line 53: Shouldn't be the reasons specified in api repo? It can be then also reused in host-operator. But only the reasons - the errors should stay here.

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,305 @@
...
+	"sigs.k8s.io/kubefed/pkg/controller/util"
+)
+
+var log = logf.Log.WithName("controller_memberstatus")
+
+// general memberstatus constants
+const (
+	OperatorNameVar = "OPERATOR_NAME"
+
+	defaultMemberStatusName = "toolchain-member-status"
```
MatousJobanek, line 41: I believe that both the `OperatorNameVar` and the `defaultMemberStatusName` could be moved to `configuration.go` file and load it as part of the config

MatousJobanek: In addition, you could use the function `k8sutil.GetOperatorName()`

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,305 @@
...
+		return err
+	}
+
+	// get and check conditions of member deployment
+	conditionsReady := true
+	deploymentConditions := []appsv1.DeploymentCondition{}
+	for _, condition := range memberDeployment.Status.Conditions {
+		deploymentConditions = append(deploymentConditions, condition)
+		conditionsReady = conditionsReady && condition.Status == corev1.ConditionTrue
+	}
```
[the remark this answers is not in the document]
alexeykazakov: It’s possible to get a state when the deployment can’t roll out a new version but the old/current one is still running.
We should make sure we cover that case.
Rajiv, can you double check that?

And, no, if there is no operator pods running at all (even some old version) then there is no controller to check the status and it will be caught by the host status.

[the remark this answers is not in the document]
MatousJobanek: > It’s possible to get a state when the deployment can’t roll out a new version but the old/current one is still running.

If there is a problem with starting the pod, then yes, you are rigth. 
Actually, I've been playing with it a bit and I observed that when I release a new version of the operator and delete the image from the repository (which causes `ImagePullBackOff` error while doing the update) then I don't see any error in any of the conditions in the `Deployment` yaml file. The only thing that shows that there is a problem is the number of the unavailable replicas:
```yaml
status:
  observedGeneration: 5
  replicas: 2
  updatedReplicas: 1
  readyReplicas: 1
  availableReplicas: 1
  unavailableReplicas: 1
  conditions:
    - type: Available
      status: 'True'
      lastUpdateTime: '2020-06-29T14:11:58Z'
      lastTransitionTime: '2020-06-29T14:11:58Z'
      reason: MinimumReplicasAvailable
      message: Deployment has minimum availability.
    - type: Progressing
      status: 'True'
      lastUpdateTime: '2020-06-29T14:12:03Z'
      lastTransitionTime: '2020-06-29T14:11:49Z'
      reason: ReplicaSetUpdated
      message: ReplicaSet "host-operator-684877b4f8" is progressing.
```

> And, no, if there is no operator pods running at all (even some old version) then there is no controller to check the status and it will be caught by the host status.

Yes, this is what I was pointing to - a case when there is some panic in the member-operator after the old replica was replaced. So, I'm curious - if we want to check the `Deployment`/`Pods` of the member operator from the host anyway, why don't we check everything there? Then we would have only one place (in host operator) where we check all member operator deployments. And vice versa, we could check from the member operator the deployment of the host operator.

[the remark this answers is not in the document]
MatousJobanek: In addition, if I break the deployment by deleting the ServiceAccount - which is the most frequest issue we've been facing for now - then there is no pod running of course. I believe that we should take care of this case for both operators - host and member.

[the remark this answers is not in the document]
alexeykazakov: > if we want to check the Deployment/Pods of the member operator from the host anyway

We don't want to check member operator deployments/pods from the host. Member deployments/pods/etc are  implementation details of the member status thing. Host status will only make sure the member is READY **and** the update time of the status is not too old. If there is no controller to update the member status then the update time will be outdated.

But I wonder if we want to rename "updateTime" to "lastProbeTime" so it's obvious what this field for.

> In addition, if I break the deployment by deleting the ServiceAccount - which is the most frequest issue we've been facing for now - then there is no pod running of course. I believe that we should take care of this case for both operators - host and member.

It's covered by the updateTime/lastProbeTime timestamp too.

pkg/controller/memberstatus/creation_test.go
```diff
@@ -0,0 +1,53 @@
...
+		// when
+		err = CreateOrUpdateResources(cl, s, MemberOperatorNs)
+
+		// then
+		require.NoError(t, err)
+
+		// check that the MemberStatus exists now
+		memberStatus := newMemberStatus()
+		err = cl.Get(context.TODO(), NamespacedName(MemberOperatorNs, defaultMemberStatusName), memberStatus)
+		require.NoError(t, err)
```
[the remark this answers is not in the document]
rajivnathan: Sure, that works. Done in 6f77a0e80d789b3953b2e1269a1d34f1f1effd7d

pkg/controller/memberstatus/creation.go
```diff
@@ -0,0 +1,23 @@
...
+
+func CreateOrUpdateResources(client client.Client, s *runtime.Scheme, namespace string) error {
+	memberStatus := &v1alpha1.MemberStatus{
+		ObjectMeta: v1.ObjectMeta{
+			Namespace: namespace,
+			Name:      defaultMemberStatusName,
+		},
+		Spec: v1alpha1.MemberStatusSpec{},
+	}
+	commonclient := commonclient.NewApplyClient(client, s)
```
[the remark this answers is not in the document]
rajivnathan: Good catch. Done in 6f77a0e80d789b3953b2e1269a1d34f1f1effd7d

pkg/controller/memberstatus/memberstatus_controller.go
rajivnathan: I thought about that but by the time I did I already made the updates to the api repo and didn't want to go through that hell again just for these strings so I'll just do it when I update the api for the status changes on the host side. Added it to the completion steps for the host operator issue.

pkg/controller/memberstatus/memberstatus_controller.go
rajivnathan: Done and Done in 6f77a0e80d789b3953b2e1269a1d34f1f1effd7d

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,305 @@
...
+		err := handler.handleStatus(reqLogger, memberStatus)
+		if err != nil {
+			reqLogger.Error(err, "status update problem")
+			unreadyComponents = append(unreadyComponents, string(handler.Name))
+		}
+	}
+
+	// if any components were not ready then set the overall status to not ready
+	if len(unreadyComponents) > 0 {
+		err := fmt.Errorf("Components not ready: %v", unreadyComponents)
```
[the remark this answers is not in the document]
rajivnathan: Done in 6f77a0e80d789b3953b2e1269a1d34f1f1effd7d

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,305 @@
...
+func (r *ReconcileMemberStatus) hostConnectionHandleStatus(reqLogger logr.Logger, memberStatus *toolchainv1alpha1.MemberStatus) error {
+	// look up host connection status
+	fedCluster, ok := r.getHostCluster()
+	if !ok {
+		notFoundMsg := errMsgHostConnectionNotFound
+		notFoundReason := reasonHostConnectionNotFound
+		memberStatus.Status.HostConnection = kubefed_v1beta1.KubeFedClusterStatus{
+			Conditions: []kubefed_v1beta1.ClusterCondition{
+				{
+					Type:          "Ready",
```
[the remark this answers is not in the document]
rajivnathan: Done in 6f77a0e80d789b3953b2e1269a1d34f1f1effd7d

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,305 @@
...
+			return reconcile.Result{}, nil
+		}
+		reqLogger.Error(err, "Unable to fetch MemberStatus resource")
+		return reconcile.Result{}, err
+	}
+
+	err = r.aggregateAndUpdateStatus(reqLogger, memberStatus)
+	if err != nil {
+		reqLogger.Error(err, "Failed to update status")
+	}
```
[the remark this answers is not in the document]
rajivnathan: I thought about this but I was worried because I read that it will requeue immediately. BUT it looks like if I do `return reconcile.Result{RequeueAfter: defaultRequeueTime}, err` then it respects the  `RequeueAfter` time. I tested it out and it works.

Done in 6f77a0e80d789b3953b2e1269a1d34f1f1effd7d

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,305 @@
...
+			},
+		}
+		return fmt.Errorf(notFoundMsg)
+	}
+	memberStatus.Status.HostConnection = *fedCluster.ClusterStatus.DeepCopy()
+
+	// check conditions of host connection
+	if !util.IsClusterReady(fedCluster.ClusterStatus) {
+		return fmt.Errorf("the host connection is not ready")
+	}
```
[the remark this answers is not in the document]
rajivnathan: This is a valid point. However, the design for the toolchain status centres around the host. Its usefulness is to be able to determine the status from a single point (the toolchainstatus resource available on the host). So for the case where there's a problem with the host deployment I think it should be evident that there's a problem because its status would not have been updated for a long time?

pkg/controller/memberstatus/memberstatus_controller.go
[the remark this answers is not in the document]
rajivnathan: +1 I like the idea of `lastProbeTime`. I'll make the change to the api as part of the status changes on the host side and will update the member as well.

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,305 @@
...
+	return nil
+}
+
+// updateStatusConditions updates Member status conditions with the new conditions
+func (r *ReconcileMemberStatus) updateStatusConditions(memberStatus *toolchainv1alpha1.MemberStatus, newConditions ...toolchainv1alpha1.Condition) error {
+	// the controller should always update at least the last updated timestamp of the status so the status should be updated regardless of whether
+	// any specific fields were updated. This way a problem with the controller can be indicated if the last updated timestamp was not updated.
+	conditionsWithTimestamps := []toolchainv1alpha1.Condition{}
+	for _, condition := range newConditions {
+		condition.LastTransitionTime = metav1.Now()
```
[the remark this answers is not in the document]
rajivnathan: For pods:
The lastTransitionTime field provides a timestamp for when the Pod last transitioned from one status to another.

The way I thought of it is for memberstatus is we are constantly updating the status (eg. the timestamp for kubefed lastProbeTime changes). I think `LastTransitionTime` is mostly useless for memberstatus and I would get rid of it altogether but it's not an optional field unfortunately so I just kept it as is for now.

pkg/controller/memberstatus/memberstatus_controller.go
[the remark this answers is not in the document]
alexeykazakov: Member status should not care about the host operator deployment or the host operator status at all. It's for the host status to take care of it.
Member status is interested in the host kubefedcluster to make sure the host cluster is available (regardless of the host operator status in the cluster). So, yes, checking the `lastProbeTime` of the host kubefedcluster does make sense in the member status.

pkg/controller/memberstatus/memberstatus_controller.go
MatousJobanek: If you want to move the constants to api repo later then I'm fine, but you will need to change it in the member-operator repo (and go through that hell again) anyway 
But it's up to you ;-)

pkg/controller/memberstatus/memberstatus_controller.go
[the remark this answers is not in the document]
MatousJobanek: excellent solution :+1: :1st_place_medal:

pkg/controller/memberstatus/memberstatus_controller.go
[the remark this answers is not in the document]
MatousJobanek: yeah, I forgot that the status info will be put together in the host-operator and not added from the member-operator side. So I agree with you :+1:

pkg/controller/memberstatus/memberstatus_controller.go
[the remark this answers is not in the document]
MatousJobanek: fair enough :+1: 
btw, shouldn't we check also the `unavailableReplicas` field in the Deployment status?

pkg/controller/memberstatus/memberstatus_controller.go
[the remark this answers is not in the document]
MatousJobanek: yeah, that makes sense :+1:

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,295 @@
...
+// general memberstatus constants
+const (
+	defaultRequeueTime = time.Second * 5
+)
+
+// custom component condition error messages & reasons
+const (
+	// deployment not found
+	reasonNoDeployment         = "DeploymentNotFound"
+	errMsgNoMemberOperatorName = "unable to look up the member operator name, the environment variable OPERATOR_NAME is not set"
```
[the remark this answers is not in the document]
rajivnathan: good catch! done in 221af19daa2191042bc55a246324ae5bbdb5fa23

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -0,0 +1,295 @@
...
+		reqLogger.Error(err, "Failed to update status")
+		return reconcile.Result{RequeueAfter: defaultRequeueTime}, err
+	}
+
+	reqLogger.Info(fmt.Sprintf("Finished updating MemberStatus, requeueing after %v", defaultRequeueTime))
+	return reconcile.Result{RequeueAfter: defaultRequeueTime}, nil
+}
+
+type statusHandler struct {
+	Name         statusComponentTag
```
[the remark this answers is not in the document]
rajivnathan: good point done in 221af19daa2191042bc55a246324ae5bbdb5fa23

pkg/controller/memberstatus/memberstatus_controller.go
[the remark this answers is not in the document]
rajivnathan: I updated it to now check `unavailableReplicas` and report an not ready status in that case as well

```
status:
  conditions:
  - lastTransitionTime: "2020-06-30T10:03:58Z"
    lastUpdatedTime: "2020-06-30T10:03:58Z"
    message: 'components not ready: [memberOperator]'
    reason: ComponentsNotReady
    status: "False"
    type: Ready
  hostConnection:
    conditions:
    - lastProbeTime: "2020-06-30T10:03:48Z"
      lastTransitionTime: "2020-06-30T10:02:48Z"
      message: /healthz responded with ok
      reason: ClusterReady
      status: "True"
      type: Ready
    region: ""
  memberOperator:
    buildTimestamp: "2020-06-30T20:14:04Z"
    conditions:
    - lastTransitionTime: "2020-06-30T10:03:58Z"
      lastUpdatedTime: "2020-06-30T10:03:58Z"
      message: the member operator deployment has unavailable replicas
      reason: DeploymentUnavailableReplicas
      status: "False"
      type: Ready
    deployment:
      name: member-operator
    revision: e38ad2d0883e3295ae32a26d96c04dfc7a9cb57c-dirty
    version: 0.0.1
```

pkg/controller/memberstatus/memberstatus_controller.go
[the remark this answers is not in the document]
rajivnathan: I added a check for the lastProbeTime to ensure it is within limits. It should be updated less than `(period + timeout) * failure threshold` seconds ago

Example of the lastProbeTime status condition for the hostConnection:
```
status:
  conditions:
  - lastTransitionTime: "2020-06-30T09:48:29Z"
    lastUpdatedTime: "2020-06-30T09:48:29Z"
    message: 'components not ready: [memberOperator hostConnection]'
    reason: ComponentsNotReady
    status: "False"
    type: Ready
  hostConnection:
    conditions:
    - lastProbeTime: "2020-06-30T09:48:19Z"
      lastTransitionTime: "2020-06-30T09:46:59Z"
      message: /healthz responded with ok
      reason: ClusterReady
      status: "True"
      type: Ready
    - lastProbeTime: "2020-06-30T09:48:19Z"
      message: 'exceeded the maximum duration since the last probe: 39s'
      reason: KubefedLastProbeTimeExceeded
      status: "False"
      type: ProbeWorking
    region: ""
  memberOperator:
    buildTimestamp: "2020-06-30T19:58:59Z"
    conditions:
    - lastTransitionTime: "2020-06-30T09:48:29Z"
      lastUpdatedTime: "2020-06-30T09:48:29Z"
      message: the member operator deployment has unavailable replicas
      reason: DeploymentUnavailableReplicas
      status: "False"
      type: Ready
    deployment:
      name: member-operator
    revision: e38ad2d0883e3295ae32a26d96c04dfc7a9cb57c-dirty
    version: 0.0.1
```

rajivnathan:
@alexeykazakov @MatousJobanek can you please take another look when you have a chance? I believe all the comments have been addressed now. I have added additional checks for the `lastProbeTime` for kubefed and `unavailableReplicas` for the member deployment as well as updated the tests. See the individual comments above for examples of how they look.

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -226,24 +280,32 @@ func (r *ReconcileMemberStatus) memberOperatorHandleStatus(reqLogger logr.Logger
...
+	}
 
-	// get and check conditions of member deployment
-	conditionsReady := true
-	deploymentConditions := []appsv1.DeploymentCondition{}
+	// get and check conditions
 	for _, condition := range memberDeployment.Status.Conditions {
-		deploymentConditions = append(deploymentConditions, condition)
-		conditionsReady = conditionsReady && condition.Status == corev1.ConditionTrue
+		if condition.Status != corev1.ConditionTrue {
```
alexeykazakov, line 295: Here is the list of current deployment condition types:
https://github.com/kubernetes/kubernetes/blob/d8fd232ea1f8c91092fb5fabb7a0f3d557a2e8fb/staging/src/k8s.io/api/apps/v1beta1/types.go#L485-L495
So, theoretically there can be `DeploymentReplicaFailure` condition with status `False`. Even if it's not what the current kube does the implementation could change in the future versions. I think it's safer to check specific condition types: `DeploymentAvailable` and `DeploymentProgressing`.

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -226,24 +280,32 @@ func (r *ReconcileMemberStatus) memberOperatorHandleStatus(reqLogger logr.Logger
 		memberStatus.Status.MemberOperator = operatorStatus
 		return err
 	}
+	operatorStatus.Deployment.Name = memberDeployment.Name
+
+	// check replicas
+	if memberDeployment.Status.UnavailableReplicas > 0 {
```
alexeykazakov, line 286: Have you checked that `UnavailableReplicas > 0` only if there is a problem with some pods?
Let's make sure `UnavailableReplicas` is 0 in case of normal a rolling update when a new pod is still starting but not fully ready yet.

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -194,6 +210,44 @@ func (r *ReconcileMemberStatus) hostConnectionHandleStatus(reqLogger logr.Logger
...
+	}
+
+	// check that the last probe time is within limits. It should be less than (period + timeout) * threshold
+	period := r.config.GetClusterHealthCheckPeriod()
+	timeout := r.config.GetClusterHealthCheckTimeout()
+	threshold := r.config.GetClusterHealthCheckFailureThreshold()
+	totalf := (period.Seconds() + timeout.Seconds()) * float64(threshold)
+	maxDuration, err := time.ParseDuration(fmt.Sprintf("%fs", totalf))
+	if err != nil {
+		return fmt.Errorf("the maximum duration since the last probe could not be determined - check the configured %s %s and %s", crtCfg.ClusterHealthCheckPeriod, crtCfg.ClusterHealthCheckTimeout, crtCfg.ClusterHealthCheckFailureThreshold)
```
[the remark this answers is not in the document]
rajivnathan: done in 89623fdd8fbeb694d378bd2d0e1c99aa567c3d55

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -194,6 +210,44 @@ func (r *ReconcileMemberStatus) hostConnectionHandleStatus(reqLogger logr.Logger
...
+	}
+
+	lastProbedTimePlusMaxDuration := lastProbeTime.Add(maxDuration)
+	currentTime := time.Now()
+	if currentTime.After(lastProbedTimePlusMaxDuration) {
+		errMsg := fmt.Sprintf("%s: %s", errMsgHostConnectionLastProbeTimeExceeded, maxDuration.String())
+		errReason := reasonHostConnectionLastProbeTimeExceeded
+		probeCondition := kubefed_v1beta1.ClusterCondition{
+			Type:          "ProbeWorking",
```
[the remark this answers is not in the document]
rajivnathan: I think overriding it is fine actually. done in 89623fdd8fbeb694d378bd2d0e1c99aa567c3d55

pkg/controller/memberstatus/memberstatus_controller.go
rajivnathan: Updated to check the specific condition types. done in 89623fdd8fbeb694d378bd2d0e1c99aa567c3d55

pkg/controller/memberstatus/memberstatus_controller.go
rajivnathan: Looks like during a normal rolling update `UnavailableReplicas` can be >0. Not sure how we want to handle this situation. Maybe we can go back to only checking the conditions and create a separate issue to come back to this to beef up how we detect problems with deployment rollouts?

alexeykazakov: > Maybe we can go back to only checking the conditions and create a separate issue to come back to this to beef up how we detect problems with deployment rollouts?

+1. Let's do it.

rajivnathan: Done

PR #181, opened 2020-07-24
Use new member status api for host connection and update component error handling

rajivnathan:
Related PR: https://github.com/codeready-toolchain/toolchain-common/pull/118

This PR updates the member status controller to use the new memberStatus.Status.Host field which now uses Toolchain Conditions instead of Kubefed conditions. It also updates to use `ValidateComponentConditionReady` to get errors directly from the component conditions.

### DIFF
--- a/pkg/controller/memberstatus/memberstatus_controller.go
+++ b/pkg/controller/memberstatus/memberstatus_controller.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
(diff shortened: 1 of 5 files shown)


pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -166,8 +167,11 @@ func (r *ReconcileMemberStatus) hostConnectionHandleStatus(reqLogger logr.Logger
 	}
 
 	// look up host connection status
-	clusterStatus, err := status.GetKubefedConditions(attributes)
-	memberStatus.Status.HostConnection = clusterStatus
+	connectionConditions := status.GetKubefedConditions(attributes)
```
xcoulon, line 170: I'm late on this, but the PR description says

> This PR updates the member status controller to use the new memberStatus.Status.Host field which now uses Toolchain Conditions instead of Kubefed conditions.

but we're still using `status.GetKubefedConditions(attributes)`? I'm probably missing something 🤔

PR #182, opened 2020-07-27
IdP name configuration

tinakurian:
making IdP configurable via configmap

Jira: https://issues.redhat.com/browse/CRT-743
Relates to: codeready-toolchain/toolchain-common#119


go.mod
```diff
@@ -26,6 +26,7 @@ require (
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.0.0+incompatible
+	github.com/codeready-toolchain/toolchain-common => github.com/tinakurian/toolchain-common v0.0.0-20200727140128-cfe3823faae8
```
tinakurian, line 29: reminder to myself!

deploy/env/e2e-tests.yaml
```diff
@@ -1,3 +1,2 @@
 member-operator:
-  identity-provider: 'testIdP'
```
MatousJobanek, line 2: Correct me if I'm wrong, but you should not remove these values from the env config files. This is btw the reason why your e2e tests are failing.
I understand that you need to primarily use the value from the ConfigMap, but you can still combine them and replace it with the value from CM, right?
Or, you could rename it to `default-identity-provider` and use this value as the default one if no value is specified in the CM.

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -33,7 +33,9 @@ func TestReconcile(t *testing.T) {
 	logf.SetLogger(logf.ZapLogger(true))
 	username := "johnsmith"
 	userID := uuid.NewV4().String()
-	config := configuration.LoadConfig()
+	cfg, err := configuration.LoadConfig(test.NewFakeClient(t))
+	require.NoError(t, err)
+	config := cfg
```
MatousJobanek, line 38: could you please tell me why you reassign the config variable? Wouldn't it be shorter to just assign it directly to `config`?:
```suggestion
	config, err := configuration.LoadConfig(test.NewFakeClient(t))
	require.NoError(t, err)
```

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -841,7 +843,8 @@ func TestDisabledUserAccount(t *testing.T) {
 	s := scheme.Scheme
 	err := apis.AddToScheme(s)
 	require.NoError(t, err)
-	config := configuration.LoadConfig()
+	cfg, err := configuration.LoadConfig(test.NewFakeClient(t))
+	config := cfg
```
MatousJobanek, line 847: same here, plus you forgot to check the error:
```suggestion
	config, err := configuration.LoadConfig(test.NewFakeClient(t))
	require.NoError(t, err)
```

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -1211,7 +1214,9 @@ func prepareReconcile(t *testing.T, username string, initObjs ...runtime.Object)
 	err := apis.AddToScheme(s)
 	require.NoError(t, err)
 	fakeClient := test.NewFakeClient(t, initObjs...)
-	config := configuration.LoadConfig()
+	cfg, err := configuration.LoadConfig(fakeClient)
+	require.NoError(t, err)
+	config := cfg
```
MatousJobanek, line 1219: same here 
```suggestion
	config, err := configuration.LoadConfig(fakeClient)
	require.NoError(t, err)
```

pkg/configuration/configuration_test.go
```diff
@@ -47,6 +51,64 @@ func TestGetAllMemberParameters(t *testing.T) {
...
+	t.Run("default", func(t *testing.T) {
+		// when
+		config := getDefaultConfiguration(t)
+
+		// then
+		assert.Equal(t, "rhd", config.GetIdP())
+	})
+	t.Run("env overwrite", func(t *testing.T) {
+		// given
+		restore := test.SetEnvVarAndRestore(t, "MEMBER_OPERATOR_CONFIG_MAP_NAME", "test-config")
```
MatousJobanek, line 66: we should set and restore also the variables retrieved from the ConfigMap:
```suggestion
		restore := test.SetEnvVarsAndRestore(t,
			test.Env("MEMBER_OPERATOR_CONFIG_MAP_NAME", "test-config"),
			test.Env("MEMBER_OPERATOR_IDENTITY_PROVIDER", ""),
			test.Env("MEMBER_OPERATOR_TEST_TEST", ""))
```

deploy/env/e2e-tests.yaml
tinakurian: @MatousJobanek shouldn't the default be set via viper? the same way we do for other values that are being set via the configMap?

pkg/controller/useraccount/useraccount_controller_test.go
tinakurian: yeah, definitely didn't mean to do that, must have been a bit tired when i did this lol, will fix

tinakurian: fixed here: a7cfb415bcaa14a4fe90ea746eb7dfc59185eb65

pkg/controller/useraccount/useraccount_controller_test.go
tinakurian: fixed here: a7cfb415bcaa14a4fe90ea746eb7dfc59185eb65 and here: c5066c27bced708ae2f020d83113d041fbe39690

pkg/controller/useraccount/useraccount_controller_test.go
tinakurian: fixed here: a7cfb415bcaa14a4fe90ea746eb7dfc59185eb65

pkg/configuration/configuration_test.go
tinakurian: fixed here: 8a2e7be9cb97a69f7afb5490c1ab47e251946fda

deploy/env/e2e-tests.yaml
MatousJobanek: It's up to us how we deal with the default values - if we use viper functionality or some other way...
I see that you changed the IdP name to `rhd` for the e2e tests - this is a feasible solution as well and much easier :+1:

pkg/configuration/configuration.go
```diff
@@ -50,8 +48,13 @@ const (
 
 	ClusterUnavailableDelay        = "cluster.unavailable.delay"
 	DefaultClusterUnavailableDelay = "60s"
+
+	identityProviderName        = "identity.provider"
+	DefaultIdentityProviderName = "rhd"
```
MatousJobanek, line 53: it's just a detail - not sure why you move it - you put the constants to the part containing KubeFed configuration constants

pkg/configuration/configuration.go
```diff
@@ -50,8 +48,13 @@ const (
...
 )
 
+var log = logf.Log.WithName("configuration")
```
MatousJobanek, line 56: not used 
```suggestion
```

pkg/configuration/configuration.go
tinakurian: fixed here: cecef2d2b60d85e6e54793c4d68db1168ee8c449

pkg/configuration/configuration.go
tinakurian: fixed here: cecef2d2b60d85e6e54793c4d68db1168ee8c449

pkg/configuration/configuration.go
tinakurian: @MatousJobanek i just noticed that all the kubefed vars are exported - is there any reason for this? i just checked and it doesn't look like they are used anywhere other than in configuration. if there is no reason, i can fix this in this PR

MatousJobanek: You can keep it there - I'm going to remove most of the constants in the next PRs anyway.

PR #191, opened 2020-08-21
add comment in case of UserAccount deletion

xcoulon:


pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -144,6 +144,11 @@ func (r *ReconcileUserAccount) Reconcile(request reconcile.Request) (reconcile.R
 		}
 	} else if util.HasFinalizer(userAcc, toolchainv1alpha1.FinalizerName) && util.IsBeingDeleted(userAcc) {
 		logger.Info("terminating UserAccount")
+		// We need to be sure that the status is updated in the when the UserAccount is deleted.
```
MatousJobanek, line 147:
```suggestion
		// We need to be sure that the status is updated when the UserAccount is deleted.
```

xcoulon: ah thanks!

PR #195, opened 2020-09-03
chore: add script that releases new version of member operator in app-sre-staging channel

MatousJobanek:
depends on the changes in https://github.com/codeready-toolchain/api/pull/184

Introduces  `app-sre/release-staging.sh` script that releases a new version of toolchain-member-operator in a `app-sre-staging` channel. As a result, there will be a separated index image for every release. It expects that the script is triggered for every commit to master.

cc: @jmelis


MatousJobanek:
@jmelis I added the Template and `oc apply` execution - I tested it in my temp cluster - see f983475

alexeykazakov (approved):

MatousJobanek:
@jmelis Please take a look at the latest state of the script - it should solve the problem with building the index images.

PR #198, opened 2020-09-10
feat: load resource usage and store it in MemberStatus

MatousJobanek:
Adds loading of Resource usage in member clusters. The usage is loaded from `NodeMetrics` resource, the allocatable capacity is taken from `Node.status`. 
The usage is counted per node role (worker, master). 
The MemberStatus then contains the percentage of the memory usage in all nodes of the given role.

Paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/161


alexeykazakov (approved):

deploy/cluster_role.yaml
```diff
@@ -19,6 +19,12 @@ rules:
   - list
   - get
   - watch
+- apiGroups:
+  - metrics.k8s.io
+  resources:
+  - "*"
+  verbs:
+  - "*"
```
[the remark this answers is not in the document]
MatousJobanek: right! Changed in 9717cd5

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -200,11 +205,101 @@ func (r *ReconcileMemberStatus) memberOperatorHandleStatus(reqLogger logr.Logger
...
+	nodeMetricsList := &v1beta1.NodeMetricsList{}
+	if err := r.client.List(context.TODO(), nodeMetricsList); err != nil {
+		return err
+	}
+
+	usagePerRole := map[string]float32{}
+	allocatablePerRole := map[string]float32{}
+NodeMetricsList:
+	for _, nodeMetric := range nodeMetricsList.Items {
+		for key, usage := range nodeMetric.Usage {
```
xcoulon, line 225: can't you just do `if usage, exists := nodeMetric.Usage["memory"]; exists { ... }` instead?

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -200,11 +205,101 @@ func (r *ReconcileMemberStatus) memberOperatorHandleStatus(reqLogger logr.Logger
...
+				} else {
+					reqLogger.Info("skipping NodeMetrics resource - there wasn't found corresponding node that would be monitored", "name", nodeMetric.Name)
+				}
+				continue NodeMetricsList
+			}
+		}
+		return fmt.Errorf("memory item not found in NodeMetrics: %v", nodeMetric)
+	}
+
+	if len(allocatableValues) > 0 {
```
xcoulon, line 242: could you add a comment to explain what you're doing here (for the sake of maintenance)

xcoulon: also, maybe rename `allocatableValues` to `pendingNodes` to explain that it's a list from which you remove entries as you get the `memory` usage from the metrics?

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -200,11 +205,101 @@ func (r *ReconcileMemberStatus) memberOperatorHandleStatus(reqLogger logr.Logger
...
+		var nodeNames []string
+		for nodeName := range allocatableValues {
+			nodeNames = append(nodeNames, nodeName)
+		}
+		return fmt.Errorf("missing NodeMetrics resource for Nodes: %v", nodeNames)
+	}
+
+	for role, usage := range usagePerRole {
```
xcoulon, line 250: I believe you could set `memberStatus.Status.ResourceUsage.MemoryUsagePerNodeRole[role]` on line 229 without having to use a temporary map. Unless you want to set all memory usage if no error was reported on line 247?

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -200,11 +205,101 @@ func (r *ReconcileMemberStatus) memberOperatorHandleStatus(reqLogger logr.Logger
...
+		return nil, errs.Wrapf(err, "unable to list Nodes")
+	}
+	allocatableValues := map[string]nodeInfo{}
+	for _, node := range nodes.Items {
+		role := getNodeRole(node)
+		if role == "" {
+			reqLogger.Info("The node doesn't have role worker nor master - is ignored in resource consumption computing", "nodeName", node.Name)
+			continue
+		}
+		for key, value := range node.Status.Allocatable {
```
xcoulon, line 269: same here, I assume you could use `if memory, exists := node.Status.Allocatable["memory"]; exists { ... }`

test/memberstatus_assertion.go
```diff
@@ -71,6 +72,29 @@ func (a *MemberStatusAssertion) HasHostConditionErrorMsg(expected string) *Membe
...
 
+type AddUsage func(map[string]int)
+
+func OfNodeRole(role string, usage int) AddUsage {
+	return func(nodesUsage map[string]int) {
+		nodesUsage[role] = usage
+	}
+}
+
+func (a *MemberStatusAssertion) HasMemoryUsage(usages ...AddUsage) *MemberStatusAssertion {
```
xcoulon, line 83: could be renamed to `HasMemoryUsages` (plural) but that's a minor detail

xcoulon (commented):
looks good overall. Just a few questions/suggestions

pkg/controller/memberstatus/memberstatus_controller.go
MatousJobanek: good point :man_facepalming: 
done in 53280b6
thanks for it :+1:

pkg/controller/memberstatus/memberstatus_controller.go
MatousJobanek: done in 53280b6

test/memberstatus_assertion.go
MatousJobanek: I thought that the word `usage` should be used as an uncountable word in this context :man_shrugging:

pkg/controller/memberstatus/memberstatus_controller.go
MatousJobanek: I added a comment here: e9a2517
the `pendingNodes` doesn't really explain what the map is really used for (from my point of view)

pkg/controller/memberstatus/memberstatus_controller.go
MatousJobanek: No, I cannot to do that. In the temporary map I store the sum of all nodes of the same role. Then I can count the ratio.

test/memberstatus_assertion.go
xcoulon: yeah, on second thought, you're probably right ;)

pkg/controller/memberstatus/memberstatus_controller.go
xcoulon: ah ok, I missed that. Maybe it's worth adding a comment, then?

pkg/controller/memberstatus/memberstatus_controller.go
xcoulon: ok, fair enough. A comment to explain the purpose is good, too.

xcoulon (approved):
thanks for addressing my comments, @MatousJobanek!
I would suggest adding a comment in https://github.com/codeready-toolchain/member-operator/pull/198/files#r488657892 to explain that memory usage is cumulated by node type?

pkg/controller/memberstatus/memberstatus_controller.go
MatousJobanek: done be025f9

PR #199, opened 2020-09-15
chore: Use predicate from toolchain-common

rajivnathan:
This is the follow-up change to https://github.com/codeready-toolchain/toolchain-common/pull/128


MatousJobanek (approved):
Looks good, but could you please tell me for which use-case you want to use it in host-operator?

MatousJobanek:
never mind, now I understand :-) https://github.com/codeready-toolchain/host-operator/pull/281#discussion_r488061024

PR #200, opened 2020-10-01
filter namespaces with predicate based on name 

xcoulon:
new custom predicate to filter out namespaces whose name starts with `openshift` or `kube`


alexeykazakov (changes requested):
This PR breaks the idler. I wish we had a e2e test to catch that but we currently don't have tests which actually run the idler :(

xcoulon:
closing this PR for now. Not needed since we changed the Idler controller logic in https://github.com/codeready-toolchain/member-operator/pull/205 (requeuing and not watching pods anymore)

PR #201, opened 2020-10-01
Idler: don't watch all namespaces/pods

alexeykazakov:
This is an implementation of the @MatousJobanek's idea of watching the Idler only and not watching pods at all. So, we could stop watching all namespaces for now.

TODO:
- [x] adjust tests

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -108,7 +108,7 @@ func main() {
 
 	// Set default manager options
 	options := manager.Options{
-		Namespace:          "", // Watch all namespaces! Idler needs to watch Pods from all namespaces
+		Namespace:          namespace,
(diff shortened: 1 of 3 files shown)


MatousJobanek (commented):
Awesome :1st_place_medal: 
Thanks a lot for this quick fix :+1:

xcoulon (approved):
This is a huge improvement in term of memory consumption without sacrificing the feature itself, so kudos for such a change 👍

MatousJobanek (approved):
Good job :+1: 
I've just renamed one test to reflect the current state (if you don't mind :-)) https://github.com/codeready-toolchain/member-operator/pull/201/commits/74d1f19b007c66a2847b488c1ae1a141cf23a801

alexeykazakov:
> I've just renamed one test to reflect the current state (if you don't mind :-))

Good catch.

PR #205, opened 2020-10-08
idler standalone client

xcoulon:
use a dedicated client for `IdlerController` which is configured to all namespaces.  
restore the watcher on pods

also, when deleting a job, make sure that the `delete` option includes child pods

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/179


pkg/controller/controller.go
```diff
@@ -20,11 +21,11 @@ func init() {
 	addToManagerFuncs = append(addToManagerFuncs, useraccount.Add)
 	addToManagerFuncs = append(addToManagerFuncs, useraccountstatus.Add)
 	addToManagerFuncs = append(addToManagerFuncs, nstemplateset.Add)
-	addToManagerFuncs = append(addToManagerFuncs, idler.Add)
+	// addToManagerFuncs = append(addToManagerFuncs, idler.Add)
```
alexeykazakov, line 24:
```suggestion
```

alexeykazakov (approved):
Looks good!

cmd/manager/main.go
```diff
@@ -108,19 +108,10 @@ func main() {
...
 
-	// Add support for MultiNamespace set in WATCH_NAMESPACE (e.g ns1,ns2)
-	// Note that this is not intended to be used for excluding namespaces, this is better done via a Predicate
-	// Also note that you may face performance issues when using this with a high number of namespaces.
-	// More Info: https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/cache#MultiNamespacedCacheBuilder
-	if strings.Contains(namespace, ",") {
-		options.Namespace = ""
-		options.NewCache = cache.MultiNamespacedCacheBuilder(strings.Split(namespace, ","))
-	}
-
```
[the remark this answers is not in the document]
xcoulon: ok, I understand what you mean. 
Done in 1de65b1

cmd/manager/main.go
```diff
@@ -167,13 +172,48 @@ func main() {
...
 	if err := mgr.Start(stopChannel); err != nil {
-		log.Error(err, "Manager exited non-zero")
+		log.Error(err, "Default manager exited non-zero")
 		os.Exit(1)
 	}
+
+	log.Info("Starting the Cmd.")
+}
+
+func newIdlerClient(cfg *rest.Config) (client.Client, cache.Cache, error) {
```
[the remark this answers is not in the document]
xcoulon: sure! done in c8ee366

cmd/manager/main.go
```diff
@@ -167,13 +182,53 @@ func main() {
...
 	}
+
+	log.Info("Starting the Cmd.")
+}
+
+// newIdlerClient creates a new client for the Idler Controller.
+// As opposed to the client used in other controllers, this one watches resources in all namespaces.
+// But since the Idler controller only cares about Idler resources (cluster-wide) and Pods (namespaced),
+// this client will not store all other namespaced resources (secrets, etc.).
+// This will help keeping a reasonnable memory usage for this operator.
```
MatousJobanek, line 204: small typo
```suggestion
// This will help keeping a reasonable memory usage for this operator.
```

pkg/apis/apis.go
```diff
@@ -18,21 +18,27 @@ import (
...
+		metrics.AddToScheme)
 
 	return addToSchemes.AddToScheme(s)
 }
+
+// AddToIdlerScheme adds all Resources to the Idler Scheme
+func AddToIdlerScheme(s *runtime.Scheme) error {
+	addToSchemes := append(apis.AddToSchemes, corev1.AddToScheme)
+	return addToSchemes.AddToScheme(s)
+}
```
MatousJobanek, line 44: This is not used anywhere
```suggestion
```

pkg/controller/idler/idler_controller.go
```diff
@@ -165,9 +168,11 @@ func (r *ReconcileIdler) ensureIdling(logger logr.Logger, idler *toolchainv1alph
 // and scales the owner down to zero and returns "true".
 // Otherwise returns "false".
 func (r *ReconcileIdler) scaleControllerToZero(logger logr.Logger, meta metav1.ObjectMeta) (bool, error) {
+	logger.Info("Scaling controller to zero", "name", meta.Name)
 	owners := meta.GetOwnerReferences()
 	for _, owner := range owners {
 		if owner.Controller != nil && *owner.Controller {
+			log.Info("owner", "owner", owner.Kind)
```
MatousJobanek, line 175: Do we still need this log? I think that it was added just for troubleshooting. If we need it, then please fix the content to log something more meaningful.

MatousJobanek (commented):
Nice solution. Great job @xcoulon !!! :+1: :1st_place_medal: 
However, I'm still missing the tests covering the changes in the controller

pkg/controller/idler/idler_controller.go
alexeykazakov: Yeah... I added most of these logs just for debugging. Not sure if we want to be so verbose in our idler logs.

cmd/manager/main.go
xcoulon: mind you, it takes 2 `n` in French! https://translate.google.com/#view=home&op=translate&sl=fr&tl=en&text=raisonnable 😂

pkg/apis/apis.go
xcoulon: good catch, it's a left-over that I should have removed!

pkg/controller/idler/idler_controller.go
xcoulon: removed in https://github.com/codeready-toolchain/member-operator/pull/205/commits/b587338166eb5aa34e9a3149cd7c90c321cdc27d

deploy/cluster_role.yaml
```diff
@@ -75,8 +75,7 @@ rules:
   resources:
   - routes
   verbs:
-  - "get"
-  - "list"
+  - "*"
```
xcoulon, line 78: arg! it was not meant to be part of this PR. I will revert this change

xcoulon: reverted in 41848cf

pkg/controller/idler/idler_controller.go
```diff
@@ -110,23 +110,18 @@ func (r *ReconcileIdler) Reconcile(request reconcile.Request) (reconcile.Result,
...
-		d = &timeout
-	}
-	log.Info("requeueing in", "duration", d)
+	requeue, after := nextPodToBeKilledAfter(idler)
+	log.Info("requeueing in", "duration", after)
 	return reconcile.Result{
-		Requeue:      true,
-		RequeueAfter: *d,
+		Requeue:      requeue,
+		RequeueAfter: after,
```
alexeykazakov, line 117: This is doesn't look right. We need to requeue with `RequeAfter` > 0 if there is no pods to track. In this case the requeueAfter should be equal to the idler timeout. Remember, there is no watcher to watch the pods.
This is exactly why we had the following code in the original PR:
```
	if d == nil || *d < time.Duration(0) {
		// No pods tracked. Requeue after the idler timout so we don't miss new pods created within the timeout.
		timeout := time.Duration(idler.Spec.TimeoutSeconds) * time.Second
		d = &timeout
	}
```
E2e should be failing now with this new change.

pkg/controller/idler/idler_controller.go
```diff
@@ -341,11 +336,13 @@ func findPodByName(idler *toolchainv1alpha1.Idler, name string) *toolchainv1alph
 	return nil
 }
 
-// nextPodToBeKilledAfter checks the start times of all the tracked pods in the Idler and returns the timeout left
+// nextPodToBeKilledAfter checks the start times of all the tracked pods in the Idler and returns `true` and the timeout left
 // for the next pod to be killed.
-func nextPodToBeKilledAfter(idler *toolchainv1alpha1.Idler) *time.Duration {
+// If there is no pod to kill, the func returns `false`, `0`
+func nextPodToBeKilledAfter(idler *toolchainv1alpha1.Idler) (bool, time.Duration) {
```
alexeykazakov, line 341: IMO it's not up to this function to decide if there should be a requeue or not.
As a matter of fact there should be always a requeue. No matter what.

pkg/controller/idler/idler_controller.go
xcoulon: huh? there is a watcher for the pods again: https://github.com/codeready-toolchain/member-operator/pull/205/files#diff-0ddf448ea61959821185b66811f19adbd7a7cc10c772d44722bc0cffa81a425bR63

alexeykazakov: The initial idea was to drop pod watchers completely and just re-queue with a timeout left before the next pod to kill **or** the idler timeout (8h atm). This was still the case when I left my comment I think...
But now I see that you are restoring the watcher. We will need to test this new approach with some pods to see if there is any performance or memory usage issues.
My understanding is that you have tested the PR without watcher only. Is it correct?

alexeykazakov: @MatousJobanek wdyt? I'm not sure what approach is better (assuming we have very similar performance/memory usage results of course). Both approaches have pros and cons.

MatousJobanek: I like the new one more - requeue idler controller when no pod is found. It looks simpler to me:
* has only one watcher
* has no mapping (between pod and idler)
* there is always only one reconcile loop for one namespace (no less no more)
* no "error logs" for namespaces where idler resource is not present... 

But to be honest, I don't have strong opinion on it, so I don't insist on any of these solutions. I understand that the solution with the watcher may look to be more straightforward - so as you said, both approaches have pros and cons. 

I would let @xcoulon decide which solution he wants to go with since we are both fine with both of them :wink:

alexeykazakov: Yeah I'm leaning towards the no-watcher approach too. Don't insist either though :) But still would prefer the no-watcher option...

xcoulon: @alexeykazakov @MatousJobanek I've removed the usage of the pod watcher in c36a02a. When there is no pod in the namespace associated with the idler, then the controller will "requeue after" the value specified in `spec.TimeoutSeconds` to see if there are new pods that need to be tracked.

pkg/controller/idler/idler_controller.go
xcoulon: I've reverted the changes here in https://github.com/codeready-toolchain/member-operator/commit/c36a02a9cf89b85a24394cd04e6a2a7140890c51

pkg/controller/idler/idler_controller.go
xcoulon: also, if we drop the pod watcher, then we don't need https://github.com/codeready-toolchain/member-operator/pull/200 either (this latter was supposed to help filtering out reconcile loops based on pods in the `openshift-*` and `kube-*` namespaces)

alexeykazakov: Looks good overall now but the original code also took care of the case when `nextPodToBeKilledAfter()` returns a negative duration. Kube ignores requeue requests with a negative duration. So let's add this as safety net.

xcoulon: oh right, I've missed that! let me take care of it!

pkg/controller/idler/idler_controller.go
```diff
@@ -126,11 +107,20 @@ func (r *ReconcileIdler) Reconcile(request reconcile.Request) (reconcile.Result,
 			"failed to ensure idling '%s'", idler.Name)
 	}
 	// Find the earlier pod to kill and requeue. Do not requeue if no pods tracked
-	requeue, after := nextPodToBeKilledAfter(idler)
-	log.Info("requeueing in", "duration", after)
+	nextTime := nextPodToBeKilledAfter(idler)
+	if nextTime == nil {
```
alexeykazakov, line 111:
```suggestion
	if nextTime == nil || *nextTime <0 {
```

pkg/controller/idler/idler_controller.go
```diff
@@ -332,10 +349,12 @@ func findPodByName(idler *toolchainv1alpha1.Idler, name string) *toolchainv1alph
 	return nil
 }
 
-// nextPodToBeKilledAfter checks the start times of all the tracked pods in the Idler and returns the timeout left
+// nextPodToBeKilledAfter checks the start times of all the tracked pods in the Idler and returns `true` and the timeout left
```
alexeykazakov, line 352:
```suggestion
// nextPodToBeKilledAfter checks the start times of all the tracked pods in the Idler and the timeout left
```

pkg/controller/idler/idler_controller.go
```diff
@@ -332,10 +349,12 @@ func findPodByName(idler *toolchainv1alpha1.Idler, name string) *toolchainv1alph
...
 // for the next pod to be killed.
+// If there is no pod to kill, the func returns `false`, `0`
```
alexeykazakov, line 354:
```suggestion
// If there is no pod to kill, the func returns `nil`
```

pkg/controller/idler/idler_controller.go
alexeykazakov: https://github.com/codeready-toolchain/member-operator/pull/205#discussion_r505642209

xcoulon: done in c8a7c74

pkg/controller/idler/idler_controller.go
alexeykazakov: I added this suggestion before you added the check to `nextPodToBeKilledAfter()` ;)

pkg/controller/idler/idler_controller_test.go
```diff
@@ -102,7 +106,11 @@ func TestEnsureIdling(t *testing.T) {
 
 		// then
 		require.NoError(t, err)
-		assert.Equal(t, reconcile.Result{}, res)
+		// no need to requeue, the controller will reconcile when a pod is created
```
MatousJobanek, line 109: This comment is wrong I guess
```suggestion
		// no pods found - the controller will requeue after the idler's timeout
```

MatousJobanek (approved):
Awesome :+1:  Great job @xcoulon :1st_place_medal:

pkg/controller/idler/idler_controller_test.go
xcoulon: ah right, good catch! thanks @MatousJobanek!

PR #209, opened 2020-10-13
refactor: addressing longstanding linter issues

xcoulon:


pkg/controller/nstemplateset/nstemplatetier_test.go
```diff
@@ -162,7 +162,7 @@ func TestGetTierTemplate(t *testing.T) {
 		t.Run("host cluster not available", func(t *testing.T) {
 			// given
 			hostCluster := test.NewGetHostCluster(cl, false, apiv1.ConditionFalse)
-			defer resetCache()
+			resetCache()
```
MatousJobanek, line 165: why not calling it with `defer`? I want to reset the cache at the end of the test.

pkg/controller/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -1277,8 +1276,8 @@ func newIdler(username, name, tier string, options ...objectMetaOption) *toolcha
 		ObjectMeta: metav1.ObjectMeta{
 			Labels: map[string]string{
 				"toolchain.dev.openshift.com/provider":    "codeready-toolchain",
-				"toolchain.dev.openshift.com/tier":        tier,
-				"toolchain.dev.openshift.com/templateref": NewTierTemplateName(tier, "clusterresources", "abcde11"),
+				"toolchain.dev.openshift.com/tier":        "advanced",
+				"toolchain.dev.openshift.com/templateref": NewTierTemplateName("advanced", "clusterresources", "abcde11"),
```
MatousJobanek, line 1280: I would rather use `basic` as the default hard-coded tier, but it's fine to keep it as it is.

pkg/controller/nstemplateset/nstemplatetier_test.go
```diff
@@ -252,12 +252,8 @@ func TestGetTierTemplate(t *testing.T) {
...
 }
 
-func resetCache() func() {
-	reset := func() {
-		tierTemplatesCache = newTierTemplateCache()
-	}
-	reset()
-	return reset
+func resetCache() {
+	tierTemplatesCache = newTierTemplateCache()
```
MatousJobanek, line 256: the original idea was to call reset immediately (in case it wasn't reset before - for example someone could forget to do that at the end of the test) and then also at the end of the test

MatousJobanek (approved):

pkg/controller/nstemplateset/nstemplateset_controller_test.go
xcoulon: well, we always passed `advanced` so I just moved the value here. However, I followed your recommendation and changed to `basic` in cc44404

pkg/controller/nstemplateset/nstemplatetier_test.go
xcoulon: I thought that for this particular test, we needed to reset the cache before calling `getTierTemplate`?

pkg/controller/nstemplateset/nstemplatetier_test.go
xcoulon: to be honest, the fact that we have `defer resetCache()` in the test did not make it obvious that the cache was also reset at the beginning of the test :(

MatousJobanek: I know, it was confusing... keep it as it is.

pkg/controller/nstemplateset/nstemplatetier_test.go
MatousJobanek: It should not happen that the cache is not reset at the beginning of any test - that's why we always call it using `defer` to make sure that the cache is always cleaned at the end of the test.

tinakurian (approved):

PR #212, opened 2020-10-21
chore: make the client for all namespaces more generic

MatousJobanek:
This PR makes the client that watches all namespaces more generic. It's being passed to all controller as another client so it can be easily reached and used in MemberStatus controller. 
The PR also removes some unnecessary code.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -32,7 +32,7 @@ import (
 	"k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	apiutil "sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -137,19 +137,14 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Setup all Controllers
-	if err := controller.AddControllersToManager(mgr, crtConfig); err != nil {
-		log.Error(err, "")
-		os.Exit(1)
-	}
-
-	idlerClient, idlerClientCache, err := newClient(cfg)
+	allNamespacesClient, allNamespacesCache, err := newAllNamespacesClient(cfg)
 	if err != nil {
 		log.Error(err, "")
 		os.Exit(1)
 	}
 
-	if err := controller.AddIdlerControllerToManager(mgr, idlerClient, idlerClientCache); err != nil {
+	// Setup all Controllers
+	if err := controller.AddControllersToManager(mgr, crtConfig, allNamespacesClient); err != nil {
 		log.Error(err, "")
 		os.Exit(1)
 	}
@@ -163,10 +158,6 @@ func main() {
 		log.Info("Waiting for cache to sync")
 		if !mgr.GetCache().WaitForCacheSync(stopChannel) {
 			log.Error(fmt.Errorf("timed out waiting for main cache to sync"), "")
-			os.Exit(1)
-		}
-		if !idlerClientCache.WaitForCacheSync(stopChannel) {
-			log.Error(fmt.Errorf("timed out waiting for idler cache to sync"), "")
 			os.Exit(1)
 		}
 		log.Info("Starting ToolchainCluster health checks.")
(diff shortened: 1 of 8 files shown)


cmd/manager/main.go
```diff
@@ -165,10 +160,6 @@ func main() {
 			log.Error(fmt.Errorf("timed out waiting for main cache to sync"), "")
 			os.Exit(1)
 		}
-		if !idlerClientCache.WaitForCacheSync(stopChannel) {
```
alexeykazakov, line 168: We don't need this?

alexeykazakov (approved):
Nice!

cmd/manager/main.go
MatousJobanek: no, because we don't need the all-namespaces client (cache) in any of the subsequent "commands":
https://github.com/codeready-toolchain/member-operator/blob/17781a03f5505e6161ad1ef44cf6754a9d278e64/cmd/manager/main.go#L163-L173

xcoulon: ah right, I did not realize it was not necessary for the subsequent commands 🤦‍♂️

cmd/manager/main.go
```diff
@@ -197,20 +188,19 @@ func main() {
...
 }
 
-// newClient creates a new "custom" client for a controller.
-// As opposed to the client used in other controllers, this one watches resources in all namespaces.
-// But since the Idler controller only cares about Idler resources (cluster-wide) and Pods (namespaced),
-// this client will not store all other namespaced resources (secrets, etc.).
-// This will help keeping a reasonable memory usage for this operator.
-func newClient(cfg *rest.Config) (client.Client, cache.Cache, error) {
+// newAllNamespacesClient creates a new client that watches (as opposed to the standard client) resources in all namespaces.
+// This client should be used only for resources and kinds that are retrieved from other namespaces than the watched one is.
```
xcoulon, line 192:
```suggestion
// This client should be used only for resources and kinds that are retrieved from other namespaces than the watched one.
```

pkg/controller/idler/idler_controller.go
```diff
@@ -53,7 +53,7 @@ func add(mgr manager.Manager, r *ReconcileIdler) error {
 
 	// Watch for changes to primary resource Idler
 	if err := c.Watch(
-		source.NewKindWithCache(&toolchainv1alpha1.Idler{}, r.cache),
+		&source.Kind{Type: &toolchainv1alpha1.Idler{}},
```
xcoulon, line 56: hmm, I thought this was needed, but maybe not anymore after this refactoring? 🤔

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -30,7 +30,7 @@ import (
 var log = logf.Log.WithName("controller_nstemplateset")
 
 // Add creates a new NSTemplateSetReconciler and starts it (ie, watches resources and reconciles the cluster state)
-func Add(mgr manager.Manager, _ *configuration.Config) error {
+func Add(mgr manager.Manager, _ *configuration.Config, _ client.Client) error {
```
xcoulon, line 33: I understand the benefit of having the same `Add` func for all controllers, but it's a bit confusing that most of them do *not* use the given `client.Client`. Anyways, it's just my thoughts, not a request to change that.

xcoulon (approved):
Approved with 1 remark and one suggestion.

pkg/controller/idler/idler_controller.go
MatousJobanek: Originally, you used for watching the "all namespaces" client that is created on our own, not that one that was delivered via the Manager. In my code, I use the "standard" namespace-scoped client for watching the Idlers and the "all namespaces" one only for manipulating the payloads.

pkg/controller/nstemplateset/nstemplateset_controller.go
MatousJobanek: In the next PR, I'm going to use the client in the MemberStatus controller - that was also the reason why I was doing all the changes (you can read it in the description of the PR). So, there will be two controllers using this client.
BTW, you could raise the same concern regarding the `configuration.Config` param - this one is used only in two controllers as well. 
We are going to have four types of the constructors:
1. taking only Manager
2. taking Manager & Config
3. taking Manager & Client
4. taking Manager & Config & Client

So do you want to address all of them separately? Yeah, we could get rid of the `addToManagerFuncs` and call them directly in `AddControllersToManager`, but I'm probably too old for such changes :smile:

pkg/controller/idler/idler_controller.go
xcoulon: ah ok, I did not notice that. Makes sense now!

pkg/controller/nstemplateset/nstemplateset_controller.go
xcoulon: yes, the `config` param raises the same concern. It's just that the `client` feels like something more "low-level" vs the `config` which is a type of ours. But let's keep it that way for now.

PR #208, opened 2020-10-13
Add Console & Che URLs to Member status

alexeykazakov:
TODO:
- [ ] Use proper (all-namespaces) client to get routes
- [ ] Tests

PR #213, opened 2020-10-22
Routes in memberstatus

MatousJobanek:
Retrieves the Console and Che routes and stores them in `MemberStatus`. This will be used in the host operator instead of directly retrieving them from the member cluster.
Most of the commits were taken from https://github.com/codeready-toolchain/member-operator/pull/208

Paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/185

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.6.0
 )
 
+replace github.com/codeready-toolchain/api => github.com/alexeykazakov/api v0.0.0-20201013003743-eed55bb9315f
+
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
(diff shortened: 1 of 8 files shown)


alexeykazakov (approved):

alexeykazakov:
When done with this PR we will need to
- make sure the routes are propagated to the toolchain status
- switch reg-service to load the routes from the toolchain status instead of loading them from MURs
- clean up MUR controller (stop loading routes)

MatousJobanek:
right. To be honest, originally, I wanted to firstly change the host-operator logic and fill the `MUR.status` with the URLs from `ToolchainStatus` and change the reg-service later when I find some time, but I can also do it the other way around.

alexeykazakov:
> right. To be honest, originally, I wanted to firstly change the host-operator logic and fill the MUR.status with the URLs from ToolchainStatus and change the reg-service later when I find some time, but I can also do it the other way around.

That would work too.

PR #215, opened 2020-11-12
Add Che client with token cache

rajivnathan:
Adds a Che client to the member operator that does the following:

- looks up the Che Keycloak server
- uses credentials from the member-operator-secret to request a token from Che's Keycloak server
- retrieves the token from the server and caches it
- reuses the cached token until it expires

More functions will be added to che.go in PRs to come. (eg. Find user and Delete user functions)

Related issue: https://issues.redhat.com/browse/CRT-846

### DIFF
--- a/pkg/che/che.go
+++ b/pkg/che/che.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"fmt"
 
-	crtConfig "github.com/codeready-toolchain/member-operator/pkg/configuration"
+	crtcfg "github.com/codeready-toolchain/member-operator/pkg/configuration"
 
 	routev1 "github.com/openshift/api/route/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -19,13 +19,13 @@ var DefaultClient *Client
 
 // Client is a client for interacting with Che services
 type Client struct {
-	config     *crtConfig.Config
+	config     *crtcfg.Config
 	k8sClient  client.Client
 	tokenCache *tokenCache
 }
 
 // InitDefaultCheClient initializes the default Che service instance
-func InitDefaultCheClient(cfg *crtConfig.Config, cl client.Client) {
+func InitDefaultCheClient(cfg *crtcfg.Config, cl client.Client) {
 	DefaultClient = &Client{
 		config:     cfg,
 		k8sClient:  cl,
--- a/pkg/che/che.go
+++ b/pkg/che/che.go
@@ -1,13 +1,8 @@
 package che
 
 import (
-	"context"
-	"fmt"
-
 	crtcfg "github.com/codeready-toolchain/member-operator/pkg/configuration"
 
-	routev1 "github.com/openshift/api/route/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
@@ -31,18 +26,4 @@ func InitDefaultCheClient(cfg *crtcfg.Config, cl client.Client) {
 		k8sClient:  cl,
 		tokenCache: newTokenCache(),
 	}
-}
-
-func getRouteURL(cl client.Client, namespace, name string) (string, error) {
-	route := &routev1.Route{}
-	namespacedName := types.NamespacedName{Namespace: namespace, Name: name}
-	err := cl.Get(context.TODO(), namespacedName, route)
-	if err != nil {
-		return "", err
-	}
-	scheme := "https"
-	if route.Spec.TLS == nil || *route.Spec.TLS == (routev1.TLSConfig{}) {
-		scheme = "http"
-	}
-	return fmt.Sprintf("%s://%s/%s", scheme, route.Spec.Host, route.Spec.Path), nil
 }
--- a/pkg/che/che.go
+++ b/pkg/che/che.go
@@ -1,0 +1,48 @@
+package che
+
+import (
+	"context"
+	"fmt"
+
+	crtConfig "github.com/codeready-toolchain/member-operator/pkg/configuration"
+
+	routev1 "github.com/openshift/api/route/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var log = logf.Log.WithName("che-client")
+
+// DefaultClient is a default implementation of a CheClient
+var DefaultClient *Client
+
+// Client is a client for interacting with Che services
+type Client struct {
+	config     *crtConfig.Config
+	k8sClient  client.Client
+	tokenCache *tokenCache
+}
+
+// InitDefaultCheClient initializes the default Che service instance
+func InitDefaultCheClient(cfg *crtConfig.Config, cl client.Client) {
+	DefaultClient = &Client{
+		config:     cfg,
+		k8sClient:  cl,
+		tokenCache: newTokenCache(),
+	}
+}
+
+func getRouteURL(cl client.Client, namespace, name string) (string, error) {
+	route := &routev1.Route{}
+	namespacedName := types.NamespacedName{Namespace: namespace, Name: name}
+	err := cl.Get(context.TODO(), namespacedName, route)
+	if err != nil {
+		return "", err
+	}
+	scheme := "https"
+	if route.Spec.TLS == nil || *route.Spec.TLS == (routev1.TLSConfig{}) {
+		scheme = "http"
+	}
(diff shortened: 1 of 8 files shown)


pkg/che/tokencache.go
```diff
@@ -0,0 +1,153 @@
...
+
+func newTokenCache() *tokenCache {
+	return &tokenCache{
+		httpClient: newHTTPClient(),
+	}
+}
+
+// getToken returns the token needed to use Che user APIs, or an error if there was a problem getting the token
+func (tc *tokenCache) getToken(cl client.Client, cfg *crtCfg.Config) (TokenSet, error) {
+	defer tc.RUnlock()
```
alexeykazakov, line 40: This function not just reads the token. It also writes it in line 62: `obtainNewToken()`. So you can't just use RLock here.
You can use something like this: https://github.com/codeready-toolchain/devcluster/blob/master/pkg/ibmcloud/cluster.go#L64

pkg/che/che.go
```diff
@@ -0,0 +1,48 @@
+package che
+
+import (
+	"context"
+	"fmt"
+
+	crtConfig "github.com/codeready-toolchain/member-operator/pkg/configuration"
```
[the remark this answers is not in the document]
rajivnathan: 🤦  - I know this, but somehow I keep making this mistake...hopefully this is the last time

fixed in 45affeaf57ecb3b1211ca474e81310e24762050e

pkg/che/tokencache.go
```diff
@@ -0,0 +1,153 @@
...
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+
+	crtCfg "github.com/codeready-toolchain/member-operator/pkg/configuration"
```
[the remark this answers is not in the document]
rajivnathan: fixed in 45affeaf57ecb3b1211ca474e81310e24762050e

pkg/che/tokencache.go
rajivnathan: oops, I actually meant to use `Unlock` and `Lock` not `RUnlock` and `RLock`. Let me try something like https://github.com/codeready-toolchain/devcluster/blob/master/pkg/ibmcloud/cluster.go#L64. I did briefly try that but ran into some issue, but let me try again.

rajivnathan: Fixed in 24cf6c55f1de637ac561cbd681c6896adb87fa6a

pkg/che/che.go
```diff
@@ -0,0 +1,48 @@
...
+// InitDefaultCheClient initializes the default Che service instance
+func InitDefaultCheClient(cfg *crtConfig.Config, cl client.Client) {
+	DefaultClient = &Client{
+		config:     cfg,
+		k8sClient:  cl,
+		tokenCache: newTokenCache(),
+	}
+}
+
+func getRouteURL(cl client.Client, namespace, name string) (string, error) {
```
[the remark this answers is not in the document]
MatousJobanek: In addition, it looks like a copy of https://github.com/codeready-toolchain/member-operator/blob/master/pkg/controller/memberstatus/memberstatus_controller.go#L375-L387 - wouldn't it be good to export it somewhere and reuse it so we don't have to maintain it at two places?

pkg/che/tokencache.go
```diff
@@ -0,0 +1,156 @@
...
+		return TokenSet{}, err
+	}
+	log.Info("Che Keycloak Route", "URL", cheKeycloakURL)
+
+	// get che user token
+	if err = tc.obtainNewToken(cheKeycloakURL+tokenPath, user, pass); err != nil {
+		return TokenSet{}, err
+	}
+
+	return *tc.token, nil
```
MatousJobanek, line 69: I would rather move it to separated method so the read and write locks are more separated, but it's just my point of view ;-)

pkg/che/tokencache.go
```diff
@@ -0,0 +1,156 @@
...
+	err := json.Unmarshal([]byte(jsonString), &token)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error when unmarshal json with access token %s ", jsonString)
+	}
+	return &token, nil
+}
+
+// tokenExpired return false if the token is not nil and good for at least one more minute
+func tokenExpired(token *TokenSet) bool {
+	return token == nil || time.Now().After(time.Unix(token.Expiration-60, 0))
```
MatousJobanek, line 139: What is the default expiration time of the token?

pkg/rest/rest.go
```diff
@@ -0,0 +1,21 @@
...
+	buf := new(bytes.Buffer)
+	buf.ReadFrom(body)
+	return buf.String()
+}
+
+// CloseResponse reads the body and close the response. To be used to prevent file descriptor leaks.
+func CloseResponse(response *http.Response) {
+	ioutil.ReadAll(response.Body)
+	response.Body.Close()
+}
```
MatousJobanek, line 21: you don't handle the errors at all here

MatousJobanek (commented):
Good job :+1: 
Please, don't forget to change the Secret in toolchain-infra

pkg/che/che.go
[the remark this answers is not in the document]
rajivnathan: I'll be using it for che API calls in che.go as well but I agree it can be a common function somewhere. Any suggestions where? Maybe a `utils` package? eg. `github.com/codeready-toolchain/member-operator/pkg/utils` with a `route.go` file?

pkg/che/tokencache.go
rajivnathan: Good suggestion, let me try splitting it up

pkg/che/tokencache.go
rajivnathan: 5 minutes

pkg/rest/rest.go
rajivnathan: oh, I just copied it from https://github.com/codeready-toolchain/devcluster/blob/6defc6cce09762e3b65af412d20ba4970dd30456/pkg/rest/rest.go 🙂  and figured errors are ignored because it's mainly for cleanup

pkg/che/tokencache.go
rajivnathan: Done in 89b1e27e60e79ef25a5ba269afd473d55c6756eb

pkg/rest/rest.go
alexeykazakov: yeah. we don't really care about errors there. We could log them though. Or at least make the linters happy by using _

pkg/che/tokencache.go
```diff
@@ -0,0 +1,156 @@
...
+	if user == "" || pass == "" {
+		return TokenSet{}, fmt.Errorf("the che admin username and/or password are not configured")
+	}
+
+	// get keycloak URL
+	cheKeycloakURL, err := utils.GetRouteURL(cl, cfg.GetCheNamespace(), cfg.GetCheKeycloakRouteName())
+	if err != nil {
+		return TokenSet{}, err
+	}
+	log.Info("Che Keycloak Route", "URL", cheKeycloakURL)
```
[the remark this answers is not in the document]
rajivnathan: ah, that's a really good point. I've addressed it in a0a013d320dc6dadea441b5e3afe7d52e9d972d6

pkg/rest/rest.go
rajivnathan: Updated the `ReadBody` func to return the error and let the caller decide what to do.
Updated the `CloseResponse` func by using `_` to make the linters happy

alexeykazakov (approved):
Looks good!

pkg/che/tokencache.go
MatousJobanek: ok, then I would make the time cushion maybe a bit smaller - eg. 10seconds - or is there any reason why do we need to check that it will be valid at least one minute?

pkg/che/tokencache.go
```diff
@@ -0,0 +1,163 @@
...
+
+// tokenExpired return false if the token is not nil and good for at least one more minute
+func tokenExpired(token *TokenSet) bool {
+	return token == nil || time.Now().After(time.Unix(token.Expiration-60, 0))
+}
+
+// newHTTPClient returns a new HTTP client with some timeout and TLS values configured
+func newHTTPClient() *http.Client {
+	var netTransport = &http.Transport{
+		Dial: (&net.Dialer{
```
MatousJobanek, line 152: just a small detail - Dial is deprecated in http.Client - from the doc:
```
	// Deprecated: Use DialContext instead, which allows the transport
	// to cancel dials as soon as they are no longer needed.
	// If both are set, DialContext takes priority.
```

MatousJobanek (approved):
Looks good, thanks for addressing my comments :+1:

pkg/che/tokencache.go
rajivnathan: Thanks, I made a note of it. Will handle it in a following PR

PR #218, opened 2020-11-19
Implement more Che client functions

rajivnathan:
Implements functions:
1. Check if a given user exists
1. Get a user ID provided the username
1. Delete a user by user ID

Additional changes:
- moves both route.go and rest.go to the utils package

Related issue: https://issues.redhat.com/browse/CRT-847


pkg/che/che.go
```diff
@@ -23,7 +40,123 @@ type Client struct {
...
+	if err != nil {
+		return false, errors.Wrapf(err, "request to find Che user '%s' failed", username)
+	}
+	defer rest.CloseResponse(res)
+	if res.StatusCode == http.StatusOK {
+		return true, nil
+	} else if res.StatusCode == http.StatusNotFound {
+		return false, nil
+	}
+	return false, errors.Errorf("request to find Che user '%s' failed, Response status: '%s'", username, res.Status)
```
alexeykazakov, line 63: It would be useful if we also return the body of the response. It may contain the details of the error.

pkg/che/che.go
```diff
@@ -23,7 +40,123 @@ type Client struct {
...
+func (c *Client) GetUserIDByUsername(username string) (string, error) {
+	reqData := url.Values{}
+	reqData.Set("name", username)
+	res, err := c.cheRequest(http.MethodGet, cheUserFindPath, reqData)
+	if err != nil {
+		return "", errors.Wrapf(err, "unable to get Che user ID for user '%s'", username)
+	}
+	defer rest.CloseResponse(res)
+	if res.StatusCode != http.StatusOK {
+		resBody, _ := rest.ReadBody(res.Body)
```
alexeykazakov, line 76: Let's at least log the error from ReadBody()?

pkg/che/che.go
```diff
@@ -23,7 +40,123 @@ type Client struct {
...
+	if res.StatusCode == http.StatusOK {
+		return true, nil
+	} else if res.StatusCode == http.StatusNotFound {
+		return false, nil
+	}
+	return false, errors.Errorf("request to find Che user '%s' failed, Response status: '%s'", username, res.Status)
+}
+
+// GetUserIDByUsername returns the user ID that maps to the given username
+func (c *Client) GetUserIDByUsername(username string) (string, error) {
```
alexeykazakov, line 67: It will return an error if the user is not found, right? Let's add it to the function description.
Is the plan to use UserExists() first and if it does exist then use GetuserIDByUsername?

pkg/che/che.go
```diff
@@ -23,7 +40,123 @@ type Client struct {
...
+
+// DeleteUser deletes the Che user with the given user ID
+func (c *Client) DeleteUser(userID string) error {
+	log.Info("Deleteing user", "userID", userID)
+	res, err := c.cheRequest(http.MethodDelete, path.Join(cheUserPath, userID), nil)
+	if err != nil {
+		return errors.Wrapf(err, "unable to delete Che user with ID '%s'", userID)
+	}
+	defer rest.CloseResponse(res)
+	if res.StatusCode != http.StatusNoContent && res.StatusCode != http.StatusNotFound {
```
alexeykazakov, line 95: I would log an info message if the response is NotFound.

pkg/che/che.go
```diff
@@ -23,7 +40,123 @@ type Client struct {
...
+	}
+	cheUser, err := readCheUser(res)
+	if err != nil {
+		return "", errors.Wrapf(err, "unable to get Che user ID for user '%s'", username)
+	}
+	return cheUser.ID, err
+}
+
+// DeleteUser deletes the Che user with the given user ID
+func (c *Client) DeleteUser(userID string) error {
```
alexeykazakov, line 88: Maybe we should make all these functions internal/not exposed and later add just one exposed function: DeleteUser(username) which would call all the internal functions. So the client (controller) just need to call that DeleteUser(username) and leave everything else to the che client.

alexeykazakov (commented):
Looks good! And great test coverage! 
Just a few minor comments.

pkg/che/che.go
rajivnathan: Nice catch! I missed that one.

pkg/che/che.go
rajivnathan: Agreed, did the same for other areas

pkg/che/che.go
rajivnathan: Right, it will return an error. I've updated the comment to explain that.

The planned usage can be seen here: https://github.com/rajivnathan/member-operator/blob/64cf4a336a8dcf416bf8ce6eee937b52b9a85c26/pkg/controller/useraccount/useraccount_controller.go#L653

pkg/che/che.go
rajivnathan: Good idea! Done.

pkg/che/che.go
rajivnathan: That's actually what I did initially, but I discovered it doesn't seem to be that simple because `UserExists` and `GetUserIDByUsername` can return 404 while the deletion is still in progress and since we need to get the UserID to do a retry of `DeleteUser` until it returns a 204 normally it was getting a bit messy to have everything handled by a `DeleteUser` function in che.go. This is the basic implementation I have (also what I used for the demo): https://github.com/rajivnathan/member-operator/blob/64cf4a336a8dcf416bf8ce6eee937b52b9a85c26/pkg/controller/useraccount/useraccount_controller.go#L653

It can probably be made a bit cleaner though, so I'm open to suggestions for improving that.

alexeykazakov: Hm.. OK. Che keeps complicating our stuff :)
Let'd keep it as is then.

alexeykazakov (approved):
Thanks!

pkg/che/che.go
```diff
@@ -23,7 +40,135 @@ type Client struct {
...
+	}
+	defer rest.CloseResponse(res)
+	if res.StatusCode != http.StatusNoContent && res.StatusCode != http.StatusNotFound {
+		resBody, readError := rest.ReadBody(res.Body)
+		if readError != nil {
+			log.Error(readError, "error while reading body of the delete Che user response")
+		}
+		err = errors.Errorf("unable to delete Che user with ID '%s', Response status: '%s' Body: '%s'", userID, res.Status, resBody)
+	} else if res.StatusCode == http.StatusNotFound {
+		log.Info("The user was not deleted because they were not found", "userID", userID)
```
MatousJobanek, line 109: A tiny typo - you can ignore it :-) 
```suggestion
		log.Info("The user was not deleted because it wasn't found", "userID", userID)
```

pkg/che/che.go
```diff
@@ -23,7 +40,135 @@ type Client struct {
...
+
+func (c *Client) cheRequest(rmethod, endpoint string, queryParams url.Values) (*http.Response, error) {
+	// get Che route URL
+	cheURL, err := route.GetRouteURL(c.k8sClient, c.config.GetCheNamespace(), c.config.GetCheRouteName())
+	if err != nil {
+		return nil, err
+	}
+
+	// create request
+	req, err := http.NewRequest(rmethod, cheURL+endpoint, nil)
```
MatousJobanek, line 122: shouldn't the param be just `method`? It's just a detail - you can ignore it as well ;-) 
```suggestion
func (c *Client) cheRequest(method, endpoint string, queryParams url.Values) (*http.Response, error) {
	// get Che route URL
	cheURL, err := route.GetRouteURL(c.k8sClient, c.config.GetCheNamespace(), c.config.GetCheRouteName())
	if err != nil {
		return nil, err
	}

	// create request
	req, err := http.NewRequest(method, cheURL+endpoint, nil)
```

MatousJobanek (approved):
Great job :1st_place_medal: 
Good solution with dealing with the complications that we are facing with Che :+1: 
And an amazing job with the tests :+1: :rocket:

pkg/che/che.go
rajivnathan: It can be, I have no problem changing it 🙂

rajivnathan:
Thanks for the reviews and the kind comments!

PR #221, opened 2020-11-26
Update useraccount controller to delete Che user

rajivnathan:
Adds logic to handle deleting the Che user and all his/her data as part of Sandbox user deactivation/removal.

Related issue: https://issues.redhat.com/browse/CRT-848


pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -608,3 +644,59 @@ func newNSTemplateSet(userAcc *toolchainv1alpha1.UserAccount) *toolchainv1alpha1
 func ToIdentityName(userID string, identityProvider IdentityProvider) string {
 	return fmt.Sprintf("%s:%s", identityProvider.GetIdP(), userID)
 }
+
+func (r *ReconcileUserAccount) lookupAndDeleteCheUser(userAcc *toolchainv1alpha1.UserAccount) error {
+
+	// If the admin username is not set then Che user deletion is not configured, just return
+	if r.config.GetCheAdminUsername() == "" {
```
xcoulon, line 651: do we have a username only, or a password/token as well?

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -608,3 +644,59 @@ func newNSTemplateSet(userAcc *toolchainv1alpha1.UserAccount) *toolchainv1alpha1
...
+		log.Info("Che user deletion is not configured, configure the Che admin credentials to enable it")
+		return nil
+	}
+
+	var userExists bool
```
xcoulon, line 656: this line is not needed, AFAICT

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -608,3 +644,59 @@ func newNSTemplateSet(userAcc *toolchainv1alpha1.UserAccount) *toolchainv1alpha1
...
+	// If the user ID is not in the status then look it up and set it in the status
+	if cheUserID == "" {
+		cheUserID, err = r.cheClient.GetUserIDByUsername(userAcc.Name)
+		if err != nil {
+			return err
+		}
+
+		// Set the Che user deletion status (store the Che user ID in the status for subsequent retries if needed)
+		// This is required because the deletion API will return an error until the user is successfully removed.
+		if !condition.HasConditionReason(userAcc.Status.Conditions, toolchainv1alpha1.UserAccountCheCleanup, toolchainv1alpha1.UserAccountDeletingCheDataReason) {
```
xcoulon, line 688: why do you need the check the condition/reason again here?

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -749,6 +777,65 @@ func TestReconcile(t *testing.T) {
...
+		// then
+		userAcc = &toolchainv1alpha1.UserAccount{}
+		err = r.client.Get(context.TODO(), types.NamespacedName{Name: username, Namespace: test.MemberOperatorNs}, userAcc)
+		require.NoError(t, err)
+
+		// Check that the finalizer is present
+		require.True(t, util.HasFinalizer(userAcc, toolchainv1alpha1.FinalizerName))
+
+		// Set the deletionTimestamp
+		userAcc.DeletionTimestamp = &metav1.Time{time.Now()} //nolint: govet
```
xcoulon, line 816: maybe you can use `metav1.Now()` to avoid having to comment with `//nolint: govet`?

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -749,6 +777,65 @@ func TestReconcile(t *testing.T) {
...
+			test.Env("MEMBER_OPERATOR_SECRET_NAME", "test-secret"),
+			// Che has two different routes for Che & Keycloak but CRW uses a single route for both.
+			// For tests we'll assume they're two different routes.
+			test.Env(configuration.MemberEnvPrefix+"_CHE_KEYCLOAK_ROUTE_NAME", "keycloak"),
+		)
+		defer restore()
+
+		defer gock.OffAll()
+		gockTokenSuccess()
+		gockFindUserNoBody("johnsmith", 400)
```
xcoulon, line 796: so the deletion should fail because Che returns a `400` response when our controller sends a request to get the user info?

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -1133,6 +1220,164 @@ func TestDisabledUserAccount(t *testing.T) {
 	})
 }
 
+func TestLookupAndDeleteCheUser(t *testing.T) {
```
xcoulon, line 1223: Isn't there some overlap between the `TestLookupAndDeleteCheUser` tests here and the `TestReconcile` tests above?

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -608,3 +644,59 @@ func newNSTemplateSet(userAcc *toolchainv1alpha1.UserAccount) *toolchainv1alpha1
...
+			if err := r.setStatusCheUserDeletionInProgress(userAcc, cheUserID); err != nil {
+				return err
+			}
+		}
+	}
+
+	// Delete the Che user. It is common for this call to return an error multiple times before succeeding
+	if err := r.cheClient.DeleteUser(cheUserID); err != nil {
+		return err
```
alexeykazakov, line 697: It's worth wrapping this error with an explanation that that error is kinda normal, so we don't get scared if see it in the logs and in the ua status.

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -597,8 +608,25 @@ func TestReconcile(t *testing.T) {
 	// Delete useraccount and ensure related resources are also removed
 	t.Run("delete useraccount removes subsequent resources", func(t *testing.T) {
 		// given
+
+		// when the member operator secret exists and has a che admin user configured then che user deletion is enabled
+		restore := test.SetEnvVarsAndRestore(t,
+			test.Env("WATCH_NAMESPACE", "toolchain-member"),
```
MatousJobanek, line 614: it's just a detail, but I would use the same namespace as is used in the rest of the test: `test.MemberOperatorNs`
The same applies for the other tests

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -597,8 +608,25 @@ func TestReconcile(t *testing.T) {
...
+			// Che has two different routes for Che & Keycloak but CRW uses a single route for both.
+			// For tests we'll assume they're two different routes.
+			test.Env(configuration.MemberEnvPrefix+"_CHE_KEYCLOAK_ROUTE_NAME", "keycloak"),
+		)
+		defer restore()
+
+		defer gock.OffAll()
+		gockTokenSuccess()
+		gockFindUserSuccess("johnsmith")
+		gockDeleteUser(204)
```
MatousJobanek, line 625: could we check that it was realy called? and the number of calls...?
for that you could add a function:
```
func SpyOnCalls(counter *int) gock.Matcher {
	matcher := gock.NewBasicMatcher()
	matcher.Add(func(_ *http.Request, _ *gock.Request) (bool, error) {
		*counter++
		return true, nil
	})
	return matcher
}
```
and use it:
```
func gockDeleteUser(code int, calls *int) {
	gock.New(testCheURL).
		Delete("api/user").
		SetMatcher(SpyOnCalls(calls)).
		MatchHeader("Authorization", "Bearer abc.123.xyz").
		Persist().
		Reply(code)
}
```
you could do the same also for the other gock mocks - since you use the `Persist()` method there and not `Times()`.

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -1133,6 +1220,164 @@ func TestDisabledUserAccount(t *testing.T) {
...
+			r, _, _ := prepareReconcile(t, username, userAcc, memberOperatorSecret, cheRoute(true), keycloackRoute(true))
+
+			//when
+			err := r.lookupAndDeleteCheUser(userAcc)
+
+			// then
+			require.EqualError(t, err, `request to find Che user 'sugar' failed: unable to obtain access token for che, Response status: '400 Bad Request'. Response body: ''`)
+			userAcc = &toolchainv1alpha1.UserAccount{}
+			err = r.client.Get(context.TODO(), types.NamespacedName{Name: username, Namespace: test.MemberOperatorNs}, userAcc)
+			require.NoError(t, err)
```
MatousJobanek, line 1267: I'm slightly missing the purpose of these retrievals - you just get the UserAccount, but don't check anything there - shouldn't you verify the non-existence of the condition? If not, then you can remove this part

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -1133,6 +1220,164 @@ func TestDisabledUserAccount(t *testing.T) {
...
+			test.Env(configuration.MemberEnvPrefix+"_CHE_KEYCLOAK_ROUTE_NAME", "keycloak"),
+		)
+		defer restore()
+
+		t.Run("get token error", func(t *testing.T) {
+			// given
+			defer gock.OffAll()
+			gockTokenFail()
+			gockFindUserSuccess(username)
+			gockDeleteUser(204)
```
MatousJobanek, line 1256: if the token retrieval fails, then you don't need these mocks
```suggestion
```

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -1133,6 +1220,164 @@ func TestDisabledUserAccount(t *testing.T) {
...
+			err = r.client.Get(context.TODO(), types.NamespacedName{Name: username, Namespace: test.MemberOperatorNs}, userAcc)
+			require.NoError(t, err)
+		})
+
+		t.Run("user not found", func(t *testing.T) {
+			// given
+			defer gock.OffAll()
+			gockTokenSuccess()
+			gockFindUserNoBody(username, 404)
+			gockDeleteUser(204)
```
MatousJobanek, line 1275: if the user wasn't found, then you don't need this mock
```suggestion
```

pkg/controller/useraccount/useraccount_controller_test.go
```diff
@@ -1133,6 +1220,164 @@ func TestDisabledUserAccount(t *testing.T) {
...
+			err = r.client.Get(context.TODO(), types.NamespacedName{Name: username, Namespace: test.MemberOperatorNs}, userAcc)
+			require.NoError(t, err)
+		})
+
+		t.Run("find user error", func(t *testing.T) {
+			// given
+			defer gock.OffAll()
+			gockTokenSuccess()
+			gockFindUserNoBody(username, 400)
+			gockDeleteUser(204)
```
MatousJobanek, line 1294: Similarly here
```suggestion
```

pkg/controller/useraccount/useraccount_controller_test.go
MatousJobanek: or you could return an error:
```suggestion
			gockDeleteUser(404)
```

pkg/controller/useraccount/useraccount_controller.go
rajivnathan: You're right, good catch! I did a bit of refactoring so must have left it out. Done

pkg/controller/useraccount/useraccount_controller.go
rajivnathan: Good point 🙂  Done

pkg/controller/useraccount/useraccount_controller.go
rajivnathan: There is a password as well, but the main purpose of this line is just to have a way to disable this Che user deletion stuff if we want/need to. Instead of creating a new flag for that I figured just whether the username is configured or not in the member operator secret could serve as a gate for that purpose.

pkg/controller/useraccount/useraccount_controller_test.go
rajivnathan: It seems it's not that straightforward, `userAcc.DeletionTimestamp` expects a pointer and metav1.Now() just returns a value.

pkg/controller/useraccount/useraccount_controller_test.go
rajivnathan: That's correct, I updated the test description and added a comment to make it a bit more clear. Thanks for the question!

pkg/controller/useraccount/useraccount_controller_test.go
rajivnathan: I think only a little bit, so figured it's not harmful. 🙂

pkg/controller/useraccount/useraccount_controller.go
rajivnathan: Good point! Done

pkg/che/che.go
```diff
@@ -33,16 +33,27 @@ type Client struct {
...
 func InitDefaultCheClient(cfg *crtcfg.Config, cl client.Client) {
+	defaultHTTPClient := newHTTPClient()
 	DefaultClient = &Client{
 		config:     cfg,
-		httpClient: newHTTPClient(),
+		httpClient: defaultHTTPClient,
 		k8sClient:  cl,
-		tokenCache: newTokenCache(),
+		tokenCache: NewTokenCache(defaultHTTPClient),
+	}
```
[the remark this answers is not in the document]
rajivnathan: Good point, would ensure if we add anything to the `NewCheClient` func then tests and prod wouldn't stray apart.

pkg/controller/useraccount/useraccount_controller.go
```diff
@@ -556,6 +564,34 @@ func (r *ReconcileUserAccount) setStatusTerminating(userAcc *toolchainv1alpha1.U
...
+			conditions = append(conditions[:i], conditions[i+1:]...)
+			updated = true
+		}
+	}
+	if !updated {
+		// Nothing changed
+		return nil
+	}
+	userAcc.Status.Conditions = conditions
+	return r.client.Status().Update(context.TODO(), userAcc)
```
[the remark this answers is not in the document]
rajivnathan: Yes! It does the same thing but without the `updated` bool, so I'm all for it. Thanks!

pkg/controller/useraccount/useraccount_controller_test.go
rajivnathan: 🤦  - yes, I'm supposed to check the conditions. Fixed

pkg/controller/useraccount/useraccount_controller_test.go
rajivnathan: That's a good idea! I've done that and added assertions for the counter. Thanks for the suggestion! 🥇

pkg/controller/useraccount/useraccount_controller_test.go
rajivnathan: True. Fixed.

pkg/controller/useraccount/useraccount_controller_test.go
rajivnathan: Done

pkg/controller/useraccount/useraccount_controller_test.go
rajivnathan: Done

pkg/controller/useraccount/useraccount_controller_test.go
rajivnathan: Done

MatousJobanek (approved):
Nice :+1: 
thanks a lot for addressing my comments :pray:

pkg/controller/useraccount/useraccount_controller.go
xcoulon: ok, I see. Thanks for the explanation

pkg/controller/useraccount/useraccount_controller_test.go
xcoulon: yeah, indeed... ok then, let's keep it like that if you prefer

pkg/controller/useraccount/useraccount_controller_test.go
xcoulon: ok, thanks for your answer

xcoulon: actually, I thought you mention a `500` error when the user's workspaces are still running, so I was a bit surprised to see a `400` response here, but I probably misunderstood the response code when we discussed about it last time (or I just assumed it was a `500` error 😜)

pkg/controller/useraccount/useraccount_controller_test.go
xcoulon: I would have preferred all tests in the same `Test..` func, but that's because I thought the overlap was bigger. My point is, if we need to add tests, we'll have to figure out what's the best location for them.

pkg/controller/useraccount/useraccount_controller.go
xcoulon: yes, something like `Keep calm` or `Stay cool` 😂

xcoulon:
thanks for addressing my comments, @rajivnathan! 🙌

pkg/controller/useraccount/useraccount_controller_test.go
rajivnathan: Oh, I think you're referring to the 500 error from the DELETE API. This test is assuming the Find user API fails with an error. Don't think it matters what error as long as it's not 404 which would trigger a different code path than desired.

xcoulon: yes, I was probably mixing stuff. But as long as both/all cases are covered, we're fine!

PR #222, opened 2020-11-27
feat: introduce webhook template & add permissions & configuration

MatousJobanek:
This PR adds:
* mutating admission webhook template
* generation of the template into an asset
* necessary permissions for deploying objects form the template
* configuration to load the member-operator-webhook image location URL

This PR doesn't introduce any real logic yet - deployment of the template as well as the logic of the webhook will come in the next PR.

### DIFF
--- a/deploy/webhook/member-operator-webhook.yaml
+++ b/deploy/webhook/member-operator-webhook.yaml
@@ -53,10 +53,10 @@ objects:
           imagePullPolicy: IfNotPresent
           resources:
             limits:
-              cpu: 500m
+              cpu: 300m
               memory: 500Mi
             requests:
-              cpu: 250m
+              cpu: 75m
               memory: 128Mi
           volumeMounts:
           - name: webhook-certs
--- a/deploy/webhook/member-operator-webhook.yaml
+++ b/deploy/webhook/member-operator-webhook.yaml
@@ -1,0 +1,107 @@
+kind: Template
+apiVersion: template.openshift.io/v1
+metadata:
+  name: member-operator-webhook
+objects:
+- apiVersion: scheduling.k8s.io/v1
+  kind: PriorityClass
+  metadata:
+    name: sandbox-users-pods
+    labels:
+      toolchain.dev.openshift.com/provider: codeready-toolchain
+  value: -10
+  globalDefault: false
+  description: "Priority class for pods in users' namespaces"
+- apiVersion: v1
+  kind: Service
+  metadata:
+    name: member-operator-webhook
+    namespace: ${NAMESPACE}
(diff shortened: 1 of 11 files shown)


deploy/webhook/member-operator-webhook.yaml
```diff
@@ -0,0 +1,107 @@
...
+          image: ${IMAGE}
+          command:
+          - member-operator-webhook
+          imagePullPolicy: IfNotPresent
+          resources:
+            limits:
+              cpu: 500m
+              memory: 500Mi
+            requests:
+              cpu: 250m
```
alexeykazakov, line 59: Isn't it too much for the cpu request? We can tweak it later though.

deploy/webhook/member-operator-webhook.yaml
```diff
@@ -0,0 +1,107 @@
...
+    reinvocationPolicy: Never
+    failurePolicy: Ignore
+    namespaceSelector:
+      matchLabels:
+        toolchain.dev.openshift.com/provider: codeready-toolchain
+parameters:
+- name: NAMESPACE
+  value: 'toolchain-member-operator'
+- name: IMAGE
+  value: quay.io/matousjobanek/member-operator-webhook:latest
```
alexeykazakov, line 105: Is it just temporary? We should already have an image in the codeready org, shouldn't we?

alexeykazakov (approved):
Looks good! Just one question regarding the repo org.

deploy/webhook/member-operator-webhook.yaml
MatousJobanek: Yeah, probably. I didn't spend much time with these values TBH :sweat_smile: :man_shrugging: 
I could have set something smaller - in an example that I saw there was:
```
            limits:
              cpu: 500m
              memory: 128Mi
            requests:
              cpu: 250m
              memory: 64Mi
```
so, I just blindly increased both vales, but it's true that I could keep the CPU smaller

deploy/webhook/member-operator-webhook.yaml
MatousJobanek: :ghost: zombie - let me fix that
and no - we don't have the image in the quay repo yet - I need to modify the CD script first :-)  That's the next step, but should be pretty easy

MatousJobanek: fixed: 0c3135c

deploy/webhook/member-operator-webhook.yaml
MatousJobanek: decreased here 7b5a7ff

PR #223, opened 2020-11-27
Users pods admission webhook

MatousJobanek:
builds on top of https://github.com/codeready-toolchain/member-operator/pull/222 so the only commit to be reviewed is: 8b81fd3

During the operator start it deploys all objects/resources from the webhook template. If they already exist, then it updates them if they are different.

The actual admission webhook sets the priorityClassName as well as the actual priority to all pods that are created in users' namespaces.

Paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/204


cmd/manager/main.go
```diff
@@ -164,6 +165,14 @@ func main() {
 			log.Error(fmt.Errorf("timed out waiting for main cache to sync"), "")
 			os.Exit(1)
 		}
+
+		log.Info("(Red)Deploying users' pods webhook")
```
alexeykazakov, line 169: (Re~~d~~)

cmd/manager/main.go
```diff
@@ -164,6 +165,14 @@ func main() {
...
+		if err := deploy.Webhook(mgr.GetClient(), mgr.GetScheme(), namespace, crtConfig.GetMemberOperatorWebhookImage()); err != nil {
+			log.Error(err, "cannot deploy mutating users' pods webhook")
+			os.Exit(1)
+		}
+		log.Info("(Red)Deployed users' pods webhook")
```
alexeykazakov, line 174: (Re~~d~~)

pkg/webhook/deploy/deployment.go
```diff
@@ -0,0 +1,82 @@
...
+		"IMAGE":     image,
+	})
+}
+
+func ensureCertSecret(cl client.Client, namespace string) ([]byte, error) {
+	certSecret := &v1.Secret{}
+	if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: certSecretName}, certSecret); err != nil && !errors.IsNotFound(err) {
+		return nil, err
+	} else if err != nil {
+		certSecret, err = cert.CreateSecret(certSecretName, namespace, "member-operator-webhook")
```
alexeykazakov, line 73: Shouldn't we update the secret if the cert has expired? If I'm not mistaken the cert will be expired in one year, right?

pkg/webhook/mutatingwebhook/mutate.go
```diff
@@ -0,0 +1,132 @@
...
+		respBody = mutate(body)
+		w.WriteHeader(http.StatusOK)
+	}
+	if _, err := w.Write(respBody); err != nil {
+		log.Error(err, "unable to write response")
+	}
+}
+
+func mutate(body []byte) []byte {
+	log.Info("received", "body", string(body))
```
alexeykazakov, line 79: Wouldn't it be too verbos to log the entire request body?

pkg/webhook/mutatingwebhook/mutate.go
```diff
@@ -0,0 +1,132 @@
...
+		err := fmt.Errorf("admission review request is nil")
+		log.Error(err, "cannot read the admission review request", "AdmissionReview", admReview)
+		return responseWithError(err)
+	}
+
+	// let's unmarshal the object to be sure that it's a pod
+	var pod *corev1.Pod
+	if err := json.Unmarshal(admReview.Request.Object.Raw, &pod); err != nil {
+		log.Error(err, "unable unmarshal pod json object", "AdmissionReview", admReview)
+		return responseWithError(errors.Wrapf(err, "unable unmarshal pod json object"))
```
alexeykazakov, line 114: Let's maybe add the request body to the error message?

alexeykazakov (commented):
Looks cool! :)

cmd/manager/main.go
MatousJobanek: thanks :smile:  7fca9e4

cmd/manager/main.go
MatousJobanek: 7fca9e4

pkg/webhook/mutatingwebhook/mutate.go
MatousJobanek: yeah, probably - remved in 164bb62

pkg/webhook/mutatingwebhook/mutate.go
MatousJobanek: good point added in 164bb62

pkg/webhook/deploy/deployment.go
MatousJobanek: you are right - I knew that I forgot something, but didn't realize it :man_facepalming:

MatousJobanek: ok, I slightly refactored the secret creation and added also the check of the expiration: efd00dc

pkg/webhook/deploy/cert/secret.go
```diff
@@ -31,11 +23,62 @@ const (
...
+	}
+
+	// already exists - check the expiration date of the certificate to see if it needs to be updated
+	cert, err := tls.X509KeyPair(certSecret.Data[ServerCert], certSecret.Data[ServerKey])
+	if err != nil {
+		log.Error(err, "creating pem from certificate and key failed")
+	} else {
+		certData, err := x509.ParseCertificate(cert.Certificate[0])
+		if err != nil {
+			log.Error(err, "parsing certificate failed")
```
alexeykazakov, line 60:
```suggestion
			log.Error(err, "parsing certificate failed; will update the secret with a freshly generated cert")
```

pkg/webhook/deploy/cert/secret.go
```diff
@@ -31,11 +23,62 @@ const (
...
+		} else if time.Now().Add(oneWeek).Before(certData.NotAfter) {
```
alexeykazakov, line 61: We check/create the certs only on the webhook/operator startup, correct? It means that if there is no updates/re-starts of the operator within the last week of expiration timeout then we will get invalid certs used.

If we still want to use that timeout let's use some timeout which is a half of the expiration timeout.
For example it the TTL is 12 months than the timeout is 6 month. It would be also nice to have both of these const declared near each other so we see the relation. Or we could just dynamically calculate the timeout based on the expiration const.

Later we could introduce some controller which would re-queue every N days to ensure the certs are up to date.

MatousJobanek: > We check/create the certs only on the webhook/operator startup, correct? It means that if there is no updates/re-starts of the operator within the last week of expiration timeout then we will get invalid certs used.

yes, you are right. I realized this as well, but I thought that later I would most likely add a controller (as you also mentioned in the comment) that would watch all objects of the template - in the same way as we do with reg-service. With the controller and postponed reconcile loops we should be able to ensure that there is at least one re-queue within a day (or a few days, as you mentioned as well)

>If we still want to use that timeout let's use some timeout which is a half of the expiration timeout.
For example it the TTL is 12 months than the timeout is 6 month. 

That's very good point, having the timeout way longer will easily solve this potential issue. No idea why I didn't realize it :man_shrugging:  let me do it :wink:

MatousJobanek: done here 54fd990
the secrets are replaced after half of the expiration timeout

alexeykazakov (approved):

PR #225, opened 2020-12-04
Drop operator limits

alexeykazakov:
There was a long thread with OpenShift folks regarding setting limits in OpenShift payloads. There were recommendations do not set them at all for most cases. Some of the reasons are defined there: https://github.com/openshift/enhancements/blob/master/CONVENTIONS.md#resources-and-limits


rajivnathan (approved):
👍

PR #228, opened 2020-12-10
Proceed with removal if user not found

rajivnathan:
Fix for bug https://issues.redhat.com/browse/CRT-891

Update the Che user deletion logic such that if the UserExists check comes back with a 404 then we won't continue further to call the DELETE API, we'll assume the Che data has been removed and continue with deactivation/removal.

It also removes the logic to remove the setting of the special condition that was used for caching the cheUserID value between attempts. We no longer need to cache the cheUserID because if we can't get it due to a 404 then it now means the user is removed, if cannot get it due to another response code then it will be an error condition set on the useraccount.

### DIFF
--- a/pkg/controller/useraccount/useraccount_controller.go
+++ b/pkg/controller/useraccount/useraccount_controller.go
@@ -651,34 +651,16 @@ func (r *ReconcileUserAccount) lookupAndDeleteCheUser(userAcc *toolchainv1alpha1
 		return err
 	}
 
-	// If the user doesn't exist and there's no deletion in progress then there's nothing left to do.
-	// We need to check whether the deletion is in progress because in some cases the user could be reported as non-existent while the deletion is still in progress.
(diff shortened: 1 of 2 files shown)


alexeykazakov (approved):
Nice cleanup!

Let's delete not used ConditionType and Reasons from API **after** we merge this PR and confirm everything works as expected.

MatousJobanek (approved):
Nice :+1: :1st_place_medal: 
in the next PR, we could make the deletion of the Che data as the last action of the finalizer. With this, it will be easier to unblock any potential problematic deletions by manual removing the finalized from the resource.

rajivnathan:
> Nice cleanup!
> 
> Let's delete not used ConditionType and Reasons from API **after** we merge this PR and confirm everything works as expected.

PR up here: https://github.com/codeready-toolchain/api/pull/208

PR #231, opened 2020-12-18
Che client status

rajivnathan:
🚨  **IMPORTANT:** 🚨 
The user deletion enabled config variable must be set to true on the member operator config maps on staging and production *before* merging this PR:

```
che:
  user_deletion_enabled: true
```

**Description:**
Adds a new section to the memberstatus specifically for any Che integrations.

**Detects the following problem conditions and exposes them as part of the member status:**
- che admin credentials not configured when Che user deletion is enabled
- che route not found
- unable to successfully access Che user API

**Additional changes:**
- adds a new config variable specifically for enabling/disabling the Che user deletion logic

**Example status:**
```
status:
  che:
    conditions:
    - lastTransitionTime: "2020-12-18T00:35:55Z"
      lastUpdatedTime: "2020-12-18T00:35:55Z"
      message: 'che user API check failed: unable to obtain access token for che,
        Response status: ''401 Unauthorized''. Response body: ''{"error":"invalid_grant","error_description":"Invalid
        user credentials"}'''
      reason: CheUserAPICheckFailed
      status: "False"
      type: Ready
  conditions:
  - lastTransitionTime: "2020-12-18T00:35:55Z"
    lastUpdatedTime: "2020-12-18T00:35:55Z"
    message: 'components not ready: [cheIntegration]'
    reason: ComponentsNotReady
    status: "False"
    type: Ready
  host:
    conditions:
    - lastTransitionTime: "2020-12-18T00:35:55Z"
      lastUpdatedTime: "2020-12-18T00:35:55Z"
      reason: HostConnectionReady
      status: "True"
      type: Ready
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -38,6 +38,6 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6 // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
 
-replace github.com/codeready-toolchain/api => /Users/rajiv/go/src/github.com/codeready-toolchain/api
+replace github.com/codeready-toolchain/api => github.com/rajivnathan/api v0.0.0-20201218021903-cf2c667c66b9
 
 go 1.14
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/codeready-toolchain/member-operator
 
 require (
-	github.com/codeready-toolchain/api v0.0.0-20201126155707-1094c3e32973
+	github.com/codeready-toolchain/api v0.0.0-20210106162323-f9e542b70ee3
 	github.com/codeready-toolchain/toolchain-common v0.0.0-20201014231429-a44cccb4b2b5
 	github.com/go-logr/logr v0.1.0
 	github.com/gofrs/uuid v3.3.0+incompatible
@@ -37,7 +37,5 @@ replace (
 	k8s.io/client-go => k8s.io/client-go v0.18.3 // Required by prometheus-operator
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6 // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
-
-replace github.com/codeready-toolchain/api => github.com/rajivnathan/api v0.0.0-20201218021903-cf2c667c66b9
 
 go 1.14
--- a/go.mod
+++ b/go.mod
@@ -33,4 +33,6 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6 // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
 
+replace github.com/codeready-toolchain/api => /Users/rajiv/go/src/github.com/codeready-toolchain/api
+
 go 1.14
--- a/pkg/che/che.go
+++ b/pkg/che/che.go
@@ -166,6 +166,12 @@ func (c *Client) cheRequest(method, endpoint string, queryParams url.Values) (*h
 	return c.httpClient.Do(req)
 }
 
+// IsCheIntegrationEnabled if the Che required configuration is set to 'true' and the Che admin username is set
+// then the member operator's che integration is enabled. Returns false otherwise
+func (c *Client) IsCheIntegrationEnabled() bool {
+	return c.config.IsCheRequired() && c.config.GetCheAdminUsername() != ""
+}
+
 // User holds the user data retrieved from the Che user API
 type User struct {
 	Name string `json:"name"`
--- a/pkg/che/che.go
+++ b/pkg/che/che.go
@@ -166,10 +166,10 @@ func (c *Client) cheRequest(method, endpoint string, queryParams url.Values) (*h
 	return c.httpClient.Do(req)
 }
 
-// IsCheIntegrationEnabled if the Che required configuration is set to 'true' and the Che admin username is set
-// then the member operator's che integration is enabled. Returns false otherwise
+// IsCheIntegrationEnabled returns true if and only if IsCheRequired is 'true' and the Che admin username
+// and password are both set and not empty. Returns false otherwise
 func (c *Client) IsCheIntegrationEnabled() bool {
-	return c.config.IsCheRequired() && c.config.GetCheAdminUsername() != ""
+	return c.config.IsCheRequired() && c.config.GetCheAdminUsername() != "" && c.config.GetCheAdminPassword() != ""
 }
 
 // User holds the user data retrieved from the Che user API
--- a/pkg/che/che.go
+++ b/pkg/che/che.go
@@ -166,12 +166,6 @@ func (c *Client) cheRequest(method, endpoint string, queryParams url.Values) (*h
 	return c.httpClient.Do(req)
 }
 
-// IsCheIntegrationEnabled returns true if and only if IsCheRequired is 'true' and the Che admin username
-// and password are both set and not empty. Returns false otherwise
-func (c *Client) IsCheIntegrationEnabled() bool {
-	return c.config.IsCheRequired() && c.config.GetCheAdminUsername() != "" && c.config.GetCheAdminPassword() != ""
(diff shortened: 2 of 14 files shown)


pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -41,6 +42,7 @@ const (
 	hostConnectionTag statusComponentTag = "hostConnection"
 	resourceUsageTag  statusComponentTag = "resourceUsage"
 	routesTag         statusComponentTag = "routes"
+	cheIntegrationTag statusComponentTag = "cheIntegration"
```
alexeykazakov, line 45: Shouldn't it be `che` for the sake of consistency? When the component tag matches the component name?

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -291,6 +296,47 @@ func (r *ReconcileMemberStatus) routesHandleStatus(reqLogger logr.Logger, member
...
+func (r *ReconcileMemberStatus) cheIntegrationHandleStatus(reqLogger logr.Logger, memberStatus *toolchainv1alpha1.MemberStatus) error {
+	if memberStatus.Status.Che == nil {
+		memberStatus.Status.Che = &toolchainv1alpha1.CheStatus{}
+	}
+
+	// Is che user deletion enabled
+	if !r.config.IsCheUserDeletionEnabled() {
+		// Che user deletion is not enabled, set condition to Ready. No further checks required
+		readyCondition := status.NewComponentReadyCondition(toolchainv1alpha1.ToolchainStatusMemberStatusCheUserDeletionNotEnabledReason)
+		memberStatus.Status.Che.Conditions = []toolchainv1alpha1.Condition{*readyCondition}
```
alexeykazakov, line 310: Technically we are doing it wrong. The condition has `LastTransitionTime` which we always update even if there is no change in the condition status. We should update the existing condition (if already exists) instead of replacing it by a new one every time.
But we are doing it wrong for all the other conditions (but the main/overall Ready condition) for all the other components too. It's not critical right now and doesn't break anything though.
So, let's keep it as is and address in a separate PR (let's create a JIRA for that so we can prioritize it properly later).

pkg/controller/memberstatus/memberstatus_controller.go
```diff
@@ -291,6 +296,47 @@ func (r *ReconcileMemberStatus) routesHandleStatus(reqLogger logr.Logger, member
...
+		return nil
+	}
+
+	if r.config.GetCheAdminUsername() == "" || r.config.GetCheAdminPassword() == "" {
```
alexeykazakov, line 314: You intended to use `isCheAdminUserConfigured()` instead, right?

pkg/configuration/configuration.go
```diff
@@ -54,6 +54,12 @@ const (
 	// defaultCheRequired is the default value for che.required param
 	defaultCheRequired = false
 
+	// varCheUserDeletionEnabled is a way to turn the Che user deletion logic on/off
+	varCheUserDeletionEnabled = "che.user.deletion.enabled"
```
alexeykazakov, line 58: I missed when was doing review for the che configuration previously but due to viper specifics in other places we usually use `.` to separate parent properties from child properties. And `_` when we need to define a multi word property name.
So, if we load the configuration form yaml, json, property file into viper it has opportunity to distinguish parent/children properties.
For example:
```suggestion
	varCheUserDeletionEnabled = "che.user_deletion_enabled"
```
could be loaded as
```
che:
  user_deletion_enabled: true
```
vs `varCheUserDeletionEnabled = "che.user.deletion.enabled"` as:
```
che:
  user:
    deletion:
      enabled: true
```
But not sure if it's worth changing it now since the other configuration properties here do not follow that. So, I'm OK to keep it as is for the consistency sake.

pkg/configuration/configuration.go
```diff
@@ -54,6 +54,12 @@ const (
...
+
+	// defaultCheUserDeletionEnabled is the default value for che.required param
+	defaultCheUserDeletionEnabled = false
```
alexeykazakov, line 61: It means we have to set this property to `true` in all our clusters first before we can merge this PR, right?

pkg/che/che.go
```diff
@@ -120,6 +120,24 @@ func (c *Client) DeleteUser(userID string) error {
...
+	if err != nil {
+		return errors.Wrapf(err, "che user API check failed")
+	}
+	defer rest.CloseResponse(res)
+	if res.StatusCode == http.StatusOK {
+		return nil
+	}
+	resBody, readError := rest.ReadBody(res.Body)
+	if readError != nil {
+		log.Error(readError, "error while reading body of the che user API check response")
```
alexeykazakov, line 136: Shouldn't we return the error message too in this case instead just logging it?

pkg/che/che.go
```diff
@@ -120,6 +120,24 @@ func (c *Client) DeleteUser(userID string) error {
 	return err
 }
 
+// UserAPICheck returns an error if the Che user API cannot be reached successfully, returns nil otherwise
+func (c *Client) UserAPICheck() error {
```
alexeykazakov, line 124: Any tests for this method?

pkg/controller/memberstatus/memberstatus_controller.go
rajivnathan: That's a really good point. I guess it has an effect on the status notifications? Opened https://issues.redhat.com/browse/CRT-900

pkg/configuration/configuration.go
rajivnathan: Yes, that's what I was trying to make clear with the 🚨  IMPORTANT 🚨  message at the top of the description but I guess that didn't work 😄 . Alternatively, we could make the default `true`.

pkg/che/che.go
rajivnathan: I figured the more important thing is the response status. If we couldn't read the body then the body string would just be empty in the error but it's not that important to know reading the body failed?

pkg/controller/memberstatus/memberstatus_controller.go
rajivnathan: I thought just `che` could be a bit too ambiguous but yeah maybe it's better for consistency and having the tag match the component name. Done in cfd576103ff935a95eeb0bf2b47a24bfebb4c7a6

pkg/controller/memberstatus/memberstatus_controller.go
rajivnathan: Yes, I refactored this stuff too much and lost track. 🙂  I was originally using it in the useraccount_controller too so it was a common function but then I created the `IsCheUserDeletionEnabled` function and used it there instead. This function is probably not necessary now but I'll still go ahead and use it because it makes the condition slightly more understandable. Done in 5fe5a24cdf688f31af552f325423ac3b5cc77129

pkg/configuration/configuration.go
rajivnathan: Ah ok, I didn't know about that. Makes sense though. I'll do that.

pkg/controller/memberstatus/memberstatus_controller.go
alexeykazakov: Status notifications relies on the overall Ready condition only. So no direct impact on that. For now it's rather about avoiding confusion with that LastTransitionTime timestamps in the component conditions.

pkg/che/che.go
rajivnathan: Ah yes, added in 128cdab96a227412f60ea41804ed4a7e53106211

pkg/configuration/configuration.go
alexeykazakov: Haha. I missed that :)

pkg/che/che.go
alexeykazakov: Ah.. I was a bit sleepy when was reviewing the code. It's the error about reading the body.. So, right, we can leave it as is then :)

go.mod
```diff
@@ -38,4 +38,6 @@ replace (
 	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20200410145947-61e04a5be9a6 // avoids case-insensitive import collision: "github.com/googleapis/gnostic/openapiv2" and "github.com/googleapis/gnostic/OpenAPIv2"
 )
 
+replace github.com/codeready-toolchain/api => github.com/rajivnathan/api v0.0.0-20201218021903-cf2c667c66b9
```
alexeykazakov, line 41: Friendly reminder to remove that :)

rajivnathan: Done in https://github.com/codeready-toolchain/member-operator/pull/231/commits/dc65c9b62207990d5524893de4c95d952972765c

alexeykazakov (approved):

PR #232, opened 2021-01-28
Webhook deploy config

rajivnathan:
By default the users' pods webhook will be deployed, however in some cases (eg. e2e tests) there can be multiple member operators installed in the same cluster. In those cases only 1 webhook is needed because the MutatingWebhookConfiguration is a cluster-scoped resource and naming can conflict.

This PR adds a new configuration option that can toggle whether or not the member operator will deploy the webhook and create the associated `MutatingWebhookConfiguration` resource as well. It will be used by the e2e test setup to disable the deployment of the users' pods webhook.

Related issue:
https://issues.redhat.com/browse/CRT-770

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -167,7 +167,7 @@ func main() {
 		}
 
 		// By default the users' pods webhook will be deployed, however in some cases (eg. e2e tests) there can be multiple member operators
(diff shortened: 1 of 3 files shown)


alexeykazakov (approved):
Looks good!

MatousJobanek (approved):
:+1:

PR #234, opened 2021-02-16
CRT-953 support multiple roles for nodes

sbryzak:
Fixes https://issues.redhat.com/browse/CRT-953

Adds support for multi-role nodes.

### DIFF
--- a/pkg/controller/memberstatus/memberstatus_controller.go
+++ b/pkg/controller/memberstatus/memberstatus_controller.go
@@ -232,9 +232,12 @@ func (r *ReconcileMemberStatus) loadCurrentResourceUsage(reqLogger logr.Logger, 
 	for _, nodeMetric := range nodeMetricsList.Items {
 		if memoryUsage, usageFound := nodeMetric.Usage["memory"]; usageFound {
 			if nodeInfo, nodeFound := allocatableValues[nodeMetric.Name]; nodeFound {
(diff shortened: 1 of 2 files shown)


xcoulon:
@sbryzak could you add some tests to verify these changes?

sbryzak:
@xcoulon we actually have e2e tests for this already which are broken when running in CRC, this fix allows the tests to pass again at least in CRC so if we get a pass in CI also then we should be well covered.

MatousJobanek:
Looks good :+1: 
I agree that there is no need for e2e tests, but could you please add a unit test for the case when one node is used for both roles?

xcoulon:
> Looks good 👍
> I agree that there is no need for e2e tests, but could you please add a unit test for the case when one node is used for both roles?

yes, at least a unit test would be nice, since if I understand correctly, our current e2e tests on OpenShift CI do pass with our without this fix.

alexeykazakov:
Agree that we need to at least some tests. The CRC test coverage is not enough because our PR checks do not run against CRC. We can't rely on manual testing.
Let's add some simple unit tests.

sbryzak:
I've added support in the test for specifying multiple roles for a node, and added a unit test that tests a multi-role node.

alexeykazakov (approved):
Thanks for the test Shane.

MatousJobanek (approved):
Cool :+1: :1st_place_medal:

PR #236, opened 2021-02-23
Add predicate to ignore CRQ status changes

rajivnathan:
Related issue: https://issues.redhat.com/browse/CRT-941

As part of the split up of the cluster resource quotas there will be even more reconciles of the NSTemplateSet controller since it's watching for changes to cluster level resources. However, the controller is only really interested in changes to the spec so this PR adds a watch predicate for generation changes so that status changes will not trigger reconciles.

### DIFF
--- a/pkg/controller/nstemplateset/nstemplateset_controller.go
+++ b/pkg/controller/nstemplateset/nstemplateset_controller.go
@@ -2,6 +2,7 @@ package nstemplateset
 
 import (
 	"context"
+	"reflect"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/pkg/apis/toolchain/v1alpha1"
 	"github.com/codeready-toolchain/member-operator/pkg/configuration"
@@ -11,7 +12,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/go-logr/logr"
-	"github.com/operator-framework/operator-sdk/pkg/predicate"
 	errs "github.com/pkg/errors"
 	"github.com/redhat-cop/operator-utils/pkg/util"
 	corev1 "k8s.io/api/core/v1"
@@ -20,9 +20,11 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
@@ -75,7 +77,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// watch for all cluster resource kinds associated with an NSTemplateSet
 	for _, clusterResource := range clusterResourceKinds {
 		// only reconcile generation changes for cluster resources
-		if err := c.Watch(&source.Kind{Type: clusterResource.objectType}, commoncontroller.MapToOwnerByLabel("", toolchainv1alpha1.OwnerLabelKey), predicate.GenerationChangedPredicate{}); err != nil {
+		if err := c.Watch(&source.Kind{Type: clusterResource.objectType}, commoncontroller.MapToOwnerByLabel("", toolchainv1alpha1.OwnerLabelKey), labelsAndGenerationPredicate{}); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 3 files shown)


pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -71,10 +71,19 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 		return err
 	}
+
+	// watch for all cluster resource kinds associated with an NSTemplateSet
 	for _, clusterResource := range clusterResourceKinds {
-		// watch for all cluster resource kinds associated with an NSTemplateSet
-		if err := c.Watch(&source.Kind{Type: clusterResource.objectType}, commoncontroller.MapToOwnerByLabel("", toolchainv1alpha1.OwnerLabelKey)); err != nil {
-			return err
+		switch clusterResource.gvk.Kind {
+		case "ClusterResourceQuota":
```
alexeykazakov, line 78: Let's create a const for `ClusterResourceQuota` and reuse it here and in https://github.com/codeready-toolchain/member-operator/blob/eb709dea0937d7645ba9341a42f29ec8c1439e28/pkg/controller/nstemplateset/cluster_resources.go#L55

alexeykazakov (approved):
Looks good! Just one minor comment.

pkg/controller/nstemplateset/nstemplateset_controller.go
```diff
@@ -71,10 +71,19 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err := c.Watch(&source.Kind{Type: &corev1.Namespace{}}, commoncontroller.MapToOwnerByLabel("", toolchainv1alpha1.OwnerLabelKey)); err != nil {
 		return err
 	}
+
+	// watch for all cluster resource kinds associated with an NSTemplateSet
 	for _, clusterResource := range clusterResourceKinds {
-		// watch for all cluster resource kinds associated with an NSTemplateSet
-		if err := c.Watch(&source.Kind{Type: clusterResource.objectType}, commoncontroller.MapToOwnerByLabel("", toolchainv1alpha1.OwnerLabelKey)); err != nil {
-			return err
+		switch clusterResource.gvk.Kind {
```
MatousJobanek, line 77: I'm wondering if we really need to watch for the changes either in the status or in the `metadata` part in any other cluster-scoped object. I guess that not, so we could use the same predicate for all of them, right?

Idler stores/updates information about the monitored pods in the status part which triggers additional (unnecessary) reconciliation as well - not so often as ClusterResourceQutoa, but still it's not really needed.

MatousJobanek: Another option would be to create a custom predicate that would trigger the reconciliation:
* either based on the changed generation number (the same as `GenerationChangedPredicate`)
* or based on a change in either a label or in an annotation

That should cover all needed actions we would need to monitor also in the future (in case we would need to rely for example on a label).

alexeykazakov: Good points Matous. If label/annotation change does not update the resource generation then yeah.. I think we want to watch them too.
For now I think it's safe to switch to generationchangedpredicate for all resources and we could introduce a new predicate for labels in a separate PR. I would stop watching statuses ASAP.

MatousJobanek: So I guess that we can drop this `switch` and use the `GenerationChangedPredicate` predicate for all cluster-scoped resources. In the next PR, we can introduce our own predicate that will monitor also changes in the labels & annotations - feel free to create a separate Jira for it @rajivnathan if you want

rajivnathan: Ok, dropped the switch and used `GenerationChangedPredicate` in b719ff0ba02ab3aff3595d84e9c851b45eb2f53d

rajivnathan: I think I'm a bit concerned adding a predicate for labels & annotations because we'd need to do a comparison for both for each and every event, I think that could be a lot of processing so we should be really sure we need to do it.

Something like this:
```
type labelAndAnnotationsPredicate struct {
	rtpredicate.Funcs
}

func (labelAndAnnotationsPredicate) Update(e event.UpdateEvent) bool {
	return !reflect.DeepEqual(e.MetaOld.GetLabels(), e.MetaNew.GetLabels()) || !reflect.DeepEqual(e.MetaOld.GetAnnotations(), e.MetaNew.GetAnnotations())
}
```

Maybe we need to see if we are relying on any label/annotation changes and see if those should actually belong to spec instead?

alexeykazakov: Annotations and labels are maps. We should be careful when comparing them.
The main reason why we rely on labels in some cases instead of using Spec is the limitations in CRDs. You can't use CRD Spec as a selector (you can't look up for resources with specific Spec field). You can use only name or labels for that.
We could limit such predicate to our labels/annotations only (the ones which starts with our prefix). But IMO it's a low priority. Let's just create a JIRA and move one. We will address the issue if/when we face the case when we need to watch label/annotation changes.

pkg/controller/nstemplateset/nstemplateset_controller.go
rajivnathan: I guess this is not needed now that we removed the switch statement

alexeykazakov: Yep

pkg/controller/nstemplateset/nstemplateset_controller.go
rajivnathan: Ok, if we're only going to be looking for some specific labels/annotations then that would be better. I'll create the issue.

rajivnathan: It turns out that the GenerationChangedPredicate is not enough. Without a predicate for labels the `ChangeTierRequest` is broken so we'll need to include the label predicate in this PR as well.

I just had a chat with Matous and he made some good points.
1. We don't seem to be using annotations for anything so we can probably only need to consider labels
2. We can just compare the maps because there is unlikely to be any labels other than ours.

So I will go ahead and update this PR to include labels as well.

rajivnathan:
@MatousJobanek I've moved the predicate to toolchain-common and updated the code to reference it. Please let me know if there's anything else that needs to be addressed before merging. Thanks!

MatousJobanek (approved):

PR #248, opened 2021-05-12
Use constant for member status name

rajivnathan:
No need for memberstatus name to be configurable, we'll just hardcode it to "toolchain-member-status"


alexeykazakov (approved):

MatousJobanek (approved):
:+1:

PR #252, opened 2021-05-14
fix: point to operator-cd.yml instead of workflow.yml in badge link

MatousJobanek:


alexeykazakov (approved):

xcoulon (approved):
ah, my apologies, I did not realize I broke the badge when renaming the GitHub workflow file :(

PR #254, opened 2021-05-20
chore: update CRDs and api repo with new kubebuilder

MatousJobanek:
result of codeready-toolchain/api#237
* CRDs use a new name format: `toolchain.dev.openshift.com_<crd-plural-name>.yaml`
* sample CRs renamed to use the new name format as well
* slightly modified generated CRD as a result of using new kubebuilder (controller-tools)
* small changes in go.mod, so we can use the new controller-tools 0.4.1 together with client-go 0.18.3
* update the imports to the api repo/types


cmd/manager/main.go
```diff
@@ -9,7 +9,7 @@ import (
 	"runtime"
 	"strings"
 
-	api "github.com/codeready-toolchain/api/pkg/apis"
+	api "github.com/codeready-toolchain/api/api/v1alpha1"
```
xcoulon, line 12: same comment as in https://github.com/codeready-toolchain/toolchain-common/pull/177#discussion_r636104231 regarding the alias for the `github.com/codeready-toolchain/api/api/v1alpha1` package

pkg/apis/apis.go
```diff
@@ -1,7 +1,7 @@
 package apis
 
 import (
-	"github.com/codeready-toolchain/api/pkg/apis"
+	api "github.com/codeready-toolchain/api/api/v1alpha1"
```
xcoulon, line 4: same here, let's use `toolchainv1alpha1` if possible

xcoulon (approved):
Thanks for undertaking these changes @MatousJobanek 🙌 
I've left a comment with regards to the alias for our `github.com/codeready-toolchain/api/api/v1alpha1` package (as I did in other PRs), but it's not something we need to address in the PR.

alexeykazakov (approved):

cmd/manager/main.go
MatousJobanek: I already did that locally yesterday, but forgot to push the changes to GH: b8f6c51

PR #255, opened 2021-05-20
Delete autoscaler buffer app if disabled in config

alexeykazakov:
This PR covers the case when the buffer app was previously deployed but then disabled in the member operator. In this case we need to delete it from the cluster.

Later, when we switch to MemberConfiguration CRD we can move this logic to a controller which watches the configuration CR and deploys or deletes the app without the need of re-deploying the operator.

We can't really test this in e2e tests because of the static nature of our configuration atm.


rajivnathan (approved):
Looks good!

sbryzak (approved):

ranakan19 (approved):
LGTM

PR #262, opened 2021-06-14
ensure namespace is deleted

ranakan19:
related to story: https://issues.redhat.com/browse/CRT-897
related PR (e2e) : https://github.com/codeready-toolchain/toolchain-e2e/pull/336

Added a condition to check that namespace is actually deleted via the api delete before returning response.


controllers/nstemplateset/namespaces.go
```diff
@@ -175,7 +183,16 @@ func (r *namespacesManager) delete(logger logr.Logger, nsTmplSet *toolchainv1alp
...
-			return true, nil
+
+			if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: ns.Name}, &corev1.Namespace{}); err != nil {
+				if errors.IsNotFound(err) {
+					return true, nil // namespace was actually deleted and thus not found by get
+				}
+			}
+			// No error implies namespace was not deleted
+			v := validationError{}
+			err = &v
```
rajivnathan, line 194: You can just use `fmt.Errorf("Namespace deletion wasn't complete")`

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -969,6 +969,51 @@ func TestDeleteNSTemplateSet(t *testing.T) {
 		AssertThatNSTemplateSet(t, namespaceName, username, r.Client).
 			DoesNotHaveFinalizer() // finalizer was not added and nothing else was done
 	})
+
+	t.Run("NSTemplateSet not deleted untill namespace is deleted", func(t *testing.T) {
```
rajivnathan, line 973: typo
```suggestion
	t.Run("NSTemplateSet not deleted until namespace is deleted", func(t *testing.T) {
```

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -969,6 +969,51 @@ func TestDeleteNSTemplateSet(t *testing.T) {
...
+		// given an NSTemplateSet resource and 1 active user namespaces ("dev")
+		nsTmplSet := newNSTmplSet(namespaceName, username, "advanced", withNamespaces("abcde11", "dev"), withDeletionTs(), withClusterResources("abcde11"))
+		devNS := newNamespace("advanced", username, "dev", withTemplateRefUsingRevision("abcde11"))
+
+		r, fakeClient := prepareController(t, nsTmplSet, devNS)
+		req := newReconcileRequest(namespaceName, username)
+
+		fakeClient.MockDelete = func(ctx context.Context, obj runtime.Object, opts ...client.DeleteOption) error {
+			fmt.Printf("Calling Mock Delete")
```
rajivnathan, line 982: Maybe this can be made slightly more clear
```suggestion
			fmt.Printf("Skipping deletion")
```

rajivnathan (commented):
Nice job! 👍  Just left a couple minor comments

controllers/nstemplateset/namespaces.go
ranakan19: that makes sense, updated.

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -969,6 +969,51 @@ func TestDeleteNSTemplateSet(t *testing.T) {
...
+	t.Run("NSTemplateSet not deleted until namespace is deleted", func(t *testing.T) {
+		// given an NSTemplateSet resource and 1 active user namespaces ("dev")
+		nsTmplSet := newNSTmplSet(namespaceName, username, "advanced", withNamespaces("abcde11", "dev"), withDeletionTs(), withClusterResources("abcde11"))
+		devNS := newNamespace("advanced", username, "dev", withTemplateRefUsingRevision("abcde11"))
+
+		r, fakeClient := prepareController(t, nsTmplSet, devNS)
+		req := newReconcileRequest(namespaceName, username)
+
+		fakeClient.MockDelete = func(ctx context.Context, obj runtime.Object, opts ...client.DeleteOption) error {
+			fmt.Printf("Skipping deletion")
```
alexeykazakov, line 982: Zombie?
```suggestion
```

alexeykazakov (commented):
Looks good!

But what about e2e tests?

controllers/nstemplateset/nstemplateset_controller_test.go
ranakan19: @alexeykazakov wrote tests here - https://github.com/codeready-toolchain/toolchain-e2e/pull/322 - but the desired scenario is not achieved. namespace is still left in terminating state. Checked the logs - it stops returning [this](https://github.com/ranakan19/member-operator/blob/CRT897/controllers/nstemplateset/namespaces.go#L188) after returning this error twice. I'm investigating the reason.

alexeykazakov: Please delete this Printf

sbryzak (approved):
Besides Alexey's suggestions, the rest looks good to me.

controllers/nstemplateset/namespaces.go
```diff
@@ -175,8 +178,15 @@ func (r *namespacesManager) delete(logger logr.Logger, nsTmplSet *toolchainv1alp
 			if err := r.Client.Delete(context.TODO(), &ns); err != nil {
 				return false, r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusTerminatingFailed, err, "failed to delete user namespace '%s'", ns.Name)
 			}
-			return true, nil
```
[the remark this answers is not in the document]
ranakan19: @alexeykazakov: 
> if err != nil {
		// Wrap the error with the status error
		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusTerminatingFailed, err, "failed to ensure namespace deletion")
	}

what is the benefit of wrapping err with statusUpdate here, when we are only returning API-level errors from ensureDelete?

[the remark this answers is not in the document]
alexeykazakov: If there is some error then we will see it in the status. If we don't wrap the errors then the only way of seeing what's wrong would be looking at logs.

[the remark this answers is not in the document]
alexeykazakov: That's the whole point of wrapping errors with the status update.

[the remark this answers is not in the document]
alexeykazakov: I've updated the ensureDelete(). Check it out: https://github.com/codeready-toolchain/member-operator/pull/262/commits/d9f83b3c0592b9fd2e8b21301edc1e5204b2e424

controllers/nstemplateset/nstemplateset_controller.go
```diff
@@ -176,13 +175,17 @@ func (r *Reconciler) deleteNSTemplateSet(logger logr.Logger, nsTmplSet *toolchai
 	username := nsTmplSet.GetName()
 
 	// delete all namespace one by one
-	deletedAny, err := r.namespaces.delete(logger, nsTmplSet)
-	if err != nil || deletedAny {
+	allDeleted, err := r.namespaces.ensureDeleted(logger, nsTmplSet)
+	// when err, status Update will not trigger reconcile, sending returning error.
+	if err != nil {
+		return reconcile.Result{}, err
```
alexeykazakov, line 181:
```suggestion
		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusTerminatingFailed, err, "failed to ensure namespace deletion")
```

controllers/nstemplateset/nstemplateset_controller.go
```diff
@@ -176,13 +175,17 @@ func (r *Reconciler) deleteNSTemplateSet(logger logr.Logger, nsTmplSet *toolchai
...
+	}
+	if !allDeleted {
 		return reconcile.Result{}, nil
```
alexeykazakov, line 184: Let's add some comments here.
```suggestion
		// One or more namespaces is not yet deleted. We can stop here. When it's finally deleted it will trigger another reconcile.
		return reconcile.Result{}, nil
```

controllers/nstemplateset/namespaces.go
```diff
@@ -189,7 +189,7 @@ func (r *namespacesManager) ensureDeleted(logger logr.Logger, nsTmplSet *toolcha
 			return false, r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusTerminatingFailed, err, "failed to get user namespace '%s'", ns.Name)
 		}
 		// No error implies namespace is in terminating state but has not been deleted yet, update status and returns false so we will re-try when the namespace is actually deleted
-		return false, r.setStatusTerminatingFailed(nsTmplSet, fmt.Sprintf("user namespace %s deletion was triggered but is not complete yet, something could be blocking ns deletion", ns.Name))
+		return false, fmt.Errorf("user namespace %s deletion was triggered but is not complete yet, something could be blocking ns deletion", ns.Name)
```
alexeykazakov, line 192: Why did you change it? We are not supposed to return any error here. Just update the status and wait for the next reconcile which will be triggered when the namespace is actually deleted.

ranakan19: thank you for catching this! there was some confusion, has been corrected now.

controllers/nstemplateset/nstemplateset_controller.go
ranakan19: done

controllers/nstemplateset/namespaces_test.go
```diff
@@ -561,38 +561,36 @@ func TestDeleteNamespsace(t *testing.T) {
...
 		t.Run("delete the first namespace", func(t *testing.T) {
 			// when
-			deleted, err := manager.delete(logger, nsTmplSet)
+			allDeleted, err := manager.ensureDeleted(logger, nsTmplSet)
 
 			// then
 			require.NoError(t, err)
-			assert.True(t, deleted)
+			assert.False(t, allDeleted)
 			// get the first namespace and check its deletion timestamp
```
alexeykazakov, line 571:
```suggestion
			// get the first namespace and check it's deleted
```

controllers/nstemplateset/namespaces_test.go
```diff
@@ -561,38 +561,36 @@ func TestDeleteNamespsace(t *testing.T) {
...
 			t.Run("delete the second namespace", func(t *testing.T) {
 				// when
-				deleted, err := manager.delete(logger, nsTmplSet)
+				allDeleted, err := manager.ensureDeleted(logger, nsTmplSet)
 
 				// then
 				require.NoError(t, err)
-				assert.True(t, deleted)
+				assert.False(t, allDeleted)
 				// get the second namespace and check its deletion timestamp
```
alexeykazakov, line 582:
```suggestion
				// get the second namespace and check it's deleted
```

controllers/nstemplateset/namespaces_test.go
```diff
@@ -602,11 +600,11 @@ func TestDeleteNamespsace(t *testing.T) {
...
-		deleted, err := manager.delete(logger, nsTmplSet)
+		allDeleted, err := manager.ensureDeleted(logger, nsTmplSet)
 
 		// then
 		require.NoError(t, err)
-		assert.False(t, deleted)
+		assert.True(t, allDeleted)
 	})
 
 	t.Run("failed to fetch namespaces", func(t *testing.T) {
```
alexeykazakov, line 610: Please also add a test for the case when client.Delete() fails.

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -968,6 +968,63 @@ func TestDeleteNSTemplateSet(t *testing.T) {
...
+				obj.DeletionTimestamp = &deletionTs
+				if err:= r.Client.Update(context.TODO(), obj); err !=nil {
+					return err
+				}
+			}
+			return nil
+		}
+		// first reconcile, deletion is triggered
+		_, err := r.Reconcile(context.TODO(), req)
+		require.NoError(t, err)
```
alexeykazakov, line 993: Also check that it returned an empty Result{}

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -968,6 +968,63 @@ func TestDeleteNSTemplateSet(t *testing.T) {
...
+		//then second reconcile to check if namespace has actually been deleted
+		_, err = r.Reconcile(context.TODO(), req)
+		require.NoError(t, err)
```
alexeykazakov, line 996: Also check that it returned an empty Result{}

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -968,6 +968,63 @@ func TestDeleteNSTemplateSet(t *testing.T) {
...
+
+		// get the first namespace and check that it is not deleted
+		firstNSName := fmt.Sprintf("%s-dev", username)
+		AssertThatNamespace(t, firstNSName, r.Client).DoesExist()
```
alexeykazakov, line 1000: You are actually testing your MockDelete() here. Not the controller. Instead of checking that the namespace still exists it's better to test that the corresponding namespace has the deletion timestamp set in lines 993 and 996. So you would make sure that the delete function was called.

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -968,6 +968,63 @@ func TestDeleteNSTemplateSet(t *testing.T) {
...
+			HasFinalizer().
+			HasConditions(UnableToTerminate("user namespace johnsmith-dev deletion was triggered but is not complete yet, something could be blocking ns deletion"))
+
+		// remove deletion timestamp, actual delete is now triggered
+		userNS := &corev1.Namespace{}
+		err = r.Client.Get(context.TODO(), types.NamespacedName{Name: firstNSName}, userNS)
+		require.NoError(t, err)
+		userNS.DeletionTimestamp = nil
+		err = r.Client.Update(context.TODO(), userNS)
+		require.NoError(t, err)
```
alexeykazakov, line 1012: What you are doing is actually making the namespace active again and the controller delete it again but with a regular Delete() in the fake client. This is not what we want. Instead of this we need to actually delete the resource from the fake client here, in the test. In this case you would simulate the actual deletion by kube.

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -968,6 +968,63 @@ func TestDeleteNSTemplateSet(t *testing.T) {
...
+		fakeClient.MockDelete = nil //now removing the mockDelete
+
+		//reconcile, deletion is triggered
+		_, err = r.Reconcile(context.TODO(), req)
```
alexeykazakov, line 1016: See my previous comment. The deletion has been already triggered in the previous reconcile. When you fix this test flow you will be able to skip this reconcile.

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -968,6 +968,63 @@ func TestDeleteNSTemplateSet(t *testing.T) {
...
+		require.NoError(t, err)
+		// second reconcile triggered when ns is deleted
+		_, err = r.Reconcile(context.TODO(), req)
```
alexeykazakov, line 1019: Check the result is empty.

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -968,6 +968,63 @@ func TestDeleteNSTemplateSet(t *testing.T) {
...
+		require.NoError(t, err)
+
+		// get the first namespace and check it IS deleted
+		AssertThatNamespace(t, firstNSName, r.Client).DoesNotExist()
+		// Check that nsTemplateSet is set to delete
+		AssertThatNSTemplateSet(t, namespaceName, username, r.Client).DoesNotHaveFinalizer().HasConditions(Terminating())
+
```
alexeykazakov, line 1026: But this is not all.
Now you need to repeat all these steps from above for the second namespace and in the end make sure that the NSTemplateSet is finally deleted.

alexeykazakov (commented):
The runtime code looks good! But the tests need a bit of love :)

controllers/nstemplateset/nstemplateset_controller_test.go
ranakan19: the test case only has one namespace, so the NStemplateset is deleted after the first ns itself.

alexeykazakov: Then we need to add another namespace to the test to cover all the usecases. Unless we already have such a test then why duplicating it here?

ranakan19: correction : there are two namespaces in the testcase. 
There are existing tests - https://github.com/codeready-toolchain/member-operator/blob/master/controllers/nstemplateset/nstemplateset_controller_test.go#L870-L971  which check the deletion of NSTemplateSet. The reasons this new test case was added:
1. The existing tests checked that ns had deletionTimeStamp and not actual deletion
2. To test the negative scenario when the NSTemplateSet should not be deleted when the ns is not deleted.

controllers/nstemplateset/namespaces_test.go
```diff
@@ -560,6 +560,18 @@ func TestDeleteNamespace(t *testing.T) {
...
+			cl.MockDelete = func(ctx context.Context, obj client.Object, opts ...client.DeleteOption) error {
+				return fmt.Errorf("client.Delete() failed")
+			}
+			// when
+			allDeleted, err := manager.ensureDeleted(logger, nsTmplSet)
+			require.Error(t, err)
+			require.False(t, allDeleted)
+
+			// set mockDelete to nil
+			cl.MockDelete = nil
```
alexeykazakov, line 573: Just move this test out of the `t.Run("with 2 user namespaces to delete", func(t *testing.T)` because it's not actually a necessarily a subtest of the 2 namespace case. And create a new `manager, cl := prepareNamespacesManager(t, nsTmplSet, devNS, codeNS)` in the beggingin of the test so you don't have to reuse this client in other tests so you can't break other tests.

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -997,36 +996,45 @@ func TestDeleteNSTemplateSet(t *testing.T) {
...
 		result, err = r.Reconcile(context.TODO(), req)
 		require.Empty(t, result)
 		require.NoError(t, err)
 
 		// get the first namespace and check it IS deleted
 		AssertThatNamespace(t, firstNSName, r.Client).DoesNotExist()
 		// Check that nsTemplateSet is set to delete
-		AssertThatNSTemplateSet(t, namespaceName, username, r.Client).DoesNotHaveFinalizer().HasConditions(Terminating())
+		AssertThatNSTemplateSet(t, namespaceName, username, r.Client).
+			DoesNotHaveFinalizer().HasConditions(Terminating())
```
alexeykazakov, line 1037: Looks good!
So now we just need to delete the second namespace and check that the finalizer is finally removed from the NSTemplateSet.

ranakan19: do we need to have two namespaces? Is it ok if the testcase only has one namespace associated with the user?

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -1025,13 +1030,24 @@ func TestDeleteNSTemplateSet(t *testing.T) {
...
 		AssertThatNamespace(t, firstNSName, r.Client).DoesNotExist()
+
+		// Check that nsTemplateSet still has finalizer
+		AssertThatNSTemplateSet(t, namespaceName, username, r.Client).
+			HasFinalizer().HasConditions(Terminating())
+
+		// deletion of secondNS would trigger another reconcile
+		result, err = r.Reconcile(context.TODO(), req)
+		require.Empty(t, result)
+		require.NoError(t, err)
```
alexeykazakov, line 1048: Is this one reconcile is enough here? Don't we have to call another reconcile?
The first one will delete the namespace (`client.Delete()` will be called) and the second one will make sure there is no namespaces left so it can delete the finalizer?

alexeykazakov (approved):
Looks good! I just have one question.

controllers/nstemplateset/namespaces.go
```diff
@@ -156,30 +159,40 @@ func (r *namespacesManager) ensureInnerNamespaceResources(logger logr.Logger, ns
...
 		if !util.IsBeingDeleted(&ns) {
-			logger.Info("deleting a user namepace associated with the deleted NSTemplateSet", "namespace", ns.Name)
+			logger.Info("deleting a user namespace associated with the deleted NSTemplateSet", "namespace", ns.Name)
 			if err := r.Client.Delete(context.TODO(), &ns); err != nil {
 				return false, r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusTerminatingFailed, err, "failed to delete user namespace '%s'", ns.Name)
 			}
-			return true, nil
+			return false, nil // The namespace deletion is triggered so we should stop here. When the namespace is actually deleted the reconcile will be triggered again
 		}
+		if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: ns.Name}, &corev1.Namespace{}); err != nil {
```
MatousJobanek, line 186: just curious - why are you getting the namespace that was alraedy fetched by `fetchNamespaces`? You should already have the latest version of the namespace, don't you?

controllers/nstemplateset/namespaces.go
```diff
@@ -156,30 +159,40 @@ func (r *namespacesManager) ensureInnerNamespaceResources(logger logr.Logger, ns
...
+			if errors.IsNotFound(err) {
+				continue // This namespace is gone. Check the next one.
+			}
+			return false, r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusTerminatingFailed, err, "failed to get user namespace '%s'", ns.Name)
+		}
+		// No error implies namespace is in terminating state but has not been deleted yet, update status and returns false so we will re-try when the namespace is actually deleted
+		return false, r.setStatusTerminatingFailed(nsTmplSet, fmt.Sprintf("user namespace %s deletion was triggered but is not complete yet, something could be blocking ns deletion", ns.Name))
```
MatousJobanek, line 193: Do we really want to set the condition to the failed state when any of the namespaces is in a terminating state? I mean - it would make sense to do it after some time (eg. 1 minute) but I'm not sure if we want to do it immediately.

controllers/nstemplateset/namespaces_test.go
```diff
@@ -533,7 +533,7 @@ func TestEnsureNamespacesFail(t *testing.T) {
 
 }
 
-func TestDeleteNamespsace(t *testing.T) {
+func TestDeleteNamespace(t *testing.T) {
```
MatousJobanek, line 536: is there any test-case where you would verify a situation when one namespace is in terminating state - has termination timestamp?

controllers/nstemplateset/namespaces.go
MatousJobanek: in addition to that, we alraedy have a similar logic in MUR - when the UserAccount is in terminating state for more than one minute, then it starts returning an error https://github.com/codeready-toolchain/host-operator/blob/master/controllers/masteruserrecord/masteruserrecord_controller.go#L292-L299 
So, I'm not sure if we need to set the failed status in the NsTemplateSet

controllers/nstemplateset/namespaces.go
ranakan19: I think you are right, i can get rid of the get. I believe in the initial implementation I was getting after delete in the same block where the latest version of ns was needed after delete.But with the current implementation the get is unnecessary. I will remove this.

controllers/nstemplateset/namespaces_test.go
ranakan19: Not in here, but [this](https://github.com/codeready-toolchain/toolchain-e2e/pull/336/files#diff-e413827761efc65be1d2b683f71859e11dc041eddacaef444a8fcb11c409aa37R323) checks for NS in terminating state and having deletion timestamp.

controllers/nstemplateset/namespaces.go
ranakan19: its not exactly immediately, its one reconcile after the deletion of namespace was triggered. If the namespace is not deleted until the next reconcile, to me it could be an indication that something is not right since namespace deletion is fairly quick. 
Regarding your second comment, if a namespace is not deleted it would block all the owning resources, which is why was updating the failed status in NStemplateset as well. 
Would you rather have a wait time of 1min before updating the status of NSTemplateSet as well?

controllers/nstemplateset/namespaces_test.go
ranakan19: Also, this particular line is just a typo fix

controllers/nstemplateset/nstemplateset_controller_test.go
ranakan19: Ideally it should, but since delete is triggered in the test itself instead of reconcile, the entire process takes one reconcile less.

controllers/nstemplateset/namespaces_test.go
MatousJobanek: I meant if there is a test-case where you would 
```
// given
create Namespace with a deletion timestamp and add it to the fake client

// when
allDeleted, err := manager.ensureDeleted(logger, nsTmplSet)

// then
require.NoError(t, err)
assert.False(t, allDeleted)
in other words: verify that it's still waiting for the namespace being completely deleted
```

ranakan19: added it now @MatousJobanek

MatousJobanek: awesome, thanks :+1:

PR #263, opened 2021-06-15
feat: migrate member to operator-sdk 1.8.0

MatousJobanek:
## Migrates member-operator repo to operator-sdk 1.8.0

* adds kubebuilder's `PROJECT` file with the definition of the operator
* adds `budle/` & ` bundle.Dockerfile` to `.gitignore` since they are generated dynamically 
* cleans README from obsolete parts
* updates CRDs with the latest changes
* updates dependencies to use k8s v0.20.2 - as a result:
    *  updates the mappers used in controllers to match the latest API - it doesn't expect a type, but a function
    * migrates most of the code from `runtime.Object` to `client.Object`, but here is still for an improvement - will be done in next PR
    * use the new API for defining the watched resources in controllers: `ctrl.NewControllerManagedBy`
    * changes `Reconcile(request ctrl.Request)` to `Reconcile(ctx context.Context, request ctrl.Request)`
* updated go to 1.16
    * in go.mod
    * in GH Actions workflows  
* don't store log inside the `Reconciler` type but create it using `log.FromContext(ctx)`
* start using `GetWatchNamespace` & `GetOperatorName` from toolchain-common
* deletes `hack/` directory - it's not needed anymore
* as for the operator's `main.go` file
    * it was moved to the root of the repo
    * updates the content to get as close as possible to the version generated by latest operator-sdk/kubebuilder
* as for make targets
    * drops `dev.mk` as it's not needed anymore (or at least for now)
    * gets `kustomize` binary via `get-tool.mk` so it can be used when generating operator manifests
    * when generating assets it uses `go-bindata` binary fetched via `get-tool.mk`  as well
    * significant changes in `manifests.mk` file to match the latest changes in CI/CD setup/logic
    * also updates `test.mk` so it takes into account also the fact that it is executed in GH Action workflow
* as for Dockerfile.tools
    * updated go to 1.16
    * updated oc to the latest 4.7
    * dropped operator-sdk as is not needed in openshift-ci anymore

plus:
* changes the name of the index image to `member-operator-index` ` (for better isolation)

### TODO in next PRs
* remove OLM related files and directories from `deploy/` folder
* add CRD samples
* investigate scorecard tests
* clean openshift/release config for openshift-ci from obsolete parts


make/manifests.mk
```diff
@@ -1,68 +1,84 @@
...
-PATH_TO_CD_GENERATE_FILE=scripts/generate-cd-release-manifests.sh
-PATH_TO_BUNDLE_FILE=scripts/push-bundle-and-index-image.sh
-PATH_TO_RECOVERY_FILE=scripts/recover-operator-dir.sh
-PATH_TO_OLM_GENERATE_FILE=scripts/olm-catalog-generate.sh
+PATH_TO_CD_GENERATE_FILE=generate-cd-release-manifests.sh
+PATH_TO_BUNDLE_FILE=push-bundle-and-index-image.sh
+PATH_TO_INSTALL_OPERATOR=install-operator.sh
+
+#OWNER_AND_BRANCH_LOCATION=codeready-toolchain/api/master
+OWNER_AND_BRANCH_LOCATION=MatousJobanek/api/operator-sdk-1.8.0
```
alexeykazakov, line 7: A reminder

alexeykazakov (approved):

controllers/idler/idler_controller.go
```diff
@@ -71,8 +55,9 @@ type Reconciler struct {
 // Note:
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
-func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	logger := r.Log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+	//logger := r.Log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
```
xcoulon, line 60: this line can be removed

controllers/memberstatus/memberstatus_controller_test.go
```diff
@@ -39,6 +38,7 @@ import (
 var requeueResult = reconcile.Result{RequeueAfter: 5 * time.Second}
 
 const defaultMemberOperatorName = "member-operator"
+const defaultMemberOperatorDeploymentName = "member-operator-controller-manager"
```
xcoulon, line 41: I wonder if we'll find ourselves with 2 deployments when we install the new version of the operator, with this deployment name? (Not a blocker, just a question)

main.go
```diff
@@ -121,43 +100,27 @@ func main() {
...
+		Namespace:              namespace,
+	})
 	if err != nil {
-		setupLog.Error(err, "")
+		setupLog.Error(err, "unable to start manager")
 		os.Exit(1)
 	}
 
-	setupLog.Info("Registering Components.")
+	mgr.GetClient()
```
xcoulon, line 123: is this call needed? it does not retain the returned `client.Client`

main.go
```diff
@@ -294,119 +248,41 @@ func main() {
...
-	}
-
-	// Add to the below struct any other metrics ports you want to expose.
-	servicePorts := []v1.ServicePort{
-		{Port: metricsPort, Name: metrics.OperatorPortName, Protocol: v1.ProtocolTCP, TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: metricsPort}},
-		{Port: operatorMetricsPort, Name: metrics.CRPortName, Protocol: v1.ProtocolTCP, TargetPort: intstr.IntOrString{Type: intstr.Int, IntVal: operatorMetricsPort}},
-	}
-
-	// Create Service object to expose the metrics port(s).
-	service, err := metrics.CreateMetricsService(ctx, cfg, servicePorts)
```
xcoulon, line 364: will the metric `service` and `servicemonitor` resources be created by the call to `ctrl.NewManager`?

controllers/memberstatus/memberstatus_controller_test.go
MatousJobanek: I'm really sorry, but I don't really understand what you meant by that question...? I'm somehow slow today morning...

controllers/idler/idler_controller.go
MatousJobanek: good catch - that slipped out of my notice - let me go over the code again and check if there are more of these

main.go
MatousJobanek: btw, you can check the resulting bundle manifests files in `bundle/` directory after running:
```
make bundle
```

controllers/idler/idler_controller.go
MatousJobanek: removed 114a7b9 :ghost:

main.go
MatousJobanek: whoops 
removed 114a7b9 :ghost: 
Thanks

controllers/memberstatus/memberstatus_controller_test.go
xcoulon: > I'm somehow slow today morning...

no worries ;) 
my point is: I believe that once we deploy the operator with these changes, we'll have a new deployment resource called `member-operator-controller-manager` and I wonder what will happen to the current `member-operator` one? are we going to have 2 sibling deployments?

main.go
xcoulon: ah ok, cool. Thanks for the clarification!

controllers/memberstatus/memberstatus_controller_test.go
MatousJobanek: Long story short: 
it's even more complicated and the easiest & cleanest way is to uninstall the operators and install them again when we merge the PRs

xcoulon: ok ;) I was just wondering how the transition would happen on the clusters. But that's also why we have a staging env 😜

MatousJobanek: The stage doesn't help us in this case much - both clusters use the same channel for updates. But I can do the manual update in stage first and then in prod, indeed.

make/manifests.mk
MatousJobanek: changed in d30cbfd

PR #268, opened 2021-06-22
build: use Codecov GitHub Action

xcoulon:
Paired with https://github.com/openshift/release/pull/19536

Fixes https://issues.redhat.com/browse/CRT-1033

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -10,9 +10,37 @@ on:
       - master
 
 jobs:
+  test-with-coverage:
+    name: Test with Coverage
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.14.x
+    
+    - name: Checkout code
+      uses: actions/checkout@v2
+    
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-build.yml
```diff
@@ -10,9 +10,37 @@ on:
...
+      with:
+        files: ./build/_output/coverage/coverage.txt
+        flags: unittests # optional
+        fail_ci_if_error: true # optional (default = false)
+        verbose: true # optional (default = false)
+    
   golangci:
     name: GolangCI Lint
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
MatousJobanek, line 43: this should be the same https://github.com/actions/virtual-environments#available-environments

MatousJobanek (approved):

.github/workflows/ci-build.yml
xcoulon: yeah, but I wanted to make sure we stick to an LTS version even if `latest` changes.

MatousJobanek: but do they use something else than LTS? It doesn't seem so...

PR #273, opened 2021-06-29
Migrate member config

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-1123

Related PR: https://github.com/codeready-toolchain/toolchain-e2e/pull/334

Migrates member operator configuration from config maps and env vars to the MemberOperatorConfig resource.

Summary of changes:
- The memberconfig controller now watches secrets
- Added a mapper for mapping secrets to the MemberOperatorConfig resource
- Added remaining configuration accessors to new configuration.go
- Removed old configuration.go
- Replaced all old configuration access to the new GetConfig way


controllers/memberoperatorconfig/cache.go
```diff
@@ -45,27 +49,50 @@ func loadLatest(cl client.Client, namespace string) error {
...
+	}
+
+	configCache.set(config, allSecrets)
 	return nil
 }
 
+func loadSecrets(cl client.Client) (map[string]map[string]string, error) {
+	var allSecrets = make(map[string]map[string]string)
+	secretList := &corev1.SecretList{}
+	err := cl.List(context.TODO(), secretList)
```
alexeykazakov, line 65: Shouldn't we list in specific namespace?

controllers/memberoperatorconfig/cache.go
```diff
@@ -45,27 +49,50 @@ func loadLatest(cl client.Client, namespace string) error {
...
+	if err != nil {
+		return allSecrets, err
+	}
+	for _, secret := range secretList.Items {
```
alexeykazakov, line 69: There is a bunch of built-in secrets exist in every namespace. They start with `default-`, `builder-`, `deployer-`.
And they all have an annotation: `kubernetes.io/service-account.name` (with different values though).
We should probably ignore these secrets.

controllers/memberoperatorconfig/mapper.go
```diff
@@ -0,0 +1,45 @@
...
+
+		if secret.GetNamespace() != controllerNS {
+			return []reconcile.Request{}
+		}
+
+		config := &toolchainv1alpha1.MemberOperatorConfig{}
+		if err := m.client.Get(context.TODO(), types.NamespacedName{Namespace: controllerNS, Name: "config"}, config); err != nil {
+			mapperLog.Error(err, "could not get MemberOperatorConfig resource", "name", "config", "namespace", controllerNS)
+			return []reconcile.Request{}
+		}
```
alexeykazakov, line 40: `Map()` should be a lightweight function without using clients/calls to API server. And you don't really use the loaded config anyway. So, just return the hardcoded `config` name as you already doing.
```suggestion
```

controllers/memberoperatorconfig/mapper_test.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+func TestSecretToMemberOperatorConfigMapper(t *testing.T) {
+	// when
+	secretData := map[string][]byte{
+		"che-admin-username": []byte("cheadmin"),
+		"che-admin-password": []byte("password"),
+	}
+	secret := newSecret("test-secret", secretData)
+	config := NewMemberOperatorConfigWithReset(t, testconfig.MemberStatus().RefreshPeriod("10s"))
```
alexeykazakov, line 27: As I mentioned above we don't need the config to be present for the mapper. So you can drop this.

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
```diff
@@ -26,7 +27,23 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	); err != nil {
+		return err
+	}
+
+	// Watch for changes to secrets that should trigger an update of the config cache
+	// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;watch;list
+	return c.Watch(
+		&source.Kind{Type: &corev1.Secret{}},
+		&handler.EnqueueRequestsFromMapFunc{
+			ToRequests: SecretToMemberOperatorConfigMapper{client: mgr.GetClient()},
```
alexeykazakov, line 43: And here too. You don't need a client here.

controllers/memberoperatorconfig/cache.go
rajivnathan: Yes, thanks for pointing it out. Done in c74fe7da69590d0d2861b41845039aaf0c19fffb

controllers/memberoperatorconfig/cache.go
rajivnathan: I went with the annotation method. Let me know if that should be good: b4189bcf653267d8a8678cd370c8e0764414c7fd

controllers/memberoperatorconfig/mapper.go
rajivnathan: Ah, didn't realize that. That makes sense. Done in cc1eb37ab554057e49588fa4b56fc1bbcda08a8f

controllers/memberoperatorconfig/mapper_test.go
rajivnathan: Removed it with the previous commit: cc1eb37ab554057e49588fa4b56fc1bbcda08a8f

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
rajivnathan: Removed it with the previous commit: cc1eb37ab554057e49588fa4b56fc1bbcda08a8f

controllers/memberoperatorconfig/mapper.go
```diff
@@ -1,43 +1,28 @@
...
 func (m SecretToMemberOperatorConfigMapper) Map(obj handler.MapObject) []reconcile.Request {
-	if secret, ok := obj.Object.(*corev1.Secret); ok {
+	if _, ok := obj.Object.(*corev1.Secret); ok {
 		controllerNS, err := k8sutil.GetWatchNamespace()
 		if err != nil {
 			mapperLog.Error(err, "could not determine watched namespace")
 			return []reconcile.Request{}
 		}
-
-		if secret.GetNamespace() != controllerNS {
```
alexeykazakov: Yeah.. dropping namespace check should be fine since this client watches only the member operator namespace. We have another client for the all-namespace watchers which is currently used in the idler and status controllers only.

alexeykazakov (approved):
Nice!

controllers/memberoperatorconfig/cache_test.go
```diff
@@ -104,12 +114,185 @@ func TestGetConfigFailed(t *testing.T) {
...
+		initconfig := NewMemberOperatorConfigWithReset(t, testconfig.MemberStatus().RefreshPeriod("1s"))
+		// given
+		cl := NewFakeClient(t, initconfig)
+
+		// when
+		err := loadLatest(cl, MemberOperatorNs)
+
+		// then
+		require.NoError(t, err)
+	})
```
MatousJobanek, line 128: it's probably not a big deal, but you don't verify the set/updated values in the cache. 
I know that you did it for the `updateConfig` function a few lines above, but that is a different function.

controllers/memberoperatorconfig/configuration.go
```diff
@@ -28,6 +145,52 @@ func (a MemberStatusConfig) RefreshPeriod() time.Duration {
...
+}
+
+func (a ToolchainClusterConfig) HealthCheckPeriod() time.Duration {
+	defaultClusterHealthCheckPeriod := "10s"
+	healthCheckPeriod := getString(a.t.HealthCheckPeriod, defaultClusterHealthCheckPeriod)
+	d, err := time.ParseDuration(healthCheckPeriod)
+	if err != nil {
+		d, _ = time.ParseDuration(defaultClusterHealthCheckPeriod)
+	}
+	return d
```
MatousJobanek, line 159: this could be extracted to a function as well:
```go
func getDuration(value *string, defaultValue string) time.Duration {
	healthCheckPeriod := getString(value, defaultValue)
	d, err := time.ParseDuration(healthCheckPeriod)
	if err != nil {
		d, _ = time.ParseDuration(defaultValue)
	}
	return d
}
```

controllers/memberoperatorconfig/configuration.go
```diff
@@ -28,6 +145,52 @@ func (a MemberStatusConfig) RefreshPeriod() time.Duration {
...
+	}
+	return defaultValue
+}
+
+func getInt(value *int, defaultValue int) int {
+	if value != nil {
+		return *value
+	}
+	return defaultValue
+}
```
MatousJobanek, line 192: It's not critical, but it could be moved to toolchain-common so it can be reused in other repos

controllers/memberoperatorconfig/mapper.go
```diff
@@ -0,0 +1,30 @@
...
+// Map maps secrets to the singular instance of MemberOperatorConfig named "config"
+func (m SecretToMemberOperatorConfigMapper) Map(obj handler.MapObject) []reconcile.Request {
+	if _, ok := obj.Object.(*corev1.Secret); ok {
+		controllerNS, err := k8sutil.GetWatchNamespace()
+		if err != nil {
+			mapperLog.Error(err, "could not determine watched namespace")
+			return []reconcile.Request{}
+		}
+		return []reconcile.Request{{NamespacedName: types.NamespacedName{Namespace: controllerNS, Name: "config"}}}
+	}
```
MatousJobanek, line 27: IIRC, we watch only in the operator namespace, so you can easily take the namespace from the secret object.
```suggestion
	if secret, ok := obj.Object.(*corev1.Secret); ok {
		return []reconcile.Request{{NamespacedName: types.NamespacedName{Namespace: secret.Namespace, Name: "config"}}}
	}
```

MatousJobanek: if you still want to use the Watch namespace env var - it would be better to store it as a parameter of the `SecretToMemberOperatorConfigMapper` type

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
```diff
@@ -67,6 +84,12 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
...
 	}
-	updateConfig(memberconfig)
+
+	allSecrets, err := loadSecrets(r.Client, request.Namespace)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	updateConfig(memberconfig, allSecrets)
 	return reconcile.Result{}, nil
```
MatousJobanek, line 94: you could replace the whole content of the `Reconcile` method with 
```
	return reconcile.Result{}, loadLatest(r.Client, request.Namespace)
```

controllers/memberstatus/memberstatus_controller.go
```diff
@@ -93,10 +92,13 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	reqLogger := r.Log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling MemberStatus")
 
-	config, err := cfg.GetConfig(r.Client, request.Namespace)
+	// retrieve the latest config and use it for this reconciliation
+	config, err := memberCfg.GetConfig(r.Client, request.Namespace)
 	if err != nil {
 		return reconcile.Result{}, errs.Wrapf(err, "unable to get MemberOperatorConfig")
 	}
+	r.config = config
```
MatousJobanek, line 100: It would be nicer to not store it as a parameter of the `Reconciler` type, but pass it as an argument to all methods/functions. In the same way, as we do with the `MemberStatus` object or with the logger.
After all, it's nothing static but we update it with every reconcile.
 WDYT?

controllers/useraccount/useraccount_controller.go
```diff
@@ -105,6 +105,13 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		}
 	}
 
+	// retrieve the latest config and use it for this reconciliation
+	config, err := memberCfg.GetConfig(r.Client, namespace)
+	if err != nil {
+		return reconcile.Result{}, errs.Wrapf(err, "unable to get MemberOperatorConfig")
+	}
+	r.config = config
```
MatousJobanek, line 113: I have the same concern here

MatousJobanek (commented):
Impressive work :+1: :1st_place_medal: 
I really like the API that is provided by the configuration type - the usage is very intuitive :+1: 

I left a few comments. The only significant concern that I have is the way the config is stored in the `Reconciler` types

controllers/memberoperatorconfig/configuration.go
rajivnathan: Put up a PR for this: https://github.com/codeready-toolchain/toolchain-common/pull/196

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
rajivnathan: Good point! I did it in f58a325fa6d1f8737001c7621f18db3ca060c54c

controllers/memberoperatorconfig/cache_test.go
rajivnathan: yeah, that's a fair point. GetConfig and loadLatest are used in different cases so it's worth verifying in both. Added tests for when the config has and hasn't changed in cabf09c52d96346671d13a4c55a314bc3e08f667

controllers/memberoperatorconfig/configuration.go
rajivnathan: I modified this slightly and added it to the toolchain-common PR: https://github.com/codeready-toolchain/toolchain-common/pull/196
```
func GetDuration(value *string, defaultValue time.Duration) time.Duration {
	durationAsString := GetString(value, "invalid value")
	d, err := time.ParseDuration(durationAsString)
	if err != nil {
		return defaultValue
	}
	return d
}
```

controllers/memberoperatorconfig/mapper.go
rajivnathan: You're right, we only watch the operator namespace so we can just take the namespace from the secret object.

controllers/memberstatus/memberstatus_controller.go
rajivnathan: Yeah, I agree with you. While working on the host operator changes I decided to pass it as an argument to functions OR if there was only one function that needed config then just call `GetConfig` from that function. I'll update this PR to do the same.

rajivnathan: Done in 71a0c3e74f849496bfac07f9c3573cef8a80e5e6

controllers/useraccount/useraccount_controller.go
rajivnathan: Done in 71a0c3e74f849496bfac07f9c3573cef8a80e5e6

controllers/memberoperatorconfig/configuration.go
MatousJobanek, line 192: Didn't you forget to update it with the functions from common? https://github.com/codeready-toolchain/toolchain-common/pull/196

make/manifests.mk
```diff
@@ -7,7 +7,7 @@ OWNER_AND_BRANCH_LOCATION=codeready-toolchain/api/master
 GH_SCRIPTS_URL=https://raw.githubusercontent.com/${OWNER_AND_BRANCH_LOCATION}/scripts
 
 TMP_DIR?=/tmp
-IMAGE_BUILDER?=podman
+IMAGE_BUILDER?=docker
```
MatousJobanek, line 10: if we decided to use docker as the default image builder, then we should do it in all repos (and ideally in separate PRs)

MatousJobanek (approved):
This is a huge improvement and very nice implementation  :+1: :1st_place_medal:

controllers/memberoperatorconfig/configuration.go
rajivnathan: Yes, thanks for reminding me.

make/manifests.mk
rajivnathan: Sure, let me undo it from this PR.

controllers/memberoperatorconfig/configuration.go
rajivnathan: Done in 1a217f60b6ea7f2ba3ed0a389b0b544eadc26c39

make/manifests.mk
rajivnathan: Reverted in 278989c3c667f9ce261e598666d73f167014cf5e

PR #275, opened 2021-07-01
fix: don't append path if contains only slash

MatousJobanek:
don't append path in the URL if contains only slash

### DIFF
--- a/pkg/che/che_test.go
+++ b/pkg/che/che_test.go
@@ -584,7 +584,7 @@ func cheRoute(tls bool) *routev1.Route { //nolint: unparam
 		},
 		Spec: routev1.RouteSpec{
 			Host: fmt.Sprintf("codeready-codeready-workspaces-operator.%s", test.MemberClusterName),
-			Path: "",
+			Path: "/",
 		},
 	}
 	if tls {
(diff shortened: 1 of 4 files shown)


pkg/che/che_test.go
```diff
@@ -584,7 +584,7 @@ func cheRoute(tls bool) *routev1.Route { //nolint: unparam
 		},
 		Spec: routev1.RouteSpec{
 			Host: fmt.Sprintf("codeready-codeready-workspaces-operator.%s", test.MemberClusterName),
-			Path: "",
+			Path: "/",
```
alexeykazakov, line 587: Can we test both cases?
Just in case the / will disappear again in the future.

Alternatively we could test the route until directly if it's easier to test.

alexeykazakov (approved):
Thanks for the great investigation of the problem!

pkg/che/tokencache_test.go
```diff
@@ -359,7 +359,7 @@ func keycloackRoute(tls bool) *routev1.Route { //nolint: unparam
 		},
 		Spec: routev1.RouteSpec{
 			Host: fmt.Sprintf("keycloak-codeready-workspaces-operator.%s", test.MemberClusterName),
-			Path: "",
+			Path: "/",
```
ibuziuk, line 362: the test should work without this change, right?
might be a good idea to have the test when path is not defined

pkg/che/che_test.go
ibuziuk: yes, would be great to have both cases covered (with & without path)

MatousJobanek:
@alexeykazakov @ibuziuk I added better tests and also modified the check of the path - it should be smarter now: 3f98d30

ibuziuk (approved):

alexeykazakov (approved):
Looks great. Thanks!

PR #277, opened 2021-07-05
feat: tolerate one missing NodeMetrics resource

MatousJobanek:
When there is a cluster update/upgrade, it may happen that one NodeMetrics resource is missing - we can tolerate that. It should report a problem only when there is more than one NodeMetrics missing.


controllers/memberstatus/memberstatus_controller_test.go
```diff
@@ -325,6 +325,25 @@ func TestOverallStatusCondition(t *testing.T) {
 			HasRoutes("https://console.member-cluster/console/", "https://codeready-codeready-workspaces-operator.member-cluster/che/", routesAvailable())
 	})
 
+	t.Run("when missing only one NodeMetrics resource then it's fine", func(t *testing.T) {
+		// given
+		memberOperatorDeployment := newMemberDeploymentWithConditions(status.DeploymentAvailableCondition(), status.DeploymentProgressingCondition())
+
+		unraedyNodeAndMetrics := newNodesAndNodeMetrics(forNode("worker-unready", []string{"worker"}, "3000000Ki"))
```
xcoulon, line 332:
```suggestion
		unreadyNodeAndMetrics := newNodesAndNodeMetrics(forNode("worker-unready", []string{"worker"}, "3000000Ki"))
```

controllers/memberstatus/memberstatus_controller_test.go
```diff
@@ -325,6 +325,25 @@ func TestOverallStatusCondition(t *testing.T) {
...
+		reconciler, req, fakeClient := prepareReconcile(t, defaultMemberStatusName, newGetHostClusterReady, allNamespacesCl, append(nodeAndMetrics, unraedyNodeAndMetrics[0], memberOperatorDeployment, newMemberStatus())...)
```
xcoulon, line 333:
```suggestion
		reconciler, req, fakeClient := prepareReconcile(t, defaultMemberStatusName, newGetHostClusterReady, allNamespacesCl, append(nodeAndMetrics, unreadyNodeAndMetrics[0], memberOperatorDeployment, newMemberStatus())...)
```

xcoulon: also, why do you need to use `append()`? Can't you just do something like 
```
reconciler, req, fakeClient := prepareReconcile(t, defaultMemberStatusName, newGetHostClusterReady, allNamespacesCl, nodeAndMetrics, unreadyNodeAndMetrics[0], memberOperatorDeployment, newMemberStatus())
```
as in the next test?

xcoulon: also, could you add comment to explain why the node is "unready"? It's not clear to me at first sight, because I'm not familiar with the logic, so a comment would be helpful.

controllers/memberstatus/memberstatus_controller_test.go
```diff
@@ -393,10 +412,11 @@ func TestOverallStatusCondition(t *testing.T) {
 				HasRoutes("https://console.member-cluster/console/", "https://codeready-codeready-workspaces-operator.member-cluster/che/", routesAvailable())
 		})
 
-		t.Run("when missing NodeMetrics for Node", func(t *testing.T) {
+		t.Run("when missing NodeMetrics for two Nodes", func(t *testing.T) {
 			// given
-			nodeAndMetrics := newNodesAndNodeMetrics(forNode("worker-123", []string{"worker"}, "3000000Ki"))
-			reconciler, req, fakeClient := prepareReconcile(t, requestName, getHostClusterFunc, allNamespacesCl, nodeAndMetrics[0], memberOperatorDeployment, memberStatus)
+			nodeAndMetrics1 := newNodesAndNodeMetrics(forNode("worker-123", []string{"worker"}, "3000000Ki"))
```
xcoulon, line 417: Could we rename those vars to something like `unreadyNodeAndMetrics1` and `unreadyNodeAndMetrics2` and add a comment (as mentioned above)

controllers/memberstatus/memberstatus_controller_test.go
MatousJobanek: No, I cannot do that, because the `nodeAndMetrics` is  an array and the `prepareReconcile` is a variadic function

xcoulon: ok, thanks, I missed that point :/

MatousJobanek:
@xcoulon thanks for the review :pray: 
I renamed the variables and added comments to make it more clear: 7597473

xcoulon (approved):
great! thanks for taking my comments into account :D

alexeykazakov (approved):

PR #281, opened 2021-07-20
Move autoscaler and webhook

rajivnathan:
- Moves the autoscaler and webhook logic from main.go to the member operator config controller.
- Also updated the cache to use common functions from toolchain-common
- Updated the cache's loadLatest func to also return the configuration so that it's more convenient to use from the controller reconcile function

Related issue: https://issues.redhat.com/browse/CRT-1177

### DIFF
--- a/controllers/memberoperatorconfig/cache.go
+++ b/controllers/memberoperatorconfig/cache.go
@@ -27,13 +27,13 @@ func (c *cache) set(config *toolchainv1alpha1.MemberOperatorConfig, secrets map[
 	c.Lock()
 	defer c.Unlock()
 	c.config = config.DeepCopy()
-	c.secrets = copyOf(secrets)
+	c.secrets = common.CopyOf(secrets)
 }
 
 func (c *cache) get() (*toolchainv1alpha1.MemberOperatorConfig, map[string]map[string]string) {
 	c.RLock()
 	defer c.RUnlock()
-	return c.config.DeepCopy(), copyOf(c.secrets)
+	return c.config.DeepCopy(), common.CopyOf(c.secrets)
 }
 
 func updateConfig(config *toolchainv1alpha1.MemberOperatorConfig, secrets map[string]map[string]string) {
@@ -45,18 +45,18 @@ func loadLatest(cl client.Client, namespace string) (Configuration, error) {
 	if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: "config"}, config); err != nil {
 		if apierrors.IsNotFound(err) {
 			cacheLog.Info("MemberOperatorConfig resource with the name 'config' wasn't found, default configuration will be used", "namespace", namespace)
-			return Configuration{}, nil
+			return Configuration{m: &toolchainv1alpha1.MemberOperatorConfigSpec{}}, nil
 		}
-		return Configuration{}, err
+		return Configuration{m: &toolchainv1alpha1.MemberOperatorConfigSpec{}}, err
 	}
 
 	allSecrets, err := common.LoadSecrets(cl, namespace)
 	if err != nil {
-		return Configuration{}, err
+		return Configuration{m: &toolchainv1alpha1.MemberOperatorConfigSpec{}}, err
 	}
 
 	configCache.set(config, allSecrets)
-	return getConfig(), nil
+	return getConfigOrDefault(), nil
 }
 
 // GetConfig returns a cached memberoperator config.
@@ -68,10 +68,10 @@ func GetConfig(cl client.Client, namespace string) (Configuration, error) {
 	if config == nil {
 		return loadLatest(cl, namespace)
 	}
-	return getConfig(), nil
+	return getConfigOrDefault(), nil
 }
 
-func getConfig() Configuration {
+func getConfigOrDefault() Configuration {
 	config, secrets := configCache.get()
 	if config == nil {
 		return Configuration{m: &toolchainv1alpha1.MemberOperatorConfigSpec{}, secrets: secrets}
@@ -84,16 +84,4 @@ func getConfig() Configuration {
 // then the function has to be exported and placed here.
 func Reset() {
 	configCache = &cache{}
-}
-
-func copyOf(originalMap map[string]map[string]string) map[string]map[string]string {
-	targetMap := make(map[string]map[string]string, len(originalMap))
-	for key, value := range originalMap {
-		secretData := make(map[string]string, len(value))
-		for k, v := range value {
-			secretData[k] = v
-		}
-		targetMap[key] = secretData
-	}
-	return targetMap
 }
--- a/controllers/memberoperatorconfig/cache.go
+++ b/controllers/memberoperatorconfig/cache.go
@@ -40,23 +40,23 @@ func updateConfig(config *toolchainv1alpha1.MemberOperatorConfig, secrets map[st
 	configCache.set(config, secrets)
 }
 
-func loadLatest(cl client.Client, namespace string) error {
+func loadLatest(cl client.Client, namespace string) (Configuration, error) {
 	config := &toolchainv1alpha1.MemberOperatorConfig{}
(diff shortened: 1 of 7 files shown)


controllers/memberoperatorconfig/cache.go
```diff
@@ -28,59 +28,61 @@ func (c *cache) set(config *toolchainv1alpha1.MemberOperatorConfig, secrets map[
...
+		return loadLatest(cl)
 	}
+	return getConfigOrDefault(), nil
+}
+
+func getConfigOrDefault() Configuration {
+	config, secrets := configCache.get()
 	if config == nil {
-		return Configuration{m: &toolchainv1alpha1.MemberOperatorConfigSpec{}, secrets: secrets}, nil
+		return Configuration{m: &toolchainv1alpha1.MemberOperatorConfigSpec{}, secrets: secrets}
```
xcoulon, line 83: I'm a bit confused with return a default config with an empty `MemberOperatorConfigSpec` yet existing `secrets`. Could you explain when it's needed?

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
```diff
@@ -48,5 +51,58 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (r *Reconciler) handleAutoscalerDeploy(logger logr.Logger, cfg Configuration, namespace string) error {
+	if cfg.Autoscaler().Deploy() {
+		logger.Info("(Re)Deploying autoscaling buffer")
+		if err := autoscaler.Deploy(r.Client, r.Client.Scheme(), namespace, cfg.Autoscaler().BufferMemory(), cfg.Autoscaler().BufferReplicas()); err != nil {
+			logger.Error(err, "failed to deploy autoscaling buffer")
```
xcoulon, line 74: I think you should just wrap the error with `"failed to deploy autoscaling buffer"` and return it, and let the caller log the whole error

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
```diff
@@ -48,5 +51,58 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+			return err
+		}
+		logger.Info("(Re)Deployed autoscaling buffer")
+	} else {
+		deleted, err := autoscaler.Delete(r.Client, r.Client.Scheme(), namespace)
+		if err != nil {
+			logger.Error(err, "failed to delete autoscaling buffer")
```
xcoulon, line 81: same here

controllers/memberoperatorconfig/cache.go
rajivnathan: It shouldn't ever be needed right now but I think it's pretty harmless to include it in case at some point we want our default config to reference a secret value.

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
rajivnathan: I initially did that but the autoscaler errors are already wrapped so I felt it was kind of unnecessary to wrap it again: https://github.com/codeready-toolchain/member-operator/blob/master/pkg/autoscaler/autoscaler.go#L30

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
rajivnathan: same here: https://github.com/codeready-toolchain/member-operator/blob/32151a402aee38c72c18267c90e1e59abdbd2145/pkg/autoscaler/autoscaler.go#L54

controllers/memberoperatorconfig/cache.go
```diff
@@ -28,59 +28,61 @@ func (c *cache) set(config *toolchainv1alpha1.MemberOperatorConfig, secrets map[
...
-		return errs.Wrap(err, "Failed to get watch namespace")
+		return Configuration{m: &toolchainv1alpha1.MemberOperatorConfigSpec{}}, errs.Wrap(err, "Failed to get watch namespace")
 	}
 
 	config := &toolchainv1alpha1.MemberOperatorConfig{}
 	if err := cl.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: "config"}, config); err != nil {
 		if apierrors.IsNotFound(err) {
 			cacheLog.Info("MemberOperatorConfig resource with the name 'config' wasn't found, default configuration will be used", "namespace", namespace)
-			return nil
+			return Configuration{m: &toolchainv1alpha1.MemberOperatorConfigSpec{}}, nil
```
MatousJobanek, line 54: why don't we call `getConfigOrDefault()` here as well?
```suggestion
			return getConfigOrDefault(), nil
```
so if the config is (accidentally) removed, we could still used the cached one

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
```diff
@@ -48,5 +51,58 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	crtConfig, err := loadLatest(r.Client)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if err := r.handleAutoscalerDeploy(reqLogger, crtConfig, request.Namespace); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	if err := r.handleUserPodsWebhookDeploy(reqLogger, crtConfig, request.Namespace); err != nil {
```
MatousJobanek, line 63: Correct me if I'm wrong - to follow the best practices and also to properly manage the deployments, the controller should also watch the resources that are created/deleted here, right?

rajivnathan: Right. As discussed let's handle that in an upcoming PR for https://issues.redhat.com/browse/CRT-1194

controllers/memberoperatorconfig/cache.go
rajivnathan: That's a good point. It's more likely that we accidentally remove the config than remove it and want the default configuration to take over so I'll make that change.

controllers/memberoperatorconfig/cache.go
xcoulon: ok

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
xcoulon: but in that case, we don't need to log the error here, since it should already be logged by the caller, shouldn't it?

rajivnathan: Looking at https://github.com/codeready-toolchain/member-operator/blob/master/pkg/autoscaler/autoscaler.go#L30 it doesn't look like it's logged. The error is just wrapped and returned.

controllers/memberoperatorconfig/cache.go
rajivnathan: Done. Updated the code as well as the message and the unit tests. b6313c747449961f0ab297c81f3cd687a35bd6b4 and ef6402100b4f3380cf0fa5ab7694d672f8f031ee

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
xcoulon: well, I assume it's logged by the reconciler? or am I missing something?

xcoulon: but we can keep it like that for now and see how it goes when an error occurs.

xcoulon (approved):

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
rajivnathan: Removed the logger.Error calls as discussed: 728a4a3987b91c5ff92c57392dd1c8147a025335

PR #283, opened 2021-07-27
remove downloading and installing of golang and oc from dockerfile

ranakan19:
oc and golang are already available as part of test configuration. Thus removing the downloading and installing of them from dockerfile.


MatousJobanek (approved):
Thanks

PR #286, opened 2021-08-04
fix: requeue after 1 second when deleting namespace

MatousJobanek:
the controller is maybe not triggered when the Namespace is finally removed - let's requeue after 1 second every time when it's deleting a namespace.


alexeykazakov (approved):

alexeykazakov:
This time it's the idler test which failed. We need to fix it too but it's for another PR of course :)

PR #295, opened 2021-09-02
Use configuration cache from toolchain-common

rajivnathan:
Summary of changes:
- Updates to use the config cache from toolchain-common
- Removes the now redundant config cache in this repo


alexeykazakov (approved):

rajivnathan:
/retest

test failed with an etcdserver error:
```
    member.go:500: waiting for NetworkPolicy 'allow-same-namespace' in namespace 'cheesecakelover05-dev'
    member.go:500: waiting for NetworkPolicy 'allow-from-openshift-monitoring' in namespace 'cheesecakelover05-dev'
    member.go:500: waiting for NetworkPolicy 'allow-from-openshift-ingress' in namespace 'cheesecakelover05-dev'
    member.go:500: waiting for NetworkPolicy 'allow-from-codeready-workspaces-operator' in namespace 'cheesecakelover05-dev'
    member.go:518: failed to wait for NetworkPolicy 'allow-from-codeready-workspaces-operator' in namespace 'cheesecakelover05-dev'
    checks.go:421: 
        	Error Trace:	checks.go:421
        	            				nstemplateset.go:40
        	            				nstemplatetier_test.go:301
        	            				nstemplatetier_test.go:167
        	Error:      	Received unexpected error:
        	            	etcdserver: leader changed
        	Test:       	TestUpdateNSTemplateTier
```
Full logs: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/codeready-toolchain_member-operator/295/pull-ci-codeready-toolchain-member-operator-master-e2e/1433582883418673152/build-log.txt

controllers/memberoperatorconfig/configuration.go
```diff
@@ -17,67 +20,111 @@ const (
...
+		return Configuration{cfg: &toolchainv1alpha1.MemberOperatorConfigSpec{}}, err
+	}
+	return newConfiguration(config, secrets), nil
+}
+
+// GetCachedConfiguration returns a Configuration directly from the cache
+func GetCachedConfiguration() Configuration {
+	config, secrets := commonconfig.GetCachedConfig()
+	return newConfiguration(config, secrets)
+}
```
MatousJobanek, line 43: As in the other repos, I would really prefer having and using only one function.

MatousJobanek (approved):
Looks good - it's great that we can reuse the same cache logic in all repos :+1: :1st_place_medal: 

While doing the review I noticed that `memberoperatorconfig_controller_test.go` could fail with panic if the error is nil: https://github.com/codeready-toolchain/member-operator/blob/42be460a035e43ba9ed181de1be5f0cd2124618b/controllers/memberoperatorconfig/memberoperatorconfig_controller_test.go#L185-L188
you could replace it with: `require.EqualError`

rajivnathan:
@MatousJobanek Good catch about the potential panic in the test but I used `Contains` because the error message is pretty long `"cannot deploy autoscaling buffer template: unable to create resource of kind: PriorityClass, version: v1: unable to get the resource '&{map[apiVersion:scheduling.k8s.io/v1 description:This priority class is to be used by the autoscaling buffer pod only globalDefault:false kind:PriorityClass metadata:map[labels:map[toolchain.dev.openshift.com/provider:codeready-toolchain] name:member-operator-autoscaling-buffer] value:-5]}': client error"`

So rather than using `require.EqualError` I just added a `require.NotNil(t, err)` just before it: https://github.com/codeready-toolchain/member-operator/pull/295/commits/8dfcf63f99d3eba700d152282b465c3b1c1586dc

controllers/memberoperatorconfig/configuration.go
rajivnathan: Agreed, I did notice that and planned to change it but I was planning to do it in a separate PR if that's alright with you.

MatousJobanek: ok, I'm fine with it :+1:

MatousJobanek:
> So rather than using require.EqualError I just added a require.NotNil(t, err) just before it: 8dfcf63

yeah, sure - that makes sense :-)

PR #304, opened 2021-10-21
chore: update 'make lint' target

xcoulon:
trigger install of v1.42.1 and uses `.golangci.yml` config file

### DIFF
--- a/make/lint.mk
+++ b/make/lint.mk
@@ -13,8 +13,11 @@ endif
 
 .PHONY: lint-go-code
 ## Checks the code with golangci-lint
-lint-go-code:
-ifeq (, $(shell which golangci-lint 2>/dev/null))
-	$(error "golangci-lint not found in PATH. Please install it using instructions on https://golangci-lint.run/usage/install/#local-installation")
-endif
-	golangci-lint ${V_FLAG} run -E gofmt,golint,megacheck,misspell
+lint-go-code: install-golangci-lint
+	# run with the same flags as on GitHub Actions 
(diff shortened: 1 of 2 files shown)


make/lint.mk
```diff
@@ -13,8 +13,11 @@ endif
...
-	golangci-lint ${V_FLAG} run -E gofmt,golint,megacheck,misspell
\ No newline at end of file
+lint-go-code: install-golangci-lint
+	# run with the same flags as on GitHub Actions 
+	$(Q)${GOPATH}/bin/golangci-lint ${V_FLAG} run --config=./.golangci.yml --verbose ./...
+
+.PHONY: install-golangci-lint
+## Install development tools.
+install-golangci-lint:
+	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(shell go env GOPATH)/bin v1.42.1
```
MatousJobanek, line 23: Correct me if I'm wrong, but didn't we agree on not installing any binaries automatically from a link in makefile targets?

alexeykazakov: I still think we should not execute any binaries we don't 100% trust in make targets **if they are used in CI or CD**. For that we should use trusted GitHub Actions.
But I also agree that it's much less critical to install/execute 3rd party binaries in make targets we use locally only but still... Installing the linter is one time task. You don't have to do it every time you run the make target. I think our current approach of detecting that it's not installed and pointing to the documentation with clear instructions how to install it is a good practice.

xcoulon:
replaced by https://github.com/codeready-toolchain/member-operator/pull/328

PR #307, opened 2021-11-02
fix: use AllNamespacesClient to list NodeMetrics

xcoulon:
Also, disable watching the NodeMetrics in this client,
so we don't get this error every second:

```
Failed to watch *v1beta1.NodeMetrics: the server does not allow this method on the requested resource (get nodes.metrics.k8s.io)
```

### DIFF
--- a/controllers/memberstatus/memberstatus_controller_test.go
+++ b/controllers/memberstatus/memberstatus_controller_test.go
@@ -98,14 +98,14 @@ func TestOverallStatusCondition(t *testing.T) {
 
 	restore := test.SetEnvVarsAndRestore(t, test.Env(commonconfig.OperatorNameEnvVar, defaultMemberOperatorName))
 	defer restore()
-	nodes, metrics := newNodesAndNodeMetrics(
+	nodeAndMetrics := newNodesAndNodeMetrics(
 		forNode("worker-123", []string{"worker"}, "4000000Ki", withMemoryUsage("1250000Ki")),
 		forNode("worker-345", []string{"worker"}, "6000000Ki", withMemoryUsage("2250000Ki")),
 		forNode("worker-567", []string{"worker"}, "6000000Ki", withMemoryUsage("500000Ki")),
 		forNode("master-123", []string{"master"}, "4000000Ki", withMemoryUsage("2000000Ki")),
 		forNode("master-456", []string{"master"}, "5000000Ki", withMemoryUsage("1000000Ki")))
-	allNamespacesCl := test.NewFakeClient(t, append(metrics, consoleRoute(), cheRoute(true))...)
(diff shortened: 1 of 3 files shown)


alexeykazakov (approved):
> Also, disable watching the NodeMetrics in this client,

I guess you mean disabling caching.

xcoulon:
> > Also, disable watching the NodeMetrics in this client,
> 
> I guess you mean disabling caching.

yes, I meant that indeed. Which should disable caching as well.

controllers/memberstatus/memberstatus_controller_test.go
```diff
@@ -98,22 +98,22 @@ func TestOverallStatusCondition(t *testing.T) {
 
 	restore := test.SetEnvVarsAndRestore(t, test.Env(commonconfig.OperatorNameEnvVar, defaultMemberOperatorName))
 	defer restore()
-	nodeAndMetrics := newNodesAndNodeMetrics(
+	nodes, metrics := newNodesAndNodeMetrics(
```
MatousJobanek, line 101: what is the reasoning behind splitting these two values?

MatousJobanek (commented):
I understand why disabling the cache for the `NodeMetrics` kind, but why do we need to use AllNamespaces for that?

controllers/memberstatus/memberstatus_controller_test.go
xcoulon: because nodes are obtained by the "default" client and metrics are obtained by the `AllNamespaces` client

xcoulon:
> I understand why disabling the cache for the `NodeMetrics` kind, but why do we need to use AllNamespaces client for that?

because NodeMetrics are cluster-scoped resources and we should use a client that is not namespaced-scoped for such resources.

xcoulon:
> > I understand why disabling the cache for the `NodeMetrics` kind, but why do we need to use AllNamespaces client for that?
> 
> because NodeMetrics are cluster-scoped resources and we should use a client that is not namespaced-scoped for such resources.

ah wait, I misread the docs. The default client should do the trick 🤦‍♂️

xcoulon:
> > > I understand why disabling the cache for the `NodeMetrics` kind, but why do we need to use AllNamespaces client for that?
> > 
> > 
> > because NodeMetrics are cluster-scoped resources and we should use a client that is not namespaced-scoped for such resources.
> 
> ah wait, I misread the docs. The default client should do the trick 🤦‍♂️

@MatousJobanek ok, so now the PR is *much* simpler: just configuring the "default" client with an option to disable caching of NodeMetrics!

MatousJobanek (approved):
Thanks for checking it one more time and for simplifying the change :+1:

xcoulon:
> Thanks for checking it one more time and for simplifying the change 👍

well, thanks to you too ;)

PR #308, opened 2021-11-04
Add member operator namespace to the tier params

alexeykazakov:
This PR sets a new param `MEMBER_OPERATOR_NAMESPACE` when applying templates. This param will be used by the appstudio tier to setup a role binding for the service account.

### DIFF
--- a/controllers/nstemplateset/nstemplatetier.go
+++ b/controllers/nstemplateset/nstemplatetier.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/codeready-toolchain/toolchain-common/pkg/configuration"
+
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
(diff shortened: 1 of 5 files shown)


controllers/nstemplateset/nstemplatetier.go
```diff
@@ -76,8 +78,15 @@ type tierTemplate struct {
 // process processes the template inside of the tierTemplate object and replaces the USERNAME variable with the given username.
 // Optionally, it also filters the result to return a subset of the template objects.
 func (t *tierTemplate) process(scheme *runtime.Scheme, username string, filters ...template.FilterFunc) ([]runtimeclient.Object, error) {
+	ns, err := configuration.GetWatchNamespace()
+	if err != nil {
+		return nil, err
+	}
```
MatousJobanek, line 84: This could be moved to `getTierTemplate` function and store the namespace as part of `tierTemplate`
https://github.com/codeready-toolchain/member-operator/blob/master/controllers/nstemplateset/nstemplatetier.go#L24-L44
but it's not a big deal - your version is fine & valid too.

MatousJobanek (approved):
:+1:

PR #316, opened 2021-12-07
Prevent rolebinding creations with validating webhook

ranakan19:
Prevent rolebinding creation giving access to all service accounts and all users by using a validating webhook. It checks if the user creating the rolebinding is not a system user or service account, but is a normal user.
fixes: [CRT1148](https://issues.redhat.com/browse/CRT-1148)
e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/503


deploy/webhook/member-operator-webhook.yaml
```diff
@@ -95,6 +95,38 @@ objects:
...
         toolchain.dev.openshift.com/provider: codeready-toolchain
+- apiVersion: admissionregistration.k8s.io/v1
+  kind: ValidatingWebhookConfiguration
+  metadata:
+    name: member-operator-validating-webhook
+    labels:
+      app: member-operator-webhook
+      toolchain.dev.openshift.com/provider: codeready-toolchain
+  webhooks:
+    - name: users.rolebindings.webhook.sandbox
```
xcoulon, line 106:
```suggestion
    - name: users.rolebindings.validating.webhook
```

pkg/webhook/validatingwebhook/validate.go
```diff
@@ -0,0 +1,119 @@
...
+		APIGroup: "rbac.authorization.k8s.io",
+	}
+	allUsersSubject := rbac.Subject{
+		Kind:     "Group",
+		Name:     "system:authenticated",
+		APIGroup: "rbac.authorization.k8s.io",
+	}
+
+	//check the requesting user is not a system user
+	if !strings.Contains(requestingUser.Username, "system:") {
```
xcoulon, line 79: @MatousJobanek isn't it going to be problematic with App Studio when multiple users are given permissions on Space's namespaces?

xcoulon: also, you could probably use somethings like `strings.HasPrefix` instead of `strings.Contains`

deploy/webhook/member-operator-webhook.yaml
alexeykazakov: @xcoulon I would keep sandbox specific name here. Also `users.rolebindings.webhook.sandbox` is consistent with `users.pods.webhook.sandbox` name we use for our user pod webhook.

deploy/webhook/member-operator-webhook.yaml
```diff
@@ -95,6 +95,38 @@ objects:
...
+      clientConfig:
+        caBundle: ${CA_BUNDLE}
+        service:
+          name: member-operator-webhook
+          namespace: ${NAMESPACE}
+          path: "/validate-users-rolebindings"
+          port: 443
+      matchPolicy: Equivalent
+      rules:
+        - operations: ["CREATE", "UPDATE"]
```
alexeykazakov, line 118: We don't really need to validate updates, do we? Create should be enough, right?

pkg/webhook/validatingwebhook/validate.go
alexeykazakov: Just checking if the username starts with `system:` won't work. Not because of multiple users per space but because SA account names start with `system:`. But our goal is not to block all users but service accounts. Our goal is to prevent Sandbox users from creating such role bindings. Other regular users should be fine (dedicated admin users, crt users, other potential sre users, etc). Such users don't have the `system:` prefix.
So, it seems that we have to get the corresponding OpenShift user object and check the labels to determine if the user was created / is managed by Sandbox. If so then reject.
Any other ideas how we could do it differently?
Sandbox users still would be able to create role bindings if they use a service account but IMO that's fine. Our main goal is to prevent accidental role binding creation.

ranakan19: @xcoulon using HasPrefix now

pkg/webhook/validatingwebhook/validate.go
```diff
@@ -0,0 +1,119 @@
...
+	if _, err := w.Write(respBody); err != nil {
+		log.Error(err, "unable to write response")
+	}
+}
+
+func validate(body []byte) []byte {
+	admReview := v1.AdmissionReview{}
+	if _, _, err := deserializer.Decode(body, nil, &admReview); err != nil {
+		log.Error(err, "unable to deserialize the admission review object", "body", body)
+		admReview.Response = responseWithError(admReview.Request.UID, err)
```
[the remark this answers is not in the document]
ranakan19: done

pkg/webhook/validatingwebhook/validate.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	// let's unmarshal the object to be sure that it's a rolebinding
+	var rb *rbac.RoleBinding
+	if err := json.Unmarshal(admReview.Request.Object.Raw, &rb); err != nil {
+		log.Error(err, "unable unmarshal rolebinding json object", "AdmissionReview", admReview)
+		admReview.Response = responseWithError(admReview.Request.UID, errors.Wrapf(err, "unable unmarshal rolebinding json object - raw request object: %v", admReview.Request.Object.Raw))
```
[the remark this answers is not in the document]
ranakan19: done

pkg/webhook/validatingwebhook/validate.go
```diff
@@ -0,0 +1,150 @@
...
+		APIGroup: "rbac.authorization.k8s.io",
+	}
+	requestingUser := &userv1.User{}
+	err := client.Get(context.TODO(), types.NamespacedName{
+		Name: admReview.Request.UserInfo.Username,
+	}, requestingUser)
+
+	if err != nil {
+		log.Error(fmt.Errorf("Cannot find the user: %w", err), "unable to find the user requesting creation")
+		return []byte(fmt.Sprintf("unable to find the user requesting creation: %s", err))
```
alexeykazakov, line 94: Shouldn't we return an admin review here instead of a plain text message?
And as we discussed it let's just log this as a warning/error and return a successful admin review. We should allow requests from unknown users (like service accounts?).

Also, please make sure you cover all the cases in the e2e tests:
We need all three types of "users" trying to create all types of the role bindings:
- SA
- User managed by sandbox
- User not managed by sanbox
And the role bindings types:
- all SAs
- all Users
- something else

pkg/webhook/validatingwebhook/validate.go
```diff
@@ -0,0 +1,150 @@
...
+		Name:     "system:serviceaccounts",
+		APIGroup: "rbac.authorization.k8s.io",
+	}
+	allUsersSubject := rbac.Subject{
+		Kind:     "Group",
+		Name:     "system:authenticated",
+		APIGroup: "rbac.authorization.k8s.io",
+	}
+	requestingUser := &userv1.User{}
+	err := client.Get(context.TODO(), types.NamespacedName{
```
alexeykazakov, line 88: Check if the subject first before obtaining the user. We do not need to pull the user if the subject is not all-service-accounts or all-users.

pkg/webhook/validatingwebhook/validate.go
```diff
@@ -0,0 +1,150 @@
...
+		Name: admReview.Request.UserInfo.Username,
+	}, requestingUser)
+
+	if err != nil {
+		log.Error(fmt.Errorf("Cannot find the user: %w", err), "unable to find the user requesting creation")
+		return []byte(fmt.Sprintf("unable to find the user requesting creation: %s", err))
+	}
+	//check if the requesting user is a sandbox user
+	if !strings.HasPrefix(requestingUsername, "system:") && requestingUser.GetLabels()[toolchainv1alpha1.ProviderLabelKey] == toolchainv1alpha1.ProviderLabelValue {
```
alexeykazakov, line 97: This is another check we can do in the very beggining. If the username starts with "system:", so it's a SA then we just allow it without checking anything else.

pkg/webhook/validatingwebhook/validate.go
```diff
@@ -0,0 +1,150 @@
...
+		for _, sub := range subjects {
+			if sub == allUsersSubject || sub == allServiceAccountsSubject {
+				log.Error(fmt.Errorf("trying to give access which is restricted"), "unable unmarshal rolebinding json object", "AdmissionReview", admReview)
```
alexeykazakov, line 100: It's not an error.

ranakan19: done in ac5d239

pkg/webhook/validatingwebhook/validate.go
ranakan19: done in ac5d239

pkg/webhook/validatingwebhook/validate.go
ranakan19: done in ac5d239

pkg/webhook/validatingwebhook/validate.go
```diff
@@ -0,0 +1,144 @@
...
+	for _, sub := range subjects {
+		if sub == allUsersSubject || sub == allServiceAccountsSubject {
+			requestingUser := &userv1.User{}
+			err := client.Get(context.TODO(), types.NamespacedName{
+				Name: admReview.Request.UserInfo.Username,
+			}, requestingUser)
+
+			if err != nil {
+				log.Error(fmt.Errorf("Cannot find the user: %w", err), "unable to find the user requesting creation")
+				return denyAdmissionRequest(admReview, errors.Wrapf(err, "unable to find the user requesting creation: %s", requestingUsername))
```
alexeykazakov, line 100: As we agreed (multiple times :) We do not want to deny if we it's unknown user (for whatever reason). Check the error. If it's NotFound than log it as a warning/info and allow the request. If it's something else then we can deny with the error message.

pkg/webhook/validatingwebhook/validate.go
```diff
@@ -0,0 +1,144 @@
...
+			}
+			//check if the requesting user is a sandbox user
+			if requestingUser.GetLabels()[toolchainv1alpha1.ProviderLabelKey] == toolchainv1alpha1.ProviderLabelValue {
+				log.Info("trying to give access which is restricted", "unable unmarshal rolebinding json object", "AdmissionReview", admReview)
+				return denyAdmissionRequest(admReview, errors.Wrapf(fmt.Errorf("trying to give access which is restricted"), "Unauthorized request to create rolebinding json object - raw request object: %v", admReview.Request.Object.Raw))
+			}
+
```
alexeykazakov, line 107: By this point you have checked the user and it's not our user. So we do not have to go over other subjects anymore. Just break the loop here.

pkg/webhook/validatingwebhook/validate.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	requestingUsername := admReview.Request.UserInfo.Username
+	// allow admission request if the user is a system user
+	if strings.HasPrefix(requestingUsername, "system:") {
+		return allowAdmissionRequest(admReview)
+	}
+	subjects := rb.Subjects
+	allServiceAccountsSubject := rbac.Subject{
+		Kind:     "Group",
+		Name:     "system:serviceaccounts",
```
alexeykazakov, line 86: Can you also check if we need to block `system:serviceaccounts:` subjects too. With `:` in the end.

pkg/webhook/validatingwebhook/validate.go
```diff
@@ -0,0 +1,147 @@
...
+		APIGroup: "rbac.authorization.k8s.io",
+	}
+	allUsersSubject := rbac.Subject{
+		Kind:     "Group",
+		Name:     "system:authenticated",
```
alexeykazakov, line 91: Same here. Check `system:autenticated:`.

pkg/webhook/validatingwebhook/validate.go
ranakan19: right, my apologies! It wasn't coming intuitively to me to allow after logging error, and i kept missing it. Thank you for pointing this out. fixed in 9d1180e

On a separate note, thanks to this error, i was able to find out that i need to create a service account for validating webhook to access kube client.

pkg/webhook/validatingwebhook/validate.go
ranakan19: done in 9d1180e

pkg/webhook/validatingwebhook/validate.go
ranakan19: done in 9d1180e

alexeykazakov (commented):
Looks good overall. But do you actually test all four cases for forbidden role bindings? Maybe I missed that but it looks like you test only one?

ranakan19:
> Looks good overall. But do you actually test all four cases for forbidden role bindings? Maybe I missed that but it looks like you test only one?

Good point, added the other three rolebindings to the test.

pkg/webhook/validatingwebhook/validate_test.go
```diff
@@ -234,6 +258,204 @@ var sandboxUserJSON = []byte(`{
 	}
 }`)
 
+var sandboxUserJSON2 = []byte(`{
```
alexeykazakov, line 261: Something like that:
```suggestion
var sandboxUserForAllServiceAccountsJSON = []byte(`{
```
would be more informative that JSON1, JSON2, etc.

alexeykazakov (commented):
Looks good! Just one minor comment regarding nameing.

pkg/webhook/validatingwebhook/validate_test.go
ranakan19: done

PR #317, opened 2021-12-13
feat: if appstudio then don't create User nor Identity

MatousJobanek:
We cannot create Identity in AppStudio stage cluster because we don't have claim override enabled there. Otherwise, it fails with: 
```
      conditions:
      - lastTransitionTime: "2021-12-13T14:35:13Z"
        message: 'User.user.openshift.io "mjobanek" is invalid: identities[0]: Invalid
          value: "DevSandbox:f:528d76ff-f708-43ed-8cd5-fe16f4fe0ce6:mjobanek": must
          be in the format <providerName>:<providerUserName>'
        reason: UnableToCreateUser
        status: "False"
        type: Ready
```
This PR skips the creation of `User` & `Identity` if the `tierName` equals to `appstudio`. It also deletes the `User` & `Identity` if it's already present.

paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/424

### DIFF
--- a/controllers/useraccount/useraccount_controller.go
+++ b/controllers/useraccount/useraccount_controller.go
@@ -114,15 +114,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 
 	if !userAcc.Spec.Disabled {
 		logger.Info("ensuring user and identity associated with UserAccount")
-		var createdOrUpdated bool
-		var user *userv1.User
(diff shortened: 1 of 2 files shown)


alexeykazakov (approved):
That will work as a temp workaround.
For the long term we should do the following:
- Sandbox should support both types of IDP clients: with two different sub and originalSub claims (old and new ID format) *and* regular clients with sub claim in the new format (originalSub claim is optional for such clients).
- Space tiers will affect that flow too but we should do the above ^^^ regardless.
cc: @sbryzak

PR #321, opened 2021-12-21
fix: delete identity & user only with owner label

MatousJobanek:
delete the Identity & User resources only when it contains the "owner" label to unblock @alexeykazakov's broken account in AppStudio stage cluster :-)

is paired with: https://github.com/codeready-toolchain/toolchain-e2e/pull/428


alexeykazakov (approved):
Thank you! :)

PR #326, opened 2022-01-11
refactor: map for params when processing a template

xcoulon:
With spaces, the `USERNAME` param gets confusing or inadequate.

### DIFF
--- a/controllers/nstemplateset/cluster_resources.go
+++ b/controllers/nstemplateset/cluster_resources.go
@@ -117,7 +117,7 @@ func (r *clusterResourcesManager) ensure(logger logr.Logger, nsTmplSet *toolchai
 
 		// get all objects of the resource kind from the template (if the template is specified)
 		if tierTemplate != nil {
-			newObjs, err = tierTemplate.process(r.Scheme, username, retainObjectsOfSameGVK(clusterResourceKind.gvk))
+			newObjs, err = tierTemplate.process(r.Scheme, map[string]string{"USERNAME": username}, retainObjectsOfSameGVK(clusterResourceKind.gvk))
 			if err != nil {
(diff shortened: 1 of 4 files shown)


controllers/nstemplateset/cluster_resources.go
```diff
@@ -117,7 +117,7 @@ func (r *clusterResourcesManager) ensure(logger logr.Logger, nsTmplSet *toolchai
 
 		// get all objects of the resource kind from the template (if the template is specified)
 		if tierTemplate != nil {
-			newObjs, err = tierTemplate.process(r.Scheme, username, retainObjectsOfSameGVK(clusterResourceKind.gvk))
+			newObjs, err = tierTemplate.process(r.Scheme, map[string]string{"USERNAME": username}, retainObjectsOfSameGVK(clusterResourceKind.gvk))
```
alexeykazakov, line 120: We could use a const string instead of hardcoding "USERNAME" many times across the package. To make it less error prone :)

controllers/nstemplateset/nstemplatetier.go
```diff
@@ -77,16 +77,13 @@ type tierTemplate struct {
 
 // process processes the template inside of the tierTemplate object and replaces the USERNAME variable with the given username.
```
alexeykazakov, line 78: We need to update the method description too.

alexeykazakov (approved):
Looks good but take a look at my minor comments.

controllers/nstemplateset/cluster_resources.go
xcoulon: done in 161ef06

controllers/nstemplateset/nstemplatetier.go
xcoulon: done in 161ef06

MatousJobanek (approved):
:+1:

PR #327, opened 2022-01-19
feat: ensure space roles in namespace

xcoulon:
create/update/delete roles and role bindings defined in the new
`NSTemplateSet.Spec.SpaceRoles` field.

Also, once the Space role and rolebindings have been created in the
space's namespace, the configuration is stored in an annotation at
the namespace level, so we can compare the new desired config with
the "last-applied" one and remove outdated resource if necessary
(for example, when a user was removed from a group, her rolebinding
should be deleted)

See also:
- https://github.com/codeready-toolchain/api/pull/286
- https://github.com/codeready-toolchain/toolchain-e2e/pull/450

Fixes https://issues.redhat.com/browse/CRT-1337
Fixes https://issues.redhat.com/browse/CRT-1338


controllers/nstemplateset/space_roles.go
```diff
@@ -0,0 +1,136 @@
...
+
+		c, err := deleteOutdatedObjects(logger, r.Client, lastAppliedSpaceRoleObjs, spaceRoleObjs)
+		if err != nil {
+			return false, r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusUpdateFailed, err, "failed to delete redundant objects in namespace '%s'", ns.Name)
+		}
+		changed = changed || c
+		// labels to apply on all new objects
+		var labels = map[string]string{
+			toolchainv1alpha1.ProviderLabelKey: toolchainv1alpha1.ProviderLabelValue,
+			toolchainv1alpha1.OwnerLabelKey:    nsTmplSet.GetName(), // TODO: is it needed?
```
MatousJobanek, line 54:
```
// TODO: is it needed?
```
yes, we need the owner label there so the mapper can map Role & RoleBindings to the name of the NSTemplateSet.

We could technically add also the `username` label and `spacerole` label so we could easily list them, but since we are not sure if we would need them, then let's keep it as it is for now (and introduces them later when/if needed.

controllers/nstemplateset/space_roles.go
```diff
@@ -0,0 +1,136 @@
...
+			"failed to list namespaces for workspace '%s'", nsTmplSet.Name)
+	}
+	logger.Info("ensuring space roles", "namespace_count", len(nss), "role_count", len(nsTmplSet.Spec.SpaceRoles))
+	changed := false
+	for _, ns := range nss {
+		// space roles to apply now
+		lastAppliedSpaceRoleObjs, err := r.getLastAppliedSpaceRoles(&ns)
+		if err != nil {
+			return false, r.wrapErrorWithStatusUpdateForSpaceRolesFailure(logger, nsTmplSet, err, "failed to retrieve last applied space roles")
+		}
```
MatousJobanek, line 40: what about checking the annotation for each of the namespace - if it matches, then we wouldn't do anything here, if it doesn't match, then it would process the templates and deleted/update/create objects that are needed. 
It would significantly decrease the load that is caused by processing the templates.

controllers/nstemplateset/space_roles.go
```diff
@@ -0,0 +1,136 @@
...
+		// create (or update existing) objects based the tier template
+		c, err = applycl.NewApplyClient(r.Client, r.Scheme).Apply(spaceRoleObjs, labels)
+		if err != nil {
+			return false, r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusNamespaceProvisionFailed, err, "failed to provision namespace '%s' with space roles", ns.Name)
+		}
+		changed = changed || c
+		if changed {
+			// store the space roles in an annotation at the namespace level, so we know what was applied and how to deal with
+			// diffs when the space roles are changed (users added or removed, etc.)
```
MatousJobanek, line 68: Updating the annotation only for the case when some object was changed wouldn't work for the case when:
1. objects are updated/deleted/created
2. the actual update of the annotation fails
3. this will trigger another reconcile
4. all objects are already up-to-date so nothing is changed
5. this part would be skipped and the annotation would stay outdated.

I guess that checking the content of the annotation at the very beginning would solve the problem - see my previous comment.

controllers/nstemplateset/space_roles.go
```diff
@@ -0,0 +1,136 @@
...
+					"failed update namespace annotation")
+			}
+		}
+	}
+	return changed, nil
+}
+
+// Get the space role objects from the templates specified in the given `spaceRoles`
+// Returns the objects indexed by kind and name, or an error if something wrong happened when processing the templates
+func (r *spaceRolesManager) getSpaceRoles(ns *corev1.Namespace, spaceRoles []toolchainv1alpha1.NSTemplateSetSpaceRole) ([]runtimeclient.Object, error) {
```
MatousJobanek, line 85: I would maybe name it `getSpaceRolesObjects` but I don't insist on it. 

btw, what do you mean by: 
```
// Returns the objects indexed by kind and name
```

controllers/nstemplateset/space_roles.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/go-logr/logr"
+)
+
+type spaceRolesManager struct {
+	*statusManager
+}
+
+// ensure ensures that the space roles for the users exist.
+// Returns `true, nil` if something was changed, `false, nil` if nothing changed, `false, err` if an error occurred
+// TODO: support update/removal/outdated
```
MatousJobanek, line 24:
```
 // TODO: support update/removal/outdated
```
that's already done, right?

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -1923,4 +1979,67 @@ var (
...
+  rules:
+    # examples
+    - apiGroups:
+        - ""
+      resources:
+        - "secrets"
+        - "serviceaccounts"
+      verbs:
+        - get
+        - list
```
MatousJobanek, line 2029: I know that it should have no effect on the result of the tests, but the `space-viewer` role has the same content as `space-admin` - it would make more sense if they were different

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -1766,45 +1786,78 @@ func getTemplateTiers(decoder runtime.Decoder) ([]runtime.Object, error) {
...
+	}, nil
+}
+
 var (
 	ns test.TemplateObject = `
 - apiVersion: v1
   kind: Namespace
   metadata:
-    name: ${USERNAME}-nsType
+    name: ${USERNAME}-NSTYPE
```
MatousJobanek, line 1853: just out of curiosity - why this change?

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -1766,45 +1786,78 @@ func getTemplateTiers(decoder runtime.Decoder) ([]runtime.Object, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	tierTemplates = append(tierTemplates, tierTmpl)
+	tierTmpl, err = createTierTemplate(decoder,
+		test.CreateTemplate(test.WithObjects(spaceViewer, spaceViewerRb), test.WithParams(namespace, username)),
+		"viewer", "appstudio", "abcde11")
+	if err != nil {
+		return nil, err
+	}
```
MatousJobanek, line 1821: as we discussed in the call - the SpaceRoles are not `appstudio` specific, but they will be used in all tiers. 
I would propose dropping this part and incorporating the SpaceRoles into the already existing tier setup - otherwise, it could bring additional confusion why we need a special `appstudio` tier for that.

controllers/nstemplateset/space_roles.go
```diff
@@ -0,0 +1,136 @@
...
+		return nil, err
+	}
+	return namespaces.Items, nil
+}
+
+// listByWorkspaceLabel returns runtimeclient.ListOption that filters by label toolchain.dev.openshift.com/workspace equal to the given workspace name
+func listByWorkspaceLabel(w string) runtimeclient.ListOption {
+	labels := map[string]string{toolchainv1alpha1.WorkspaceLabelKey: w}
+	return runtimeclient.MatchingLabels(labels)
+}
```
MatousJobanek, line 136: as it was discussed in the call - don't list them by `workspace` label as SpaceRoles are not appstudio specific feature.

instead of that, you could use the already existing function `fetchNamespacesByOwner`

controllers/nstemplateset/space_roles_test.go
```diff
@@ -0,0 +1,210 @@
...
+				nsTmplSet := newNSTmplSet(commontest.MemberOperatorNs, "oddity", "appstudio",
+					withSpaceRoles(map[string][]string{
+						"admin-appstudio-abcde11": {"user1", "user2"},
+					}))
+				ns := newNamespace(nsTmplSet.Spec.TierName, "oddity", "appstudio", // ns.name=oddity=appstudio
+					withTemplateRefUsingRevision("abcde10"), // starting with an older revision
+					withWorkspaceLabel("oddity"),
+				)
+				mgr, memberClient := prepareSpaceRolesManager(t, nsTmplSet, ns)
+				_, err := mgr.ensure(logger, nsTmplSet)
```
MatousJobanek, line 92: could you please add a comment mentioning why the ensure is called here? and could you please add it to the following ones as well?
```suggestion
				// running ensure to precreate the resources for the initial set of SpaceRoles
				_, err := mgr.ensure(logger, nsTmplSet)
```

controllers/nstemplateset/status.go
```diff
@@ -102,6 +110,17 @@ func (r *statusManager) setStatusClusterResourcesProvisionFailed(nsTmplSet *tool
 		})
 }
 
+func (r *statusManager) setStatuSpaceRolesProvisionFailed(nsTmplSet *toolchainv1alpha1.NSTemplateSet, message string) error {
```
MatousJobanek, line 113: small typo: 
```suggestion
func (r *statusManager) setStatusSpaceRolesProvisionFailed(nsTmplSet *toolchainv1alpha1.NSTemplateSet, message string) error {
```

controllers/nstemplateset/space_roles_test.go
```diff
@@ -0,0 +1,210 @@
...
+				require.NoError(t, err)
+				assert.True(t, createdOrUpdated)
+				// verify that role bindings for `user1` and `user2`` still exist and thet one was added for `user3`
+				AssertThatRole(t, "oddity-appstudio", "space-admin", memberClient).Exists()              // unchanged
+				AssertThatRoleBinding(t, "oddity-appstudio", "user1-space-admin", memberClient).Exists() // unchanged
+				AssertThatRoleBinding(t, "oddity-appstudio", "user2-space-admin", memberClient).Exists() // unchanged
+				AssertThatRoleBinding(t, "oddity-appstudio", "user3-space-admin", memberClient).Exists() // created
+			})
+
+			t.Run("remove admin user", func(t *testing.T) {
```
MatousJobanek, line 110: could you please verify also the case when one whole SpaceRole is removed with all usernames? Expecting that the `Role` would be deleted from the namespace.

MatousJobanek (commented):
This is really good stuff :+1: Simple and elegant solution that covers all cases that we need :rocket: :1st_place_medal:

controllers/nstemplateset/namespaces.go
```diff
@@ -131,7 +132,7 @@ func (r *namespacesManager) ensureInnerNamespaceResources(logger logr.Logger, ns
 		if err != nil {
 			return r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusUpdateFailed, err, "failed to process template for TierTemplate with name '%s'", currentRef)
 		}
-		if _, err := deleteRedundantObjects(logger, r.Client, false, currentObjs, newObjs); err != nil {
+		if _, err := deleteOutdatedObjects(logger, r.Client, currentObjs, newObjs); err != nil {
```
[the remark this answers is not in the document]
xcoulon: maybe "obsolete" is a better wording, then? WDYT?

controllers/nstemplateset/space_roles.go
xcoulon: thanks for your response @MatousJobanek! I was hoping you'd notice this `//TODO` comment and provide me with an answer during the review 😛

controllers/nstemplateset/space_roles.go
xcoulon: > I would maybe name it getSpaceRolesObjects but I don't insist on it.

sure, let me do that

> btw, what do you mean by:

ah never mind, this comment is outdated, I'll remove it

controllers/nstemplateset/namespaces.go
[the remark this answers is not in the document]
xcoulon: renamed in 26cbaa7 (unless or until we have a better name)

controllers/nstemplateset/space_roles.go
```diff
@@ -0,0 +1,136 @@
...
+		lastAppliedSpaceRoleObjs, err := r.getLastAppliedSpaceRoles(&ns)
+		if err != nil {
+			return false, r.wrapErrorWithStatusUpdateForSpaceRolesFailure(logger, nsTmplSet, err, "failed to retrieve last applied space roles")
+		}
+		spaceRoleObjs, err := r.getSpaceRoles(&ns, nsTmplSet.Spec.SpaceRoles)
+		if err != nil {
+			return false, r.wrapErrorWithStatusUpdateForSpaceRolesFailure(logger, nsTmplSet, err, "failed to retrieve space roles to apply")
+		}
+
+		c, err := deleteOutdatedObjects(logger, r.Client, lastAppliedSpaceRoleObjs, spaceRoleObjs)
```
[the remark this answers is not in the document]
xcoulon: renamed to `deleted` in d89f12b

controllers/nstemplateset/space_roles.go
xcoulon: comment removed in e5d2100

controllers/nstemplateset/space_roles.go
xcoulon: good idea! done in 0034497

controllers/nstemplateset/space_roles.go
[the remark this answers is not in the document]
xcoulon: aaaand renamed to `applied` in line 61 in 2534490

controllers/nstemplateset/space_roles.go
xcoulon: I removed the `if changed` statement in 9ae349a since at this point (with https://github.com/codeready-toolchain/member-operator/commit/0034497a60a4d157b14e5ccacc0f538286313943), we should expect that eithr obsolete resources were removed or resources were created/updated.

xcoulon: and if an error occurred when updating the annotation, then there will be a new reconcile loop, no resource will need to be created/updated/removed, but the annotation will be updated (hopefully for good this time)

controllers/nstemplateset/space_roles.go
xcoulon: done in a58732d

controllers/nstemplateset/space_roles.go
xcoulon: yes ;) 
removed in c3c53cc

controllers/nstemplateset/nstemplateset_controller_test.go
xcoulon: sure, but indeed, it doesn't affect the tests

xcoulon: done in 0ca2085

controllers/nstemplateset/nstemplateset_controller_test.go
xcoulon: because I think it is more visible (ie, understandable) if the character sequence is fully capitalised. But it's just a cosmetic change.

controllers/nstemplateset/space_roles.go
xcoulon: done in 182a0cf

controllers/nstemplateset/nstemplateset_controller_test.go
xcoulon: done in 4c78c4e. I also refactored the `prepareTemplateTiers` func so that it's visually easier to see what templates are generated

controllers/nstemplateset/space_roles_test.go
xcoulon: done in a1919ee

controllers/nstemplateset/space_roles_test.go
xcoulon: done in 789711c (as a subtest of the creation test, to avoid too much code duplication)

controllers/nstemplateset/space_roles.go
```diff
@@ -0,0 +1,116 @@
...
+		if currentSpaceRolesAnnotation, exists := ns.Annotations[toolchainv1alpha1.LastAppliedSpaceRolesAnnotationKey]; exists && currentSpaceRolesAnnotation != "" {
+			if err := json.Unmarshal([]byte(currentSpaceRolesAnnotation), &lastAppliedSpaceRoles); err != nil {
+				return false, errors.Wrap(err, "unable to decode current space roles in annotation")
+			}
+		}
+		// compare last-applied vs spec to see if there's anything obsolete to do
+		if reflect.DeepEqual(nsTmplSet.Spec.SpaceRoles, lastAppliedSpaceRoles) {
+			logger.Info("no space role to update", "namespace", ns.Name)
+			continue
+		}
```
MatousJobanek, line 48: I guess that the code doesn't set the status if the SpaceRoles are not up-to-date:
```suggestion
		}
		if err := r.setStatusUpdatingIfNotProvisioning(nsTmplSet); err != nil {
			return false, err
		}
```

controllers/nstemplateset/space_roles.go
```diff
@@ -0,0 +1,116 @@
...
+		// space roles to apply now
+		spaceRoleObjs, err := r.getSpaceRolesObjects(&ns, nsTmplSet.Spec.SpaceRoles)
+		if err != nil {
+			return false, r.wrapErrorWithStatusUpdateForSpaceRolesFailure(logger, nsTmplSet, err, "failed to retrieve space roles to apply")
+		}
+
+		deleted, err := deleteObsoleteObjects(logger, r.Client, lastAppliedSpaceRoleObjs, spaceRoleObjs)
+		if err != nil {
+			return false, r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusUpdateFailed, err, "failed to delete redundant objects in namespace '%s'", ns.Name)
```
MatousJobanek, line 61: today morning I just realized that we should apply all objects at first and then delete the obsolete ones. The reason is: when an administrator wants to change a SpaceRole of one SpaceBinding, we should make sure that the user access is not affected by the change anyhow - the user should always have the access.

controllers/nstemplateset/space_roles.go
```diff
@@ -0,0 +1,116 @@
...
+		changed = changed || deleted
+		// labels to apply on all new objects
+		var labels = map[string]string{
+			toolchainv1alpha1.ProviderLabelKey: toolchainv1alpha1.ProviderLabelValue,
+			toolchainv1alpha1.OwnerLabelKey:    nsTmplSet.GetName(),
+		}
+		logger.Info("creating space role objects")
+		for _, obj := range spaceRoleObjs {
+			logger.Info(fmt.Sprintf("creating/updating %s/%s", obj.GetNamespace(), obj.GetName()))
+		}
```
MatousJobanek, line 72: do we really need that? wouldn't it be better to add some logging mechanism to the ApplyClient?

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -1722,89 +1734,156 @@ func withTemplateRefUsingRevision(revision string) objectMetaOption {
...
-						Namespace: test.HostOperatorNs,
-					},
-					Spec: toolchainv1alpha1.TierTemplateSpec{
-						TierName: tierName,
-						Type:     typeName,
-						Revision: revision,
-						Template: tmpl,
-					},
-				}
+				t.Logf("adding tiertemplate '%s", tierTmpl.Name)
```
MatousJobanek, line 1845: this produces quite a lot of lines in the test log and (I guess that) they are not super useful 
```suggestion
```

controllers/nstemplateset/space_roles.go
```diff
@@ -0,0 +1,116 @@
...
+	spaceRoleObjects := []runtimeclient.Object{}
+	for _, spaceRole := range spaceRoles {
+		tierTemplate, err := getTierTemplate(r.GetHostCluster, spaceRole.TemplateRef)
+		if err != nil {
+			return nil, err
+		}
+		for _, username := range spaceRole.Usernames {
+			objs, err := tierTemplate.process(r.Scheme, map[string]string{
+				Namespace: ns.Name,
+				Username:  username,
```
MatousJobanek, line 107: btw, I just noticed that the `Username` parameter is used in all cases when the templates are processed - it could be moved to the `process` function itself. Can be done in separate PR :wink:

controllers/nstemplateset/nstemplateset_controller.go
```diff
@@ -199,30 +211,27 @@ func (r *Reconciler) deleteNSTemplateSet(logger logr.Logger, nsTmplSet *toolchai
...
-			return false, errs.Wrapf(err, "failed to delete object '%s' of kind '%s' in namespace '%s'", currentObj.GetName(), currentObj.GetObjectKind().GroupVersionKind().Kind, currentObj.GetNamespace())
+			return false, errs.Wrapf(err, "failed to delete outdated object '%s' of kind '%s' in namespace '%s'", currentObj.GetName(), currentObj.GetObjectKind().GroupVersionKind().Kind, currentObj.GetNamespace())
 		} else if errors.IsNotFound(err) {
 			continue // continue to the next object since this one was already deleted
 		}
-		logger.Info("deleted redundant object", "objectName", currentObj.GetObjectKind().GroupVersionKind().Kind+"/"+currentObj.GetName())
-		if deleteOnlyOne {
-			return true, nil
-		}
+		logger.Info("deleted outdated object", "objectName", currentObj.GetObjectKind().GroupVersionKind().Kind+"/"+currentObj.GetName())
```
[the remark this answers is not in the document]
xcoulon: ah right, I missed that! done in a4ebfa7

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -1722,89 +1734,156 @@ func withTemplateRefUsingRevision(revision string) objectMetaOption {
...
-						tmplContent = test.CreateTemplate(test.WithObjects(advancedCrq, emptyCrq, clusterTektonRb), test.WithParams(username))
-					default:
-						tmplContent = test.CreateTemplate(test.WithObjects(ns, rb, role), test.WithParams(username))
-					}
-				}
-				tmplContent = strings.ReplaceAll(tmplContent, "nsType", typeName)
-				tmpl := templatev1.Template{}
-				_, _, err := decoder.Decode([]byte(tmplContent), nil, &tmpl)
+	// templates
+	tmpls := map[string]map[string]map[string]string{
```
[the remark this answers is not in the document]
xcoulon: sure! done in 997d213

controllers/nstemplateset/nstemplateset_controller_test.go
xcoulon: sure, but those logs will only be printed if the test fails, AFAIK

controllers/nstemplateset/space_roles.go
xcoulon: I would rather keep it there for the sake of readability: having all params used to process a template in a single place is simpler for maintenance IMHO. Even if it causes a bit of duplication in the code

controllers/nstemplateset/space_roles.go
xcoulon: removed in 3d00213. We can add them in toolchain-common later if needed.

controllers/nstemplateset/space_roles.go
xcoulon: done in 46a41e0

MatousJobanek (approved):
Looks good, thanks a lot for addressing my comments. :pray: 
I have noticed one more thing - you don't verify the condition in the NSTemplateSet.Status in your test. I believe that it would be worth checking it for all cases - the error ones as well as the successful ones. WDYT?

xcoulon:
> Looks good, thanks a lot for addressing my comments. 🙏 I have noticed one more thing - you don't verify the condition in the NSTemplateSet.Status in your test. I believe that it would be worth checking it for all cases - the error ones as well as the successful ones. WDYT?

good point, let me add that!

xcoulon:
> 87b2b97



> > Looks good, thanks a lot for addressing my comments. 🙏 I have noticed one more thing - you don't verify the condition in the NSTemplateSet.Status in your test. I believe that it would be worth checking it for all cases - the error ones as well as the successful ones. WDYT?
> 
> good point, let me add that!

done in https://github.com/codeready-toolchain/member-operator/pull/327/commits/87b2b97ab99a0a9137983f7a2983bd2a29eeb05a

PR #333, opened 2022-01-28
Use useraccount tier label instead of nstemplateset tiername

rajivnathan:
This is a follow up from https://github.com/codeready-toolchain/host-operator/pull/580 that is a prerequisite before NSTemplateSet can be removed from UserAccount.

No new/updated e2e test because it should be covered by the existing assertions: https://github.com/codeready-toolchain/toolchain-e2e/blob/2c84a1ea19e3085ae4e055d00a7db6315bbce74c/testsupport/user_assertions.go#L72-L103


controllers/useraccount/useraccount_controller.go
```diff
@@ -167,8 +167,8 @@ func (r *Reconciler) ensureUserAndIdentity(logger logr.Logger, userAcc *toolchai
 	var createdOrUpdated bool
 	var user *userv1.User
 	var err error
-	// for AppStudio use-case - create User & Identity resources if user is not using appstudio tier
-	if userAcc.Spec.NSTemplateSet == nil || userAcc.Spec.NSTemplateSet.TierName != "appstudio" {
+	// for non-AppStudio use-case - create User & Identity resources if user is not using appstudio tier
+	if userAcc.Spec.NSTemplateSet == nil || (userAcc.Labels != nil && userAcc.Labels[toolchainv1alpha1.TierLabelKey] != "appstudio") {
```
MatousJobanek, line 171: you don't need the first evaluation `userAcc.Spec.NSTemplateSet == nil` any more there, otherwise, when you do the migration it would provision `User` & `Identity` even for appstudio users (the `NSTemplateSet` will be nill for all UserAccounts). 
In addition, the default behavior should be the creation of the resources (`User` & `Identity`), not deleting them - the resources should not be present only and only when the tier label is `appstudio`, so when either the labels are nil or the label is not present, then it should follow the default procedure.
```suggestion
	if userAcc.Labels == nil || userAcc.Labels[toolchainv1alpha1.TierLabelKey] != "appstudio" {
```

PR #336, opened 2022-02-08
fix: set syncIndex to 0 and then to deleted while deleting UserAccount

MatousJobanek:
https://coreos.slack.com/archives/CHK0J6HT6/p1644277305114849


rajivnathan (approved):
Thanks 🥇

PR #337, opened 2022-02-09
add provider label to user and identity

ranakan19:
All resources created by sandbox should have the provider label populated for uniformity and easy identification of resources. fix wrt - https://issues.redhat.com/browse/CRT-1436
related e2e test - https://github.com/codeready-toolchain/toolchain-e2e/pull/474

This also unblocks the logic for https://github.com/codeready-toolchain/member-operator/pull/316. Although the webhook here could use the owner label instead, so it wasn't a hard block.

### DIFF
--- a/controllers/useraccount/useraccount_controller.go
+++ b/controllers/useraccount/useraccount_controller.go
@@ -3,9 +3,10 @@ package useraccount
 import (
 	"context"
 	"fmt"
-	commonidentity "github.com/codeready-toolchain/toolchain-common/pkg/identity"
 	"reflect"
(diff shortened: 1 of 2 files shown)


MatousJobanek (commented):
Looks good. Could you please update also e2e tests so it verifies that the new label is added to both identity and user?

ranakan19:
> Looks good. Could you please update also e2e tests so it verifies that the new label is added to both identity and user?

e2e test added here - https://github.com/codeready-toolchain/toolchain-e2e/pull/474

PR #339, opened 2022-02-16
Adjust logic and add unit test

rajivnathan:
Unit tests for https://github.com/codeready-toolchain/member-operator/pull/337

After adding the unit test with a User/Identity that had nil labels the tests were failing because the assertUser function checks the following:
```
assert.Equal(t, userAcc.Name, user.Labels["toolchain.dev.openshift.com/owner"])
assert.Equal(t, toolchainv1alpha1.ProviderLabelValue, user.Labels[toolchainv1alpha1.ProviderLabelKey])
```

So I'm assuming it's always expected that both owner and provider labels should be present? At least that's the assumption I went with to make the unit tests pass. I changed the `addProviderLabel` func to `addOwnerAndProviderLabels` so it adds both. Please verify that that's what we actually want.

### DIFF
--- a/controllers/useraccount/useraccount_controller.go
+++ b/controllers/useraccount/useraccount_controller.go
@@ -254,7 +254,7 @@ func (r *Reconciler) ensureUser(logger logr.Logger, config membercfg.Configurati
 	logger.Info("user already exists")
 	// migration step - add provider label to existing users if it doesn't exist
 	//TODO - remove this after migration complete
-	err := addProviderLabel(user, r.Client)
+	err := addOwnerAndProviderLabels(user, r.Client, userAcc.Name)
 	if err != nil {
 		logger.Error(err, "Unable to update labels to add provider")
 	}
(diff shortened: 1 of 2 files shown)


controllers/useraccount/useraccount_controller.go
```diff
@@ -385,13 +385,29 @@ func setOwnerAndProviderLabel(object metav1.Object, owner string) {
...
 		labels[toolchainv1alpha1.ProviderLabelKey] = toolchainv1alpha1.ProviderLabelValue
+		object.SetLabels(labels)
+		changed = true
+	}
+
+	if _, exists := object.GetLabels()[toolchainv1alpha1.OwnerLabelKey]; !exists {
+		labels := object.GetLabels()
+		if labels == nil {
+			labels = make(map[string]string)
+		}
```
MatousJobanek, line 404: it's just a detail, but this case shouldn't ever happen - if the labels are nil, then they would be set in the first case (when checking provider label)

MatousJobanek (approved):

PR #345, opened 2022-03-03
feat: support optional resources

MatousJobanek:
Implements https://issues.redhat.com/browse/CRT-1447
If a TierTemplate object is marked as optional, then it checks if the corresponding API group is present in the cluster. If not, then it doesn't apply the object. 
In addition, it also adds DBaaSTenant as an additional cluster resource to be watched, but only if the API group is present.


controllers/nstemplateset/client.go
```diff
@@ -0,0 +1,53 @@
+package nstemplateset
+
+import (
+	"github.com/codeready-toolchain/api/api/v1alpha1"
```
xcoulon, line 4: could be aliased to `toolchainv1alpha1` like everywhere else

controllers/nstemplateset/client_test.go
```diff
@@ -0,0 +1,199 @@
...
+		changed, err := apiClient.ApplyToolchainObjects(logger, objects(dBaaSTenant), additionalLabel)
+
+		// then
+		require.NoError(t, err)
+		assert.True(t, changed)
+		assertObject(t, fakeClient, true)
+	})
+}
+
+func objects(objects ...runtimeclient.Object) []runtimeclient.Object {
```
xcoulon, line 92:
```suggestion
func copyObjects(objects ...runtimeclient.Object) []runtimeclient.Object {
```

so we know that the given objects will be copied

controllers/nstemplateset/client_test.go
```diff
@@ -0,0 +1,199 @@
...
+			assert.Contains(t, dBaaSTenant.Labels, toolchainv1alpha1.ProviderLabelKey)
+			assert.Equal(t, toolchainv1alpha1.ProviderLabelValue, dBaaSTenant.Labels[toolchainv1alpha1.ProviderLabelKey])
+			assert.Contains(t, dBaaSTenant.Labels, toolchainv1alpha1.OwnerLabelKey)
+			assert.Equal(t, "john", dBaaSTenant.Labels[toolchainv1alpha1.OwnerLabelKey])
+			assert.Contains(t, dBaaSTenant.Labels, "foo")
+			assert.Equal(t, "bar", dBaaSTenant.Labels["foo"])
+		})
+	} else {
+		// should not exist
+		AssertObject(t, client, "john-dev", "john-dev-tenant", dBaaSTenant, nil)
```
xcoulon, line 123: we should have an `AssertObjectDoesNotExist()` func instead of passing `nil` as the last argument here. But that can be addressed in a separate PR.

controllers/nstemplateset/client_test.go
```diff
@@ -0,0 +1,199 @@
...
+
+func objects(objects ...runtimeclient.Object) []runtimeclient.Object {
+	var objs []runtimeclient.Object
+	for i := range objects {
+		objs = append(objs, objects[i].DeepCopyObject().(runtimeclient.Object))
+	}
+	return objs
+}
+
+func assertObject(t *testing.T, client *test.FakeClient, expectDBaaSTenant bool) {
```
xcoulon, line 100: should be named `assertObjects` since it verifies the `Role`, the `Namespace` and optionally the `DBaaSTenant`

controllers/nstemplateset/client_test.go
```diff
@@ -0,0 +1,199 @@
...
+func TestApplyToolchainObjects(t *testing.T) {
+	// given
+	role := newRole("john-dev", "edit-john", "john")
+	devNs := newNamespace("advanced", "john", "dev")
+	dBaaSTenant := newDBaaSTenant("john-dev-tenant", "john")
+	additionalLabel := map[string]string{
+		"foo": "bar",
+	}
+
+	t.Run("when create two", func(t *testing.T) {
```
xcoulon, line 32:
```suggestion
	t.Run("when creating two objects", func(t *testing.T) {
```

controllers/nstemplateset/client_test.go
```diff
@@ -0,0 +1,199 @@
...
+		// when
+		changed, err := apiClient.ApplyToolchainObjects(logger, objects(role, devNs), additionalLabel)
+
+		// then
+		require.NoError(t, err)
+		assert.True(t, changed)
+		assertObject(t, fakeClient, false)
+	})
+
+	t.Run("when create only one, the second is present", func(t *testing.T) {
```
xcoulon, line 45:
```suggestion
	t.Run("when creating only one object because the other one already exists", func(t *testing.T) {
```

controllers/nstemplateset/client_test.go
```diff
@@ -0,0 +1,199 @@
...
+		// then
+		require.NoError(t, err)
+		assert.False(t, changed)
+		assertObject(t, fakeClient, false)
+	})
+
+	t.Run("when only DBaaSTenant is supposed to be applied and the group for DBaaS is present", func(t *testing.T) {
+		// given
+		apiClient, fakeClient := prepareAPIClient(t)
+		apiClient.AvailableAPIGroups = append(apiClient.AvailableAPIGroups, metav1.APIGroup{Name: "dbaas.redhat.com"})
```
xcoulon, line 78: shouldn't we check/match the group version as well?

controllers/nstemplateset/cluster_resources.go
```diff
@@ -35,14 +40,19 @@ type listExistingResources func(cl runtimeclient.Client, username string) ([]run
...
 }
 
 func newToolchainObjectKind(gvk schema.GroupVersionKind, emptyObject runtimeclient.Object, listExistingResources listExistingResources) toolchainObjectKind {
 	return toolchainObjectKind{
-		gvk:                   gvk,
-		object:                emptyObject,
-		listExistingResources: listExistingResources,
+		gvk:    gvk,
+		object: emptyObject,
+		listExistingResources: func(cl runtimeclient.Client, username string, availableAPIGroups []metav1.APIGroup) (objects []runtimeclient.Object, e error) {
```
xcoulon, line 50: there's something a bit confusing here: `listExistingResources` has the same type as `listExistingResourcesIfAvailable` at line 33 above, but it uses the same name as type in line 30.

maybe we should just drop the `listExistingResourcesIfAvailable` type definition, if it's not used anywhere else?

controllers/nstemplateset/cluster_resources_test.go
```diff
@@ -103,12 +106,24 @@ func TestClusterResourceKinds(t *testing.T) {
...
 			// when
-			existingResources, err := clusterResourceKind.listExistingResources(fakeClient, "johny")
+			existingResources, err := clusterResourceKind.listExistingResources(fakeClient, "johny", apiGroups)
 
 			// then
 			require.Error(t, err)
 			require.Len(t, existingResources, 0)
 		})
+
+		t.Run("listExistingResources should return not return any resource when APIGroup is missin for gvk "+clusterResourceKind.gvk.String(), func(t *testing.T) {
```
xcoulon, line 116:
```suggestion
		t.Run("listExistingResources should return not return any resource when APIGroup is missing for gvk "+clusterResourceKind.gvk.String(), func(t *testing.T) {
```

controllers/nstemplateset/cluster_resources_test.go
```diff
@@ -103,12 +106,24 @@ func TestClusterResourceKinds(t *testing.T) {
...
+			// given
+			fakeClient := test.NewFakeClient(t, anotherObject, johnyObject, johnyObject2, namespace)
+
+			// when
+			existingResources, err := clusterResourceKind.listExistingResources(fakeClient, "johny", newAPIGroups("apps", ""))
+
+			// then
+			require.NoError(t, err)
+			require.Len(t, existingResources, 0)
```
xcoulon, line 125:
```suggestion
			require.Empty(t, existingResources)
```

pkg/apis/apis.go
```diff
@@ -1,9 +1,10 @@
 package apis
 
 import (
+	dbaasv1alpha1 "github.com/RHEcosystemAppEng/dbaas-operator/api/v1alpha1"
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
-	v1 "k8s.io/api/admissionregistration/v1"
-	v12 "k8s.io/api/scheduling/v1"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
```
xcoulon, line 6: such better aliases! 🤩

controllers/nstemplateset/client_test.go
xcoulon: see https://github.com/codeready-toolchain/member-operator/pull/347 (with `AssertObjectNotFound`, which is a shorter name and matches with the `NotFound` error)

controllers/nstemplateset/client.go
MatousJobanek: done cf2f848

controllers/nstemplateset/client_test.go
MatousJobanek: renamed dd55e14

controllers/nstemplateset/client_test.go
MatousJobanek: will be addressed in your PR - as part of solving the conflicts

controllers/nstemplateset/client_test.go
MatousJobanek: renamed 6c01621

controllers/nstemplateset/client_test.go
MatousJobanek: for some reason, I thought that the version is not part of the result, but it is
added the check here: 74ff7e3
Thanks - good catch :+1:

controllers/nstemplateset/cluster_resources.go
MatousJobanek: good point - didn't notice that cb30d01

controllers/nstemplateset/cluster_resources_test.go
MatousJobanek: fixed as part of https://github.com/codeready-toolchain/member-operator/commit/74ff7e3d8aa087a70c8b6a8b3863eac8d4ff2642

controllers/nstemplateset/cluster_resources_test.go
MatousJobanek: :man_facepalming:

MatousJobanek:
@xcoulon Thanks a lot for your review - I believe that I either addressed or answered all your comments. :pray:

xcoulon (approved):
thanks for addressing my comments/suggestions! 🙌

PR #360, opened 2022-04-26
refactor: returned bool from ApplyToolchainObjects

xcoulon:
return `true` when at least one resource is applied,
`false` otherwise.

Notes:
In some tests, the rolebinding resources were replaced by
idlers in order to have changes in the specs, which in turn
trigger a resource update at the client level (ie, the resource
generation changed).
As a consequence, when the only change in the resource is at the labels
level, the client returns `false` and the controller moves on the next
resource, instead of exiting the reconcile loop. This means less reconcile
loops to process a whole NSTemplateSet update (also affects the unit-tests)

Also, rename more `code` namespaces to `stage` in unit tests


xcoulon:
closing, not needed anymore

PR #361, opened 2022-04-26
feat: recreate space-role role and rolebindings

xcoulon:
recreate role and rolebindings when missing (eg, deleted by user)
note: for this particular use-case, all missing resources are recreated
at once (in all related namespaces) and the NSTemplateSet status condition
stays at `provisioned`.

### DIFF
--- a/controllers/nstemplateset/space_roles.go
+++ b/controllers/nstemplateset/space_roles.go
@@ -3,6 +3,7 @@ package nstemplateset
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"reflect"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
@@ -38,13 +39,13 @@ func (r *spaceRolesManager) ensure(logger logr.Logger, nsTmplSet *toolchainv1alp
 				return false, errors.Wrap(err, "unable to decode current space roles in annotation")
 			}
 		}
-		// compare last-applied vs spec to see if there's anything obsolete to do
-		if reflect.DeepEqual(nsTmplSet.Spec.SpaceRoles, lastAppliedSpaceRoles) {
-			logger.Info("no space role to update", "namespace", ns.Name)
-			continue
-		}
-		if err := r.setStatusUpdatingIfNotProvisioning(nsTmplSet); err != nil {
-			return false, err
+		// compare last-applied vs spec to see if there's anything obsolete
+		// note: we only set the NSTemplateSet status to `provisioning` if there are "configuration" changes,
+		// but for other cases (such as restoring resources deleted by a user), we don't set the NSTemplateSet status to `provisioning`.
+		if !reflect.DeepEqual(nsTmplSet.Spec.SpaceRoles, lastAppliedSpaceRoles) {
+			if err := r.setStatusUpdatingIfNotProvisioning(nsTmplSet); err != nil {
+				return false, err
+			}
 		}
 		lastAppliedSpaceRoleObjs, err := r.getSpaceRolesObjects(&ns, lastAppliedSpaceRoles) // nolint:gosec
 		if err != nil {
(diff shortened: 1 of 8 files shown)


controllers/nstemplateset/space_roles_test.go
```diff
@@ -72,11 +72,19 @@ func TestEnsureSpaceRoles(t *testing.T) {
...
+				_, err := mgr.ensure(logger, nsTmplSet)
+
+				// then
+				// verify that the `last-applied-space-roles` annotation was set on namespace
+				require.NoError(t, err)
+				lastApplied, err := json.Marshal(nsTmplSet.Spec.SpaceRoles)
+				require.NoError(t, err)
+				AssertThatNamespace(t, "oddity-appstudio", memberClient.Client).
+					HasAnnotation(toolchainv1alpha1.LastAppliedSpaceRolesAnnotationKey, string(lastApplied))
+			})
```
MatousJobanek, line 87: just out of curiosity - why the original code couldn't work? why do we need to call the `ensure` method one more time?

test/nstemplateset_assertion.go
```diff
@@ -161,6 +161,15 @@ func UpdateFailed(msg string) toolchainv1alpha1.Condition {
...
 
+func UnableToProvisionSpaceRoles(msg string) toolchainv1alpha1.Condition {
+	return toolchainv1alpha1.Condition{
+		Type:    toolchainv1alpha1.ConditionReady,
+		Status:  corev1.ConditionFalse,
+		Reason:  toolchainv1alpha1.NSTemplateSetUnableToProvisionSpaceRolesReason,
+		Message: msg,
+	}
+}
+
```
MatousJobanek, line 172: where is this used?

MatousJobanek (approved):
Looks good - just two comments/questions.

controllers/nstemplateset/space_roles.go
```diff
@@ -61,33 +62,36 @@ func (r *spaceRolesManager) ensure(logger logr.Logger, nsTmplSet *toolchainv1alp
 			toolchainv1alpha1.ProviderLabelKey: toolchainv1alpha1.ProviderLabelValue,
 			toolchainv1alpha1.OwnerLabelKey:    nsTmplSet.GetName(),
 		}
-		logger.Info("creating space role objects")
+		logger.Info("creating space role objects", "count", len(spaceRoleObjs))
```
rajivnathan, line 65: This is a minor detail but I guess it would be more clear to say 'applying' since `len(spaceRoleObjs)` is the expected number of objects rather than the number we will actually create right? eg. If 1 object is deleted we would only expect to create 1

```suggestion
		logger.Info("applying space role objects", "count", len(spaceRoleObjs))
```

controllers/nstemplateset/space_roles.go
```diff
@@ -61,33 +62,36 @@ func (r *spaceRolesManager) ensure(logger logr.Logger, nsTmplSet *toolchainv1alp
...
+			sr, err := json.Marshal(nsTmplSet.Spec.SpaceRoles)
+			if err != nil {
+				return false, r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusProvisionFailed, err,
+					fmt.Sprintf("failed to marshal space roles to update '%s' annotation on namespace", toolchainv1alpha1.LastAppliedSpaceRolesAnnotationKey))
+			}
+			if ns.Annotations == nil {
+				ns.Annotations = map[string]string{}
+			}
+			ns.Annotations[toolchainv1alpha1.LastAppliedSpaceRolesAnnotationKey] = string(sr)
+			logger.Info("updated annotation on namespace", toolchainv1alpha1.LastAppliedSpaceRolesAnnotationKey, string(sr))
```
rajivnathan, line 87: It looks like this message got moved before the call to Update when it was previously after? I guess it should be moved to after the Update?

controllers/nstemplateset/space_roles_test.go
xcoulon: indeed, it's not needed. I guess I wrote this test with the expectation that we would need an extra reconcile loop once all resources have been created/updated/deleted, but that was _before_ we agreed on the new strategy ("recreate missing resources and that's all"). 
Reverted in b92fadf

test/nstemplateset_assertion.go
xcoulon: it's a left-over, not needed. Removed in b4a67cb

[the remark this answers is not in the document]
xcoulon: removed in b4a67cb

controllers/nstemplateset/space_roles.go
```diff
@@ -38,13 +39,13 @@ func (r *spaceRolesManager) ensure(logger logr.Logger, nsTmplSet *toolchainv1alp
...
 		}
-		// compare last-applied vs spec to see if there's anything obsolete to do
-		if reflect.DeepEqual(nsTmplSet.Spec.SpaceRoles, lastAppliedSpaceRoles) {
-			logger.Info("no space role to update", "namespace", ns.Name)
-			continue
-		}
-		if err := r.setStatusUpdatingIfNotProvisioning(nsTmplSet); err != nil {
-			return false, err
+		// compare last-applied vs spec to see if there's anything obsolete
+		// note: we only set the NSTemplateSet status to `provisioning` if there are "configuration" changes,
```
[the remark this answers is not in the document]
xcoulon: good point!

controllers/nstemplateset/space_roles.go
xcoulon: good point!

controllers/nstemplateset/space_roles.go
```diff
@@ -61,33 +62,36 @@ func (r *spaceRolesManager) ensure(logger logr.Logger, nsTmplSet *toolchainv1alp
...
-		if err := r.Client.Update(context.TODO(), &ns); err != nil { // nolint:gosec
-			return false, r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusProvisionFailed, err,
-				"failed update namespace annotation")
+		if !reflect.DeepEqual(nsTmplSet.Spec.SpaceRoles, lastAppliedSpaceRoles) {
+			// store the space roles in an annotation at the namespace level, so we know what was applied and how to deal with
+			// diffs when the space roles are changed (users added or removed, etc.)
+			sr, err := json.Marshal(nsTmplSet.Spec.SpaceRoles)
+			if err != nil {
+				return false, r.wrapErrorWithStatusUpdate(logger, nsTmplSet, r.setStatusProvisionFailed, err,
+					fmt.Sprintf("failed to marshal space roles to update '%s' annotation on namespace", toolchainv1alpha1.LastAppliedSpaceRolesAnnotationKey))
```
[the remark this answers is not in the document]
xcoulon: we could use `spew.Sdump` and include the result, but to be honest, I don't think it's ever going to fail: the `NSTemplateSetSpaceRole` type is pretty simple and the resource should come from Etcd, meaning, it's supposed to come from a previous serialization.

controllers/nstemplateset/space_roles.go
xcoulon: good point. Reverted in ef1c026

controllers/nstemplateset/space_roles.go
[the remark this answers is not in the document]
rajivnathan: Good point, yeah let's leave it the way it is then

rajivnathan (approved):

PR #364, opened 2022-06-02
Add resourcequotas to operator SA role

alexeykazakov:
Required by https://github.com/codeready-toolchain/host-operator/pull/648


sbryzak (approved):

xcoulon (approved):

MatousJobanek:
the test is failing because it cannot pull the latest image of host-operator. The reason is that the image wasn't built for the last Shane's commit because of the GH action issue https://github.com/codeready-toolchain/host-operator/runs/6703189044?check_suite_focus=true
https://github.com/actions/cache/issues/811

PR #365, opened 2022-06-06
Fix appstudio user and identity handling after user tier migration

rajivnathan:
Fixes the problem where User and Identity resources were being created again for AppStudio users. In order to check if a user is part of the appstudio tier we now need to check the NSTemplateSet.

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/567


alexeykazakov (approved):

rajivnathan:
Closing since we shouldn't create a dependency between UserAccount and NSTemplateSet

PR #369, opened 2022-06-15
Retain apply error message

rajivnathan:


ranakan19 (approved):
Looks Good! Thank you for catching this :)

PR #370, opened 2022-06-24
Workaround for DBaaSTenant resource updates

rajivnathan:
An error occurs when trying to promote a Dev Sandbox user from one tier to another that has the DBaaSTenant resource template. The "[vdbaastenant.kb.io](http://vdbaastenant.kb.io/)" admission webhook is blocking the update of DBaaSTenant resources and subsequently blocks the space promotion from proceeding. As a workaround the DBaaSTenant resources can be deleted and recreated instead of updated.

Example of error:
``` 
admission webhook "vdbaastenant.kb.io" denied the request: spec.inventoryNamespace: 
Invalid value: "rajivnathan-dev": the namespace rajivnathan-dev is already managed by tenant tenant-rajivnathan-dev, it cannot be managed by another tenant unable to update the resource
...
```

### DIFF
--- a/controllers/nstemplateset/cluster_resources.go
+++ b/controllers/nstemplateset/cluster_resources.go
@@ -2,6 +2,7 @@ package nstemplateset
 
 import (
 	"context"
+	"fmt"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	applycl "github.com/codeready-toolchain/toolchain-common/pkg/client"
@@ -101,6 +102,7 @@ var clusterResourceKinds = []toolchainObjectKind{
 			}
 			return applycl.SortObjectsByName(list), nil
 		}),
+
 	newToolchainObjectKind(
 		dbaasv1alpha1.GroupVersion.WithKind("DBaaSTenant"),
 		&dbaasv1alpha1.DBaaSTenant{},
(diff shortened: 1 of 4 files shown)


controllers/nstemplateset/cluster_resources.go
```diff
@@ -237,6 +239,23 @@ CurrentObjects:
 					if err := r.setStatusUpdatingIfNotProvisioning(nsTmplSet); err != nil {
 						return false, err
 					}
+
+					// Special handling of DBaaSTenants is required because an admission webhook "vdbaastenant.kb.io" denies updates. Delete them and allow them to be created again as a workaround.
+					if currentObject.GetObjectKind().GroupVersionKind().Kind == "DBaaSTenant" {
```
MatousJobanek, line 244: Do we know what kind of error we got for that? I saw the message, but I'm curious about the error type. Just thinking if we couldn't handle that case based on the error type rather than on GVK.

rajivnathan: We discussed in a separate thread: It's blocked by the DBaaS admission webhook. Not sure we would want to make it a generic behaviour. In the DBaaS case, they have mentioned it won't break functionality to delete and recreate the resource but that may not be the same for other resources and operators. So we'll move forward with a temp workaround if we cannot get a fix from DBaaS soon

rajivnathan:
It looks like this has been fixed on the DBaaS side so this workaround is no longer needed.

PR #371, opened 2022-07-12
Handle migration from CRW to Dev Spaces

rajivnathan:
Related Issue: https://issues.redhat.com/browse/CRT-1562

- updates the `lookupAndDeleteCheUser` func to check whether the MemberOperatorConfig's Che Namespace and Route match `crw` and `devspaces` respectively. If the configuration matches then it will enter the DevSpaces deletion flow which will look up the User resource in order to get the UID. It then calls the che-db-cleaner service to delete the user using the UID.
- also updates the member status to check if we are in DevSpaces mode and if so it will skip the legacy Che User API checks  since DevSpaces does not have such APIs anymore
- no updates to e2e tests since this requires DevSpaces installation and our e2e environment does not have it but unit tests have good enough coverage

### DIFF
--- a/controllers/memberstatus/memberstatus_controller.go
+++ b/controllers/memberstatus/memberstatus_controller.go
@@ -295,13 +295,13 @@ func (r *Reconciler) cheHandleStatus(reqLogger logr.Logger, memberStatus *toolch
 	}
 
 	if !r.isCheAdminUserConfigured(config) {
-		err := fmt.Errorf("Che admin user credentials are not configured but Che user deletion is enabled")
+		err := fmt.Errorf("the Che admin user credentials are not configured but Che user deletion is enabled")
 		errCondition := status.NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusMemberStatusCheAdminUserNotConfiguredReason, err.Error())
 		memberStatus.Status.Che.Conditions = []toolchainv1alpha1.Condition{*errCondition}
 		return err
 	}
 
-	// Get che route for testing user API
+	// Ensure it's possible to construct the che URL for using the Che user API
 	if _, err := r.cheDashboardURL(config); err != nil {
 		wrappedErr := errs.Wrapf(err, "Che dashboard URL unavailable but Che user deletion is enabled")
 		errCondition := status.NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusMemberStatusCheRouteUnavailableReason, wrappedErr.Error())
@@ -309,11 +309,13 @@ func (r *Reconciler) cheHandleStatus(reqLogger logr.Logger, memberStatus *toolch
 		return err
 	}
 
-	// User API check
-	if err := r.CheClient.UserAPICheck(); err != nil {
-		errCondition := status.NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusMemberStatusCheUserAPICheckFailedReason, err.Error())
-		memberStatus.Status.Che.Conditions = []toolchainv1alpha1.Condition{*errCondition}
-		return err
+	// User API check (not applicable after migration from Che to Dev Spaces)
+	if config.Che().Namespace() != "crw" && config.Che().RouteName() != "devspaces" {
+		if err := r.CheClient.UserAPICheck(); err != nil {
+			errCondition := status.NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusMemberStatusCheUserAPICheckFailedReason, err.Error())
+			memberStatus.Status.Che.Conditions = []toolchainv1alpha1.Condition{*errCondition}
+			return err
+		}
 	}
 
 	readyCondition := status.NewComponentReadyCondition(toolchainv1alpha1.ToolchainStatusMemberStatusCheReadyReason)
--- a/controllers/useraccount/useraccount_controller.go
+++ b/controllers/useraccount/useraccount_controller.go
@@ -8,6 +8,7 @@ import (
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
 	membercfg "github.com/codeready-toolchain/member-operator/controllers/memberoperatorconfig"
 	"github.com/codeready-toolchain/member-operator/pkg/che"
+	"github.com/codeready-toolchain/member-operator/pkg/utils/user"
 	commoncontroller "github.com/codeready-toolchain/toolchain-common/controllers"
 	"github.com/codeready-toolchain/toolchain-common/pkg/condition"
 	commonconfig "github.com/codeready-toolchain/toolchain-common/pkg/configuration"
@@ -427,24 +428,23 @@ func (r *Reconciler) deleteIdentityAndUser(logger logr.Logger, userAcc *toolchai
 // if the user existed and something wrong happened. If the users don't exist,
 // this func returns `false, nil`
 func (r *Reconciler) deleteUser(logger logr.Logger, userAcc *toolchainv1alpha1.UserAccount) (bool, error) {
-	userList := &userv1.UserList{}
-	err := r.Client.List(context.TODO(), userList, listByOwnerLabel(userAcc.Name))
+	userList, err := user.GetUsersByOwnerName(r.Client, userAcc.Name)
 	if err != nil {
 		return false, err
 	}
 
-	if len(userList.Items) == 0 {
+	if len(userList) == 0 {
 		return false, nil
 	}
 
 	logger.Info("deleting the User resources")
 
 	// Delete User associated with UserAccount
-	if err := r.Client.Delete(context.TODO(), &userList.Items[0]); err != nil {
+	if err := r.Client.Delete(context.TODO(), &userList[0]); err != nil {
 		return false, err
 	}
 	// Return here, as deleting the user should cause another reconcile of the UserAccount
-	logger.Info(fmt.Sprintf("deleted User resource [%s]", userList.Items[0].Name))
+	logger.Info(fmt.Sprintf("deleted User resource [%s]", userList[0].Name))
 	return true, nil
 }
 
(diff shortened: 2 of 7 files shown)


pkg/utils/user/user.go
```diff
@@ -0,0 +1,22 @@
...
+
+	userv1 "github.com/openshift/api/user/v1"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetUsersByOwnerName gets the user resources by matching owner label.
+func GetUsersByOwnerName(cl client.Client, owner string) ([]userv1.User, error) {
+	userList := &userv1.UserList{}
```
xcoulon, line 14: Could you add some tests for this utility func?

pkg/che/che.go
```diff
@@ -204,3 +205,39 @@ func readCheUserFromJSON(jsonString string) (*User, error) {
...
+		return errors.Wrapf(err, "unable to delete Dev Spaces user with ID '%s'", userID)
+	}
+	defer rest.CloseResponse(res)
+	if res.StatusCode != http.StatusOK && res.StatusCode != http.StatusNotFound {
+		resBody, readError := rest.ReadBody(res.Body)
+		if readError != nil {
+			log.Error(readError, "error while reading body of the delete Dev Spaces user response")
+		}
+		err = errors.Errorf("unable to delete Dev Spaces user with ID '%s', Response status: '%s' Body: '%s'", userID, res.Status, resBody)
+	} else if res.StatusCode == http.StatusNotFound {
```
xcoulon, line 239: you could probably switch the `if`/`else if` statements to simplify the code, or even use a `switch` statement

pkg/che/che.go
```diff
@@ -204,3 +205,39 @@ func readCheUserFromJSON(jsonString string) (*User, error) {
...
+		log.Info("The Dev Spaces user was not deleted because it wasn't found", "userID", userID)
+	}
+	return err
```
xcoulon, line 242: `err` should be `nil` at this point, shouldn't it?

controllers/memberstatus/memberstatus_controller.go
```diff
@@ -295,25 +295,27 @@ func (r *Reconciler) cheHandleStatus(reqLogger logr.Logger, memberStatus *toolch
...
 		return err
 	}
 
-	// User API check
-	if err := r.CheClient.UserAPICheck(); err != nil {
-		errCondition := status.NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusMemberStatusCheUserAPICheckFailedReason, err.Error())
-		memberStatus.Status.Che.Conditions = []toolchainv1alpha1.Condition{*errCondition}
-		return err
+	// User API check (not applicable after migration from Che to Dev Spaces)
+	if config.Che().Namespace() != "crw" && config.Che().RouteName() != "devspaces" {
```
xcoulon, line 313: could we have something like `config.Che().IsDevSpaces()` to simplify the code/abstract the logic here?

controllers/useraccount/useraccount_controller.go
```diff
@@ -427,24 +428,23 @@ func (r *Reconciler) deleteIdentityAndUser(logger logr.Logger, userAcc *toolchai
 // if the user existed and something wrong happened. If the users don't exist,
 // this func returns `false, nil`
 func (r *Reconciler) deleteUser(logger logr.Logger, userAcc *toolchainv1alpha1.UserAccount) (bool, error) {
-	userList := &userv1.UserList{}
-	err := r.Client.List(context.TODO(), userList, listByOwnerLabel(userAcc.Name))
+	userList, err := user.GetUsersByOwnerName(r.Client, userAcc.Name)
```
xcoulon, line 431: from the code below, it seems like we only treat the first entry in `userList`, so should this `GetUsersByOwnerName` func be renamed `GetUserByOwnerName` instead, and return a single user? and maybe an error if multiple users match?

pkg/utils/user/user.go
rajivnathan: Actually, since it was only used by the useraccount controller I moved it to a local function there and the test coverage is good as well. Done in 55da18804f2cf11841825cb7a3866bfc9813813d

pkg/che/che.go
rajivnathan: I'm not sure how switching would simplify it. Do you have an example?

pkg/che/che.go
rajivnathan: It may not be. For example, in the condition `if res.StatusCode != http.StatusOK && res.StatusCode != http.StatusNotFound {` then we set `err` but in the `else if` case we do not

controllers/memberstatus/memberstatus_controller.go
rajivnathan: That's a good suggestion thanks! Done in d9f8ea8d5cc43844df8c6f2c5f2f27c1e4a0dc69 and 2a51ae5758897b4faabb53c0fe843bcad6e1b677

pkg/che/che.go
xcoulon: something like this:
```
switch res.StatusCode {
case http.StatusOK
  ...
case http.StatusNotFound
  ...
default:
  ...
}
```

pkg/utils/user/user.go
xcoulon: ok, thanks

pkg/che/che.go
xcoulon: that being said, it's just a suggestion. Feel free to keep it as-is if you think it's better ;)

controllers/useraccount/useraccount_controller.go
rajivnathan: It's a good suggestion but when I tried to change it I realized this function should return `false, nil` if no users are found so then we would have to have a check like `if err.Error() != "no users found" {` so I guess it wouldn't really simplify things much because of this. So maybe we can keep it as is? WDYT?

pkg/che/che.go
rajivnathan: Actually, you're right the switch statement is more readable. Done in 0e60c78ad3da40600118153d2f535388f1d3dfec Thanks for the suggestion!

rajivnathan:
@xcoulon I think I've addressed all of your comments, could you please take another look?

controllers/useraccount/useraccount_controller.go
```diff
@@ -615,6 +614,14 @@ func (r *Reconciler) lookupAndDeleteCheUser(logger logr.Logger, config membercfg
 		return nil
 	}
 
+	if config.Che().IsDevSpacesMode() {
+		err := r.deleteDevSpacesUser(logger, userAcc)
+		if err != nil {
+			logger.Error(err, "DevSpaces user deletion failed", "user account", userAcc.Name)
+		}
+		return nil // log error but do not fail deletion if the DevSpaces db cleanup failed, treat it as a best effort
```
ibuziuk, line 622: if deletion failed and the record is still in the database this would mean that after reprovisioning user will not be able to use the database. Can we stop the user deprovioning in this case and log the error instead?

pkg/che/che_test.go
```diff
@@ -525,6 +525,73 @@ func TestCheRequest(t *testing.T) {
...
+			k8sClient:  cl,
+		}
+
+		// when
+		defer gock.OffAll()
+		gock.New(dbCleanerURL).
+			Delete("johnsmith456").
+			Persist().
+			Reply(200)
+		err := cheClient.DevSpacesDBCleanerDelete("johnsmith456")
```
ibuziuk, line 547: this is ok for tests, but the expected input is valid UUID (input like that would not pass the validation on the db cleaner end)

controllers/memberoperatorconfig/configuration.go
```diff
@@ -171,6 +171,10 @@ func (a CheConfig) RouteName() string {
 	return commonconfig.GetString(a.che.RouteName, "codeready")
 }
 
+func (a CheConfig) IsDevSpacesMode() bool {
+	return a.Namespace() == "crw" && a.RouteName() == "devspaces"
```
ibuziuk, line 175: I would recommend checking the existence of the `codeready-workspaces-operator` in the first place. Basically, if the managed addon namespace still exists that would mean that CRW was not fully uninstalled from the cluster via OCM according to the migration SOP.

controllers/memberstatus/memberstatus_controller.go
```diff
@@ -295,25 +295,27 @@ func (r *Reconciler) cheHandleStatus(reqLogger logr.Logger, memberStatus *toolch
 	}
 
 	if !r.isCheAdminUserConfigured(config) {
-		err := fmt.Errorf("Che admin user credentials are not configured but Che user deletion is enabled")
+		err := fmt.Errorf("the Che admin user credentials are not configured but Che user deletion is enabled")
 		errCondition := status.NewComponentErrorCondition(toolchainv1alpha1.ToolchainStatusMemberStatusCheAdminUserNotConfiguredReason, err.Error())
 		memberStatus.Status.Che.Conditions = []toolchainv1alpha1.Condition{*errCondition}
 		return err
 	}
```
ibuziuk, line 302: This check is also relevant only for the CRW (not DevSpaces) I believe

pkg/che/che_test.go
rajivnathan: ack, since we get the UID directly from the User resource in the code it should be fine

controllers/memberoperatorconfig/configuration.go
rajivnathan: Since the migration is a one time thing maybe we can just confirm the managed addon has been removed and everything is ready before changing the sandbox configuration?

controllers/useraccount/useraccount_controller.go
rajivnathan: Done in 71bf1730fb24cb6d4ad626cf50ccb1ed76ab7ea6

ibuziuk (approved):

controllers/useraccount/useraccount_controller.go
xcoulon: ok, let's keep it as-is for now, then.

xcoulon (approved):

controllers/memberoperatorconfig/configuration.go
MatousJobanek: @ibuziuk don't forget that the `codeready-workspaces-operator` got stuck in the `terminating` state in the stage cluster. So checking that namespace wouldn't help at all.

MatousJobanek (approved):

PR #372, opened 2022-07-14
Namespace update fix

rajivnathan:
Fixes https://issues.redhat.com/browse/CRT-1457

Updates the `ensureNamespace` func to check whether the namespace is missing any labels that are specified in its corresponding namespace template so that if any are missing the template will be applied.

Note: For existing namespaces, this will only handle new labels or changed label values. If a label is removed from a template the label will *not* be removed from the namespace. This is intentional because in order to handle removal correctly, the previous template would need to be known and compared along with the other labels that are added automatically by the controller. That is outside the scope of this fix.

### DIFF
--- a/controllers/nstemplateset/namespaces.go
+++ b/controllers/nstemplateset/namespaces.go
@@ -76,10 +76,10 @@ func (r *namespacesManager) ensure(logger logr.Logger, nsTmplSet *toolchainv1alp
 func (r *namespacesManager) ensureNamespace(logger logr.Logger, nsTmplSet *toolchainv1alpha1.NSTemplateSet, tierTemplate *tierTemplate, userNamespace *corev1.Namespace) error {
 	logger.Info("ensuring namespace", "namespace", tierTemplate.typeName, "tier", nsTmplSet.Spec.TierName)
 
-	namespaceNeedsUpdate := false
+	createOrUpdateNamespace := false
 	if userNamespace == nil {
 		// userNamespace does not exist, need to create the namespace
-		namespaceNeedsUpdate = true
+		createOrUpdateNamespace = true
 		logger.Info("namespace needs to be created")
 	} else {
 		// userNamespace exists, check if the namespace needs to be updated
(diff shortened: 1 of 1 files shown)


controllers/nstemplateset/namespaces.go
```diff
@@ -76,13 +76,72 @@ func (r *namespacesManager) ensure(logger logr.Logger, nsTmplSet *toolchainv1alp
...
+	}
+
+	// create namespace before creating inner resources because creating the namespace may take some time
+	if namespaceNeedsUpdate {
 		return r.ensureNamespaceResource(logger, nsTmplSet, tierTemplate)
 	}
 	return r.ensureInnerNamespaceResources(logger, nsTmplSet, tierTemplate, userNamespace)
 }
 
+// namespaceHasExpectedLabelsFromTemplate checks if the namespace has the expected labels from the template object
```
MatousJobanek, line 101: could you please just add a comment that it checks only if the namespace has the labels and doesn't check those that could have been removed?

controllers/nstemplateset/namespaces.go
```diff
@@ -76,13 +76,72 @@ func (r *namespacesManager) ensure(logger logr.Logger, nsTmplSet *toolchainv1alp
 func (r *namespacesManager) ensureNamespace(logger logr.Logger, nsTmplSet *toolchainv1alpha1.NSTemplateSet, tierTemplate *tierTemplate, userNamespace *corev1.Namespace) error {
 	logger.Info("ensuring namespace", "namespace", tierTemplate.typeName, "tier", nsTmplSet.Spec.TierName)
 
-	// create namespace before created inner resources because creating the namespace may take some time
+	namespaceNeedsUpdate := false
```
MatousJobanek, line 79: maybe 
```suggestion
	createOrUpdateNamespace := false
```
since it applies also to creation

MatousJobanek (approved):

controllers/nstemplateset/namespaces.go
rajivnathan: Done in 035cdca7d68d85445177585f27f3d28cdadb2dcd

controllers/nstemplateset/namespaces.go
rajivnathan: Done in 035cdca7d68d85445177585f27f3d28cdadb2dcd

PR #374, opened 2022-08-05
ensure namespace created

ranakan19:
fix wrt CRT1471

### DIFF
--- a/controllers/nstemplateset/namespaces.go
+++ b/controllers/nstemplateset/namespaces.go
@@ -5,6 +5,9 @@ import (
 	"fmt"
 	"sort"
 
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+
 	rbac "k8s.io/api/rbac/v1"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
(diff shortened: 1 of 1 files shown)


controllers/nstemplateset/namespaces.go
```diff
@@ -154,9 +157,18 @@ func (r *namespacesManager) ensureInnerNamespaceResources(logger logr.Logger, ns
...
+	// ensure the namespace created is ready
+	userNamespaceCreated := &corev1.Namespace{}
+	if err := r.Client.Get(context.TODO(), types.NamespacedName{Name: namespace.Name}, userNamespaceCreated); err != nil {
+		if errors.IsNotFound(err) {
+			return fmt.Errorf(`Unable to get the usernamespace '%s'`, namespace.Name)
+		}
+	}
+	if userNamespaceCreated.Status.Phase != "Active" {
+		return fmt.Errorf(`Namespace %s is in %s Phase`, userNamespaceCreated.Name, userNamespaceCreated.Status.Phase)
+	}
```
MatousJobanek, line 169: why do we need this? We already check if the namespace is active or not
https://github.com/codeready-toolchain/member-operator/blob/218698f00f6cb2d41274bd15100677e651be92e5/controllers/nstemplateset/namespaces.go#L301
if it's not active, then we cannot apply the inner-namespace objects

ranakan19:
not needed as pointed out by matous

PR #378, opened 2022-09-20
fix(#343): Add support for scale subresource

### LINKED ISSUE #343: Idler should support scale sub resources
The idler currently works on K8s built in resources such as Pods, Deployments and ReplicaSets to perform kill/scale-to-zero operations on long running user workloads. 

Yet a user workload's custom resource may also control the replicas via `scale` sub resource (https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource).

In this case scaling deployments and replica sets to zero won't have any effect on the replicas as the custom resource controller will immediately bring back the replicas. The idler will then kill pods and the scale sub resource brings them back immediately.

A sample of such a workload using the scale sub resource is the Camel K integration custom resource.
https://github.com/apache/camel-k/blob/main/config/crd/bases/camel.apache.org_integrations.yaml#L5911

In case the deployment is owned by a custom resource the idler should also support the scale sub resource to make sure that the user workload is stopped properly.

christophd:
Custom resources can use scale subresource to manage the replicas of deployments. In this case the idler needs to update the autoscalingv1.Scale resource in order to terminate the deployment (replicas=0).

Update the idler component to also support deployments managed by scale sub resources. In case a deployment is owned by a custom resource that holds a scale sub resource the idler will patch the scale resource instead of the deployment directly.

This makes sure that the idler is able to terminate deployments owned by scale resources, too.

At the moment this is being used by Camel K workloads such as Integration and KameletBinding custom resources.

Fixes #343

### DIFF
--- a/controllers/idler/idler_controller.go
+++ b/controllers/idler/idler_controller.go
@@ -10,6 +10,8 @@ import (
 	"github.com/codeready-toolchain/toolchain-common/pkg/condition"
 	notify "github.com/codeready-toolchain/toolchain-common/pkg/notification"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/scale"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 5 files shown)


christophd:
@astefanutti FYI maybe you can also have a look, thx

xcoulon (approved):
This looks great and also interesting: I wonder if we could use this `scalesClient` to also scale down the "regular" deployments, replicasets, statefulsets, etc.

alexeykazakov (approved):
Thank you for the contribution!
Ideally we would like to cover it by e2e tests too but that would be tricky because we do not have camel-k installed in our e2e tests environment.

astefanutti:
> @astefanutti FYI maybe you can also have a look, thx

@christophd awesome, thanks!

astefanutti:
> This looks great and also interesting: I wonder if we could use this `scalesClient` to also scale down the "regular" deployments, replicasets, statefulsets, etc.

@xcoulon yes, core workload resources have the scale sub-resource that can be use with the scale polymorphic client (kubernetes/kubernetes#49504). The tricky part is making sure to update the owning resource with a scale sub-resource.

PR #388, opened 2022-10-27
fix(#343): Support scale resources that do not have specReplicasPath field set

christophd:
The replicas field defined in specReplicasPath may be optional in the custom resource. In this case getting the scale resource fails with internal server error. In this case try to set a new scale resource with replicas=0. This fixes idling of Camel K integrations and bindings that do not explicitly have a `spec.replicas` field set and default to replicas=1.

Signed-off-by: Christoph Deppisch <cdeppisch@redhat.com>

### DIFF
--- a/controllers/idler/idler_controller.go
+++ b/controllers/idler/idler_controller.go
@@ -9,6 +9,7 @@ import (
 	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
 	"github.com/codeready-toolchain/toolchain-common/pkg/condition"
 	notify "github.com/codeready-toolchain/toolchain-common/pkg/notification"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/scale"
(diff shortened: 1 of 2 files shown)


controllers/idler/idler_controller.go
```diff
@@ -299,6 +300,24 @@ func (r *Reconciler) scaleDeploymentToZero(logger logr.Logger, namespace string,
 					return true, nil
 				}
 
+				return false, err
+			} else if errors.IsInternalError(err) { // Failed to get existing scale resource - try to set a new one
```
alexeykazakov, line 304: Is the internal error the only way to detect a missing optional replica field?
If so the let's at least log an infra message and include the original error message there too.

alexeykazakov (commented):
Could you add a unit test for this please.

controllers/idler/idler_controller.go
christophd: yes, unfortunately this is being reported as a 500 internal error `Internal error occurred: the spec replicas field \".spec.replicas\" does not exist`

I am afraid that this will be the default path for almost all resources with a scale sub resource as a missing replicas field is equal to the default replicas=1. Do we really want all of those errors to be logged then?

alexeykazakov: Ok. Let's keep it as is. But please add a test :)

christophd: I have added a unit test and I narrowed down the internal server error to have the specific error message saying that the spec.replicas field does not exist.

controllers/idler/idler_controller.go
```diff
@@ -299,6 +301,24 @@ func (r *Reconciler) scaleDeploymentToZero(logger logr.Logger, namespace string,
 					return true, nil
 				}
 
+				return false, err
+			} else if errors.IsInternalError(err) && strings.Contains(err.Error(), "the spec replicas field \".spec.replicas\" does not exist") {
```
alexeykazakov, line 305: What if they error message is changed in the future? Not sure we should rely on the message :( There is no any e2e test which would catch it. So just checking the error code is probably safer...

alexeykazakov (commented):
Thanks for the tests!

controllers/idler/idler_controller.go
christophd: yes true, the error message is subject to change so I have removed the check

alexeykazakov (approved):

PR #392, opened 2022-11-21
Update api version

mfrancisc:
see [codeready-toolchain/api#327](https://github.com/codeready-toolchain/api/pull/327)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/codeready-toolchain/member-operator
 
 require (
-	github.com/codeready-toolchain/api v0.0.0-20221102132326-7a7f4eb137db
+	github.com/codeready-toolchain/api v0.0.0-20221121081259-4f672e808b89
 	github.com/codeready-toolchain/toolchain-common v0.0.0-20221117081323-252a38460fa1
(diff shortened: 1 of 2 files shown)


MatousJobanek (approved):
this is usually unnecessary as you didn't remove nor change anything - you only added a new parameter that doesn't have any impact on member-operator nor registration-service.

mfrancisc:
Ok, thanks for clarifying it! I saw that api version was up to date in all repositories so I went updating it anyway.

So you confirm that it's fine for *-operator, registration-service and toolchain-e2e to depend on different versions of codready-toolchain/api ?

MatousJobanek:
> So you confirm that it's fine for *-operator, registration-service and toolchain-e2e to depend on different versions of codready-toolchain/api ?

as soon as the changes in the API repo don't have any impact on these repositories, nor the logic/functionality, then yes - it's totally fine if they are from time to time slightly behind :+1: 

However, if the change in the API:
* either breaks backward compatibility (renames or removes a field)
* or is not fully isolated (logic in other repositories may transitively rely on it)

then it's good to update the versions in other repos too.

PR #411, opened 2023-02-18
Set UserID and AccountID in User resource if they are set in UserAccount

sbryzak:
Propagate both the UserID and AccountID annotation values from `UserAccount` to `User` during reconciliation, if both values are set.  If the values are not set then delete the annotations on the `User`.

Fixes https://issues.redhat.com/browse/CRT-1718

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/670

### DIFF
--- a/controllers/useraccount/useraccount_controller.go
+++ b/controllers/useraccount/useraccount_controller.go
@@ -386,6 +386,34 @@ func setLabelsAndAnnotations(object metav1.Object, userAcc *toolchainv1alpha1.Us
 			object.SetAnnotations(annotations)
 			changed = true
 		}
+
+		annotations = object.GetAnnotations()
+		set := false
+		userID, ok := userAcc.Annotations[toolchainv1alpha1.SSOUserIDAnnotationKey]
+		if ok {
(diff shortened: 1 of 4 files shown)


mfrancisc (commented):
Looks good 👍  
but seems that the deletion of the annotations is not covered by the tests.

controllers/useraccount/useraccount_controller.go
```diff
@@ -386,6 +386,34 @@ func setLabelsAndAnnotations(object metav1.Object, userAcc *toolchainv1alpha1.Us
...
+		}
+
+		// Delete the UserID and AccountID annotations if they don't exist in the UserAccount
+		if !set && object.GetAnnotations() != nil {
+			annotations = object.GetAnnotations()
+			delete(annotations, toolchainv1alpha1.SSOUserIDAnnotationKey)
+			delete(annotations, toolchainv1alpha1.SSOAccountIDAnnotationKey)
+			object.SetAnnotations(annotations)
+			changed = true
+		}
```
alexeykazakov, line 416: If you believe we need it, then let's add a test for this :)

sbryzak: Test added

alexeykazakov (approved):

PR #415, opened 2023-03-01
refactor: skip idling when timeout is 0

xcoulon:
no need to report an error, since timeout=0 is valid for '___noidling' tiers

see also:
- https://github.com/codeready-toolchain/api/pull/349
- discussion on [Slack#forum-dev-sandbox](https://redhat-internal.slack.com/archives/CHK0J6HT6/p1677582613636039)

### DIFF
--- a/controllers/idler/idler_controller_test.go
+++ b/controllers/idler/idler_controller_test.go
@@ -343,28 +343,47 @@ func TestEnsureIdling(t *testing.T) {
 }
 
 func TestEnsureIdlingFailed(t *testing.T) {
-	t.Run("Fail if Idler.Spec.TimeoutSec is invalid", func(t *testing.T) {
-		assertInvalidTimeout := func(timeout int32) {
-			// given
-			idler := &toolchainv1alpha1.Idler{
-				ObjectMeta: metav1.ObjectMeta{
-					Name: "john-dev",
-				},
-				Spec: toolchainv1alpha1.IdlerSpec{TimeoutSeconds: timeout},
-			}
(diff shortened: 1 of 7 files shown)


controllers/idler/idler_controller_test.go
```diff
@@ -343,28 +343,47 @@ func TestEnsureIdling(t *testing.T) {
...
-			require.NoError(t, err)
-			assert.Equal(t, reconcile.Result{}, res)
-			memberoperatortest.AssertThatIdler(t, idler.Name, cl).HasConditions(memberoperatortest.FailedToIdle("timeoutSeconds should be bigger than 0"))
+		// when
+		res, err := reconciler.Reconcile(context.TODO(), req)
+
+		// then
+		require.NoError(t, err)
+		assert.Equal(t, reconcile.Result{}, res)
+		memberoperatortest.AssertThatIdler(t, idler.Name, cl).HasNoConditions()
```
mfrancisc, line 365: I was wondering if setting a condition like `NoIdling` in this case could make sense, if it makes it more clear when debugging/checking the objects in the cluster 🤷‍♂️

rajivnathan: +1

alexeykazakov (approved):

controllers/idler/idler_controller_test.go
xcoulon: done in 33b2b1c

mfrancisc: @xcoulon thanks for addressing it! 

And good catch regarding the mixed usage of `v1` and `metav1` package !

PR #417, opened 2023-03-08
Add Pendo integration to web console plugin

sbryzak:
This PR adds a webserver console plugin that serves both a health check endpoint (/api/v1/status) and a script server that serves the Pendo Javascript resources.

Fixes https://issues.redhat.com/browse/CRT-1723
Fixes https://issues.redhat.com/browse/CRT-1724
Fixes https://issues.redhat.com/browse/CRT-1725


cmd/consoleplugin/static/exposed-pendo-chunk-0f1e4648dfea45bc1052.min.js
```diff
@@ -0,0 +1 @@
+"use strict";(self.webpackChunkconsole_plugin_template=self.webpackChunkconsole_plugin_template||[]).push([[484],{248:(n,t,e)=>{e.r(t),e.d(t,{default:()=>c}),function(n,t,e,i,o){var a,s,c,d,l;for((o=n[i]=n[i]||{})._q=o._q||[],s=0,c=(a=["initialize","identify","updateOptions","pageLoad","track"]).length;s<c;++s)!function(n){o[n]=o[n]||function(){o._q[n===a[0]?"unshift":"push"]([n].concat([].slice.call(arguments,0)))}}(a[s]);(d=t.createElement(e)).async=!0,d.src="https://cdn.pendo.io/agent/static//pendo.js",(l=t.getElementsByTagName(e)[0]).parentNode.insertBefore(d,l)}(window,document,"script","pendo");const i="toolchain.dev.openshift.com/",o=`${i}sso-user-id`,a=`${i}sso-account-id`;let s=!1;const c=(n,t)=>{if("identify"===n){const{user:n}=t;if(n){const t=n.metadata.annotation?.[o],e=n.metadata.annotation?.[a];t&&window.pendo&&(s=!0,window.pendo[s?"identify":"initialize"]({visitor:{id:t},...e?{account:{id:e}}:null}))}}}}}]);
```
alexeykazakov, line 1: Looks like some generated file? How did you generate it?

sbryzak: @alexeykazakov these are generated from Christian's project at https://github.com/christianvogt/console-plugin-pendo.  It requires the nodejs tooling installed to build, I'm not sure if we want to introduce that kind of dependency to our build process but we can if required.  One way might be to include the actual JS/Typescript source and have Podman spin up a Node container to perform the build.

alexeykazakov: Yes, I would like to have the plugin source code in our codeready-toolchain GitHub org.

I think the easiest option would be to move the actual JS/Typescript source here, in member-operator repo. Then add a new make target to generate the minimized JS out of it using nodejs.

And there are two options after that:
1. include that make target in our CI/CD targets so this min.js is always auto-generated (so we don't commit it to the repo)
2. keep this make target standalone and require developers to run the target manually every time the typescript source code is changed and commit it to GitHub. Disadvantage is that it's another manual step and it's hard to verify the proposed change/PR because the min.js is not readable (you have to trust the PR author that he/she did the generation properly) which makes some kind of e2e tests more critical in this case but it's also much harder to introduce any e2e tests with this option.

sbryzak: No problem, I'll take care of that.  I think option 1 might be the best, simply because it doesn't have the disadvantages of option 2.

cmd/consoleplugin/main.go
```diff
@@ -1,5 +1,125 @@
...
+		Addr:    ":8080",
+		Handler: mux,
+	}
+
+	setupLog.Info("Web Console Plugin server configured.")
+
+	go func() {
+		setupLog.Info("Listening...")
+
+		if err := consolePluginServer.ListenAndServeTLS("/etc/webhook/certs/"+cert.ServerCert, "/etc/webhook/certs/"+cert.ServerKey); err != nil {
```
alexeykazakov, line 110: I think we shouldn't start a TLS service. I guess you copied it from the web hook service but it's different. Admition web hook are special services in Kube. Here we need a simple http service.
Unless web console does require an internal service running with TLS... Which I doubt.

pkg/consoleplugin/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,19 @@
+package healthcheck
+
+import (
+	"net/http"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var (
+	log = logf.Log.WithName("web_console_health_check_webhook")
```
alexeykazakov, line 9: Copy-paste. It's not a webhook.

pkg/consoleplugin/scriptserver/scriptserver.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"embed"
+	"net/http"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sync"
+)
+
+var (
+	log = logf.Log.WithName("web_console_script_server_webhook")
```
alexeykazakov, line 11: Same here.

alexeykazakov (commented):
We also need more unit tests. And some e2e tests as well.

pkg/consoleplugin/scriptserver/scriptserver.go
sbryzak: fixed

pkg/consoleplugin/healthcheck/healthcheck.go
sbryzak: fixed

xcoulon:
@sbryzak @alexeykazakov out of curiosity: what's the rational for embedding an HTTP server into the member operator vs having a standalone service such as registration service? 
For example, do we need to run the whole e2e suite on OpenShift CI when the pendo plugin is updated?

alexeykazakov:
It's much easier to piggy-back on the member operator OLM bundle then creating a completely separate service which would have to be deployed and managed by the member operator anyway. So it's just because of the urgency of this work.

And yes, I think we need to run e2e tests every time we change the plugin service.

pkg/consoleplugin/scriptserver/scriptserver_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+	s.HandleScriptRequest(resp, req)
+
+	body, err := io.ReadAll(resp.Body)
+	require.NoError(t, err)
+
+	require.Len(t, body, 1896)
+	require.True(t, strings.HasPrefix(string(body), "// initialize pendo"))
+
+}
```
MatousJobanek, line 25: could you please cover also the other files, and some (or all) the error cases? you can do it in for cycle and ideally multiple calls in parallel to verify that the locks and cache work as expected.

pkg/consoleplugin/scriptserver/scriptserver.go
```diff
@@ -0,0 +1,77 @@
...
+
+	fileData, err := staticFiles.ReadFile("static" + path)
+	if err != nil {
+		return nil, err
+	}
+
+	s.cache[path] = fileData
+
+	return fileData, nil
+}
```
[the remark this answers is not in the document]
sbryzak: We will almost certainly need to perform some string replacement for the file contents so this is in preparation for that.

pkg/consoleplugin/scriptserver/static/plugin-manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "console-plugin-template",
```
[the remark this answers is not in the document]
christianvogt: I suggest updating the plugin-manifest.json and rebuilding the plugin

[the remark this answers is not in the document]
alexeykazakov: Fixed. Thanks.

pkg/consoleplugin/scriptserver/static/pendo.ts
```diff
@@ -0,0 +1,56 @@
...
+const SSO_ACCOUNT_ID = `${PREFIX}sso-account-id`;
+
+let initialized = false;
+
+export default (eventType: string, properties?: any) => {
+    if (eventType === 'identify') {
+        const { user } = properties;
+        if (user) {
+            const ssoUserId = user.metadata.annotation?.[SSO_USER_ID];
+            const ssoAccountId = user.metadata.annotation?.[SSO_ACCOUNT_ID];
```
christianvogt, line 38:
```suggestion
            const ssoUserId = user.metadata.annotations?.[SSO_USER_ID];
            const ssoAccountId = user.metadata.annotations?.[SSO_ACCOUNT_ID];
```

pkg/consoleplugin/scriptserver/static/exposed-pendo-chunk-0f1e4648dfea45bc1052.min.js
```diff
@@ -0,0 +1 @@
+"use strict";(self.webpackChunktoolchain_member_web_console_plugin=self.webpackChunktoolchain_member_web_console_plugin||[]).push([[484],{248:(n,t,e)=>{e.r(t),e.d(t,{default:()=>c}),function(n,t,e,i,o){var a,s,c,d,l;for((o=n[i]=n[i]||{})._q=o._q||[],s=0,c=(a=["initialize","identify","updateOptions","pageLoad","track"]).length;s<c;++s)!function(n){o[n]=o[n]||function(){o._q[n===a[0]?"unshift":"push"]([n].concat([].slice.call(arguments,0)))}}(a[s]);(d=t.createElement(e)).async=!0,d.src="https://cdn.pendo.io/agent/static/{INSERT_KEY_HERE}/pendo.js",(l=t.getElementsByTagName(e)[0]).parentNode.insertBefore(d,l)}(window,document,"script","pendo");const i="toolchain.dev.openshift.com/",o=`${i}sso-user-id`,a=`${i}sso-account-id`;let s=!1;const c=(n,t)=>{if("identify"===n){const{user:n}=t;if(n){const t=n.metadata.annotation?.[o],e=n.metadata.annotation?.[a];t&&window.pendo&&(s=!0,window.pendo[s?"identify":"initialize"]({visitor:{id:t},...e?{account:{id:e}}:null}))}}}}}]);
```
christianvogt, line 1:
```suggestion
"use strict";(self.webpackChunktoolchain_member_web_console_plugin=self.webpackChunktoolchain_member_web_console_plugin||[]).push([[484],{248:(n,t,e)=>{e.r(t),e.d(t,{default:()=>c}),function(n,t,e,i,o){var a,s,c,d,l;for((o=n[i]=n[i]||{})._q=o._q||[],s=0,c=(a=["initialize","identify","updateOptions","pageLoad","track"]).length;s<c;++s)!function(n){o[n]=o[n]||function(){o._q[n===a[0]?"unshift":"push"]([n].concat([].slice.call(arguments,0)))}}(a[s]);(d=t.createElement(e)).async=!0,d.src="https://cdn.pendo.io/agent/static/{INSERT_KEY_HERE}/pendo.js",(l=t.getElementsByTagName(e)[0]).parentNode.insertBefore(d,l)}(window,document,"script","pendo");const i="toolchain.dev.openshift.com/",o=`${i}sso-user-id`,a=`${i}sso-account-id`;let s=!1;const c=(n,t)=>{if("identify"===n){const{user:n}=t;if(n){const t=n.metadata.annotations?.[o],e=n.metadata.annotations?.[a];t&&window.pendo&&(s=!0,window.pendo[s?"identify":"initialize"]({visitor:{id:t},...e?{account:{id:e}}:null}))}}}}}]);
```

pkg/consoleplugin/scriptserver/static/exposed-pendo-chunk-0f1e4648dfea45bc1052.min.js
```diff
@@ -0,0 +1 @@
+"use strict";(self.webpackChunktoolchain_member_web_console_plugin=self.webpackChunktoolchain_member_web_console_plugin||[]).push([[484],{248:(n,t,e)=>{e.r(t),e.d(t,{default:()=>c}),function(n,t,e,i,o){var a,s,c,d,l;for((o=n[i]=n[i]||{})._q=o._q||[],s=0,c=(a=["initialize","identify","updateOptions","pageLoad","track"]).length;s<c;++s)!function(n){o[n]=o[n]||function(){o._q[n===a[0]?"unshift":"push"]([n].concat([].slice.call(arguments,0)))}}(a[s]);(d=t.createElement(e)).async=!0,d.src="https://cdn.pendo.io/agent/static/{INSERT_KEY_HERE}/pendo.js",(l=t.getElementsByTagName(e)[0]).parentNode.insertBefore(d,l)}(window,document,"script","pendo");const i="toolchain.dev.openshift.com/",o=`${i}sso-user-id`,a=`${i}sso-account-id`;let s=!1;const c=(n,t)=>{if("identify"===n){const{user:n}=t;if(n){const t=n.metadata.annotations?.[o],e=n.metadata.annotations?.[a];t&&window.pendo&&(s=!0,window.pendo[s?"identify":"initialize"]({visitor:{id:t},...e?{account:{id:e}}:null}))}}}}}]);
```
christianvogt, line 1:
```suggestion
"use strict";(self.webpackChunktoolchain_member_web_console_plugin=self.webpackChunktoolchain_member_web_console_plugin||[]).push([[484],{248:(n,t,e)=>{e.r(t),e.d(t,{default:()=>c}),function(n,t,e,i,o){var a,s,c,d,l;for((o=n[i]=n[i]||{})._q=o._q||[],s=0,c=(a=["initialize","identify","updateOptions","pageLoad","track"]).length;s<c;++s)!function(n){o[n]=o[n]||function(){o._q[n===a[0]?"unshift":"push"]([n].concat([].slice.call(arguments,0)))}}(a[s]);(d=t.createElement(e)).async=!0,d.src="https://cdn.pendo.io/agent/static/{INSERT_KEY_HERE}/pendo.js",(l=t.getElementsByTagName(e)[0]).parentNode.insertBefore(d,l)}(window,document,"script","pendo");const i="toolchain.dev.openshift.com/",o=`${i}sso-user-id`,a=`${i}sso-account-id`;let s=!1;const c=(n,t)=>{if("identify"===n){const{user:n}=t;if(n){const t=n.metadata.annotations?.[o],e=n.metadata.annotations?.[a];t&&window.pendo&&(window.pendo[s?"identify":"initialize"]({visitor:{id:t},...e?{account:{id:e}}:null}),s=!0)}}}}}]);
```

pkg/consoleplugin/consolepluginserver.go
```diff
@@ -0,0 +1,69 @@
...
+
+	for _, opt := range options {
+		s.options = append(s.options, string(opt))
+	}
+
+	s.mux = http.NewServeMux()
+	ss := scriptserver.NewScriptServer()
+	s.mux.HandleFunc("/", ss.HandleScriptRequest)
+	s.svr = &http.Server{ //nolint:gosec
+		Addr:    ":9443",
```
mfrancisc, line 40: small detail - should this port to be configured ? maybe setting it from env variables could make it more flexible.

deploy/consoleplugin/member-operator-console-plugin.yaml
```diff
@@ -3,125 +3,136 @@ kind: Template
...
+          startupProbe:
+            failureThreshold: 180
+            httpGet:
+              path: /api/v1/status
+              port: 8080
+              scheme: HTTP
+            initialDelaySeconds: 1
+            periodSeconds: 1
+            successThreshold: 1
+            timeoutSeconds: 1
```
mfrancisc, line 100: I've noticed that application is exposed on port `9443` (https I guess) , I was wondering if we should have the probes for that port as well ?

pkg/consoleplugin/consolepluginserver.go
alexeykazakov: It's hard to make it truly configurable because the port is also defined in the Service manifest. I think it's fine to hardcode it.

deploy/consoleplugin/member-operator-console-plugin.yaml
alexeykazakov: @sbryzak I wonder if we can just replace the `:8080/api/v1/status` by `:9443/plugin-manifest.json`... Let's just use the manifest as the health check.

pkg/consoleplugin/scriptserver/scriptserver.go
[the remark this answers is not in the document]
alexeykazakov: Just for the context. We need to insert the Pendo key into the JS file. The key is supposed to be loaded from our configuration. Currently it's hardcoded: https://github.com/codeready-toolchain/member-operator/pull/417/commits/6f109baabe662a310f3d944836c0dfcff8ddf0d2 But we need to move it start using configuration in this PR.

Ideally we should apply the current configuration without restarting the service but it's fine for now to load configuration when the service is being started IMO. We can improve later.

deploy/consoleplugin/member-operator-console-plugin.yaml
alexeykazakov: I just tested it and it works fine. The only problem I see is the logs... Since it's the same endpoint now it's going to be hard to hide the status check pings in the logs.

pkg/consoleplugin/scriptserver/scriptserver.go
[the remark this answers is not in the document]
sbryzak: Just to add to this, it's non-trivial to populate a map at startup time because the actual content filenames are dynamic, for example `exposed-pendo-chunk-0f1e4648dfea45bc1052.min.js`.  This filename would change if we were to regenerate it after making changes to the source JS, meaning we would have to then update the populating of the map cache code, thereby introducing a vector for human error occurring.

controllers/memberoperatorconfig/configuration.go
```diff
@@ -232,3 +236,11 @@ type WebhookConfig struct {
...
 	return commonconfig.GetBool(a.w.Deploy, true)
 }
+
+type WebConsolePluginConfig struct {
+	w toolchainv1alpha1.WebConsolePlugin
+}
+
+func (a WebConsolePluginConfig) Deploy() bool {
+	return commonconfig.GetBool(a.w.Deploy, true)
+}
```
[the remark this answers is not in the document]
sbryzak: Added.

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
```diff
@@ -64,6 +65,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		return reconcile.Result{}, err
 	}
 
+	if err := r.handleWebConsolePluginDeploy(reqLogger, crtConfig, request.Namespace); err != nil {
+		return reconcile.Result{}, err
+	}
```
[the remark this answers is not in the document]
sbryzak: Tests added

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
```diff
@@ -2,6 +2,7 @@ package memberoperatorconfig
 
 import (
 	"context"
+	deploy2 "github.com/codeready-toolchain/member-operator/pkg/consoleplugin/deploy"
```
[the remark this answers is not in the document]
sbryzak: fixed

pkg/consoleplugin/scriptserver/scriptserver_test.go
sbryzak: All non-dynamically generated files are now included in the tests in some form.

controllers/memberoperatorconfig/configuration_test.go
```diff
@@ -318,6 +318,33 @@ func TestWebhook(t *testing.T) {
...
+		cfg := commonconfig.NewMemberOperatorConfigWithReset(t)
+		memberOperatorCfg := Configuration{cfg: &cfg.Spec}
+
+		assert.False(t, memberOperatorCfg.WebConsolePlugin().Deploy())
+	})
+	t.Run("non-default", func(t *testing.T) {
+		cfg := commonconfig.NewMemberOperatorConfigWithReset(t, testconfig.WebConsolePlugin().Deploy(false))
+		memberOperatorCfg := Configuration{cfg: &cfg.Spec}
+
+		assert.False(t, memberOperatorCfg.WebConsolePlugin().Deploy())
```
alexeykazakov, line 332:
```suggestion
		cfg := commonconfig.NewMemberOperatorConfigWithReset(t, testconfig.WebConsolePlugin().Deploy(true))
		memberOperatorCfg := Configuration{cfg: &cfg.Spec}

		assert.True(t, memberOperatorCfg.WebConsolePlugin().Deploy())
```

make/run-cicd-script.mk
```diff
@@ -1,4 +1,5 @@
-OWNER_AND_BRANCH_LOCATION=codeready-toolchain/toolchain-cicd/master
+#OWNER_AND_BRANCH_LOCATION=codeready-toolchain/toolchain-cicd/master
+OWNER_AND_BRANCH_LOCATION=sbryzak/toolchain-cicd/add_consoleplugin_image
```
alexeykazakov, line 2:
```suggestion
OWNER_AND_BRANCH_LOCATION=codeready-toolchain/toolchain-cicd/master
```

alexeykazakov (commented):
Overall looks good.
However I see that the service crashloops if enabled:

```{"level":"info","ts":"2023-03-21T02:13:35.485Z","msg":"Waited for 1.012199691s due to client-side throttling, not priority and fairness, request: GET:https://172.30.0.1:443/apis/autoscaling/v2beta1?timeout=32s\n"}
{"level":"error","ts":"2023-03-21T02:13:38.798Z","logger":"configuration","msg":"failed to retrieve Configuration","error":"no kind is registered for the type v1alpha1.MemberOperatorConfig in scheme \"pkg/runtime/scheme.go:100\"","stacktrace":"github.com/codeready-toolchain/member-operator/controllers/memberoperatorconfig.GetConfiguration\n\t/home/igels/go/src/github.com/codeready-toolchain/member-operator/controllers/memberoperatorconfig/configuration.go:33\nmain.main\n\t/home/igels/go/src/github.com/codeready-toolchain/member-operator/cmd/consoleplugin/main.go:98\nruntime.main\n\t/snap/go/10029/src/runtime/proc.go:250"}
{"level":"error","ts":"2023-03-21T02:13:38.798Z","logger":"setup","msg":"Error retrieving Configuration","error":"no kind is registered for the type v1alpha1.MemberOperatorConfig in scheme \"pkg/runtime/scheme.go:100\"","stacktrace":"main.main\n\t/home/igels/go/src/github.com/codeready-toolchain/member-operator/cmd/consoleplugin/main.go:100\nruntime.main\n\t/snap/go/10029/src/runtime/proc.go:250"}
```
Let's address it.

JS code generation and e2e test can be addressed in separate PRs IMO

deploy/consoleplugin/member-operator-console-plugin.yaml
mfrancisc: Nice! 👍 

Maybe I've missed what `plugin-manifest.json` is, but indeed testing the same server is a good check!

alexeykazakov: `plugin-manifest.json` is just a file which defines a web-console plugin. This is what the weo-console will do first when loading the plugin. It will download `plugin-manifest.json`. So if it's available then the service is fully ready.

mfrancisc: Got it, cool! Thanks for the additional details!

alexeykazakov (approved):
The PR is good enough to be merged IMO so we can get it enabled in staging ASAP.
We can address e2e testing and JS generation in the following PRs.

pkg/consoleplugin/scriptserver/scriptserver.go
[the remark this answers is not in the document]
MatousJobanek: > Ideally we should apply the current configuration without restarting the service but it's fine for now to load configuration when the service is being started IMO. We can improve later.

@alexeykazakov in this case, you would need to watch the MemberConfig CR in the service and invalidate the cache when the key is changed - this seems to be a kind of overkill for such a small use case. Instead of it, I would rather set the key as a variable in the Deployment manifest of the service and let the member-operator redeploy the service when the key is changed.

>  This filename would change if we were to regenerate it after making changes to the source JS

@sbryzak if the filename is changed, then it needs to be one in our GH repository, which means:
1. making a PR
2. mering it 
3. releasing a new version of member operator including this service
4. redeploying the service with a new tag that contains the modified/renamed filename

> meaning we would have to then update the populating of the map cache code, thereby introducing a vector for human error occurring.

not at all - this will be fully automatic as described above :wink:

PR #428, opened 2023-04-07
Add email domain to pendo

alexeykazakov:


sonyccd (approved):
chatGPT said it would work

PR #429, opened 2023-04-07
Fix pendo plugin for non-sandbox users

alexeykazakov:
It fixes web-console with pendo plugin enabled for users which do not have the email annotation set in the User object. Like cluster admins.


sonyccd (approved):
Added conditionals so that if the email address is missing, it will be left out of the Pendo init function.

PR #434, opened 2023-05-05
refactor: use 'runtimeclient'

xcoulon:
preparing for upcoming changes, using 'runtimeclient' as a package alias everywhere it is used


xcoulon:
not needed for now

PR #435, opened 2023-05-05
add permissions (and ClusterRole) needed to apply appstudio tenant tier

MatousJobanek:
https://github.com/codeready-toolchain/host-operator/pull/797


config/rbac/appstudio_pipelines_runner_binding.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: ClusterRoleBinding
+metadata:
+  name: appstudio-pipelines-runner
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: appstudio-pipelines-runner
+subjects:
+- kind: ServiceAccount
+  name: controller-manager
```
alexeykazakov, line 11: Why do we need to create a role binding for the control manager? Why not for the member operator SA?

Can you add a detailed comment to this yaml explaining why we need this?

alexeykazakov (approved):
Looks good but I have a question regarding the controller-manager permissions.

config/rbac/appstudio_pipelines_runner_binding.yaml
MatousJobanek: Ok, I'm idiot. I didn't realize that it won't be actually deployed as it is, which means that the ClusterRole won't be available with that name :man_facepalming: :man_facepalming: :man_facepalming:

PR #436, opened 2023-05-08
create appstudio-pipelines-runner ClusterRole separately

MatousJobanek:


config/appstudio-pipelines-runner/controller-manager/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
+# Contains the ClusterRole and ClusterRoleBinding that grants necessary permissions
+# to the operator SA. It adds a prefix to both of them to differentiate the ClusterRole
+# from the one that should be created in the member cluster directly. This is needed
+# because OLM aggregates all ClusterRoles that are applied to the operator SA into one.
```
alexeykazakov, line 4: Does the aggregation happens on the OLM side when the operator is being installed (so it creates a new role on fly) or it's actually on the operator-sdk side during the operator build?

alexeykazakov (approved):

config/appstudio-pipelines-runner/controller-manager/kustomization.yaml
MatousJobanek: yeah, aggregation is done by operator-sdk. fixed 183ed3c

PR #440, opened 2023-05-15
fix: limit usage of * in RBAC

MatousJobanek:
because * includes also escalate permissions

there needs to be a follow-up on appstudio side as well

### DIFF
--- a/config/rbac/appstudio_api_permissions_role.yaml
+++ b/config/rbac/appstudio_api_permissions_role.yaml
@@ -1,0 +1,36 @@
+# permissions to scale Camel K resources.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: appstudio-api-permissions
+rules:
+- apiGroups:
+  - appstudio.redhat.com
+  - tekton.dev
+  - jvmbuildservice.io
+  resources:
+  - "*"
+  verbs:
+  # TODO fix after changing the ADR and associated appstudio teemplates
+  - "*"
+- apiGroups:
+  - managed-gitops.redhat.com
+  - results.tekton.dev
+  resources:
+  - "*"
+  verbs:
+  - get
+  - list
+  - watch
+  - update
+  - patch
+  - delete
+  - create
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  - configmaps
+  verbs:
+  # TODO fix after changing the ADR and associated appstudio teemplates
+  - "*"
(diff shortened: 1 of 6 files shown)


alexeykazakov (approved):

config/rbac/appstudio_api_permissions_role.yaml
```diff
@@ -0,0 +1,36 @@
+# permissions to scale Camel K resources.
```
mfrancisc, line 1: is this up to date with the actual purpose of this role?

controllers/nstemplateset/nstemplateset_controller.go
```diff
@@ -89,9 +89,8 @@ type Reconciler struct {
 //+kubebuilder:rbac:groups=toolchain.dev.openshift.com,resources=nstemplatesets/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=toolchain.dev.openshift.com,resources=nstemplatesets/finalizers,verbs=update
 
-//+kubebuilder:rbac:groups="",resources=namespaces;limitranges,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups="",resources=namespaces;resourcequotas,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=rbac.authorization.k8s.io;authorization.openshift.io,resources=rolebindings;roles;clusterroles;clusterrolebindings,verbs=*
+//+kubebuilder:rbac:groups="",resources=namespaces;limitranges;resourcequotas;pods/exec,verbs=get;list;watch;create;update;patch;delete
```
mfrancisc, line 92: just trying to understand - why is this controller doing with `pods/exec` resources ?

config/rbac/appstudio_api_permissions_role.yaml
MatousJobanek: Good catch. bad copy-paste - fixed in f8fbda0

controllers/nstemplateset/nstemplateset_controller.go
MatousJobanek: this is needed to be able to apply sandbox templates https://github.com/codeready-toolchain/host-operator/blob/master/deploy/templates/nstemplatetiers/base1ns/ns_dev.yaml#L21-L37

mfrancisc (approved):
Thanks for addressing my comments 👍

MatousJobanek:
unfortunately, this won't be as easy as it seems to be. It's very hard to reference to an already existing ClusterRole which is no managed by OLM and it's impossible to do that in e2e (or dev) environment where we deploy two versions of member-operator. The reason is that the OLM doesn't know which SA in which namespace the ClusterRole should be assigned to. Plus, there is also a conflict in the name of the created ClusterRoleBinding.

MatousJobanek:
I moved the PR to a draft until I figure out how to solve the problem I mentioned in my previous comment https://github.com/codeready-toolchain/member-operator/pull/440#issuecomment-1549520309

PR #451, opened 2023-06-09
use different uuid lib

ranakan19:
fix wrt: https://issues.redhat.com/browse/SANDBOX-78

### DIFF
--- a/config/appstudio-pipelines-runner/base/appstudio_pipelines_runner_role.yaml
+++ b/config/appstudio-pipelines-runner/base/appstudio_pipelines_runner_role.yaml
@@ -7,15 +7,6 @@ rules:
 - apiGroups:
   - tekton.dev
   resources:
-  - pipelineruns
-  verbs:
-  - get
-  - list
-  - update
-  - patch
-- apiGroups:
-  - tekton.dev
-  resources:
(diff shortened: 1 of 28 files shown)


alexeykazakov (approved):

mfrancisc (approved):
Looks good! 

Just one query: are we planning to use this library in all the repos? 
I'm asking because in host-operator I've used google/uuid : https://github.com/codeready-toolchain/host-operator/blob/c899290d8eeaa9944940b1320a89710fd1b775db/test/spacerequest/spacerequest.go#L7

MatousJobanek (approved):
+1 to @mfrancisc. Let's align this in other repos if possible

ranakan19:
> Just one query: are we planning to use this library in all the repos? I'm asking because in host-operator I've used google/uuid : https://github.com/codeready-toolchain/host-operator/blob/c899290d8eeaa9944940b1320a89710fd1b775db/test/spacerequest/spacerequest.go#L7

@mfrancisc host-operator does use gofrs/uuid, so will continue to use that in host as well. I made the change to use gofrs/uuid where you pointed, check it out here - https://github.com/codeready-toolchain/host-operator/pull/826

PR #453, opened 2023-06-15
feat: add support for patching service account resources in member operator

mfrancisc:
Add patch support for SA objects.

Jira: https://issues.redhat.com/browse/SANDBOX-87


controllers/nstemplateset/client_test.go
```diff
@@ -120,21 +120,18 @@ func TestApplyToolchainObjects(t *testing.T) {
...
-			return fmt.Errorf("should not update")
-		}
 
 		// when
 		changed, err := apiClient.ApplyToolchainObjects(logger, copyObjects(sa), additionalLabel)
 
 		// then
 		require.NoError(t, err)
-		assert.False(t, changed)
+		assert.True(t, changed)
```
alexeykazakov, line 134: Do you actually check that it was patched?

mfrancisc: Good point. I'm adding a more specific test, but while working at it , I've noticed that the existing SA might loose some values, for example and existing `ServiceAccount.Secrets` field will be set to `nil` with the current version of the code.

I was wondering if we want to support patch only for the labels and annotations or do we want to patch any addition that is being made to the template that has a ServiceAccount?

MatousJobanek: the risk of losing the list of secrets is exactly why we disabled the updates of SAs and why we wanted to use the "PATCH" instead - so it patches only the changes in labels & annotations and it doesn't override the list of secrets.

>I was wondering if we want to support patch only for the labels and annotations or do we want to patch any addition that is being made to the template that has a ServiceAccount?

I'm not aware of anything else we would need to update apart from annotations & secrets, +1 to supporting only that.

The alternative to "PATCH" would be reusing the already existing SA content:
1. Get already existing SA
2. Update only annotations & labels inside of the object
3. call update with the same object (using the same `resourceVersion`) - we could theoretically move this into the logic of `ApplyToolchainObjects`

MatousJobanek: I'm curious how this is done by other tools. I think that we discussed that when we were looking at the options of managing updates of resources in kcp - there is maybe some kind of update option that would preserve the list of secrets. @rajivnathan maybe remembers

rajivnathan: >there is maybe some kind of update option that would preserve the list of secrets. @rajivnathan maybe remembers
Ah, you're talking about Server-Side Apply(SSA) I think. See https://ymmt2005.hatenablog.com/entry/2020/04/14/An_example_of_using_dynamic_client_of_k8s.io/client-go and https://kubernetes.io/docs/reference/using-api/server-side-apply/

If some fields in the SAs are managed by one applier and other fields managed by another applier (not just managed by our controllers) then SSA is the better option because it specifically handles multiple appliers collaborating on a single resource. It will also return an error when one applier unintentionally tries to overwrite a value if it was set by another user. Then the applier has some options on how to proceed, see https://kubernetes.io/docs/reference/using-api/server-side-apply/#conflicts

I don't know much about the background for this story so I'm not really sure whether SSA is better or not for this case. It looks like @mfrancisc's change does the job. If there aren't concerns about "field management" then it's probably unnecessary to use SSA.

mfrancisc: @rajivnathan thanks for the useful information. 

Using SSA would provide us more features like:
- more fine grained control over the fields we update
- enables better management of the resources when multiple clients/controllers are supposed to update the same resource. 
- better patching strategy then Strategic Merge Patch (from the blog example in case of config maps)

But I don't think it's actually needed in our case, the SA we create/update for now (for appustudio and appstudio-env tiers as far as I can see) are not supposed to be updated by other clients/controllers right ? Plus setting up the dynamic client for SSA requires some extra code and setup in the member operator. I don't think we instantiate the dynamic client anywhere now.

But it's useful information, thanks!

alexeykazakov: Those SAs should not be updated by other RHTAP controllers. They can be updated (the secret reference part) by kubernetes though.
So I would keep it as simple as possible for now. If we can update the annotations and labels only (supposed to be managed and updated by the member operator only) and leave everything else alone then IMO it's good enough. We can revisit it later and look at the SSA option later if needed.

alexeykazakov: And let's make sure this usecase when we are updating a SA with some existing secret references is fully covered by our unit tests.

xcoulon (approved):

MatousJobanek (approved):
Looks great :+1: 
Thanks a lot for this improvement :1st_place_medal:

controllers/nstemplateset/client_test.go
MatousJobanek: right - SSA was the thing I wasn't able to remember. Thanks @rajivnathan for refreshing my goldfish memory :smile: 
Anyway, I fully agree that using SSA is a kind of overkill for this kind of usecase. As Alexey has mentioned, no one else should update the SAs (after all, there is not much else that could be updated anyway :-)). 
For now, what @mfrancisc implemented is fully sufficient and should cover all needed cases we can face with updating SAs. :+1: :rocket:

PR #464, opened 2023-07-21
Use toolchainv1alpha1.NamespaceTypeDefault

rajivnathan:
See https://github.com/codeready-toolchain/api/pull/369

### DIFF
--- a/controllers/nstemplateset/status.go
+++ b/controllers/nstemplateset/status.go
@@ -71,7 +71,7 @@ func (r *statusManager) updateStatusProvisionedNamespaces(nsTmplSet *toolchainv1
 	sort.Slice(provisionedNamespaces, func(i, j int) bool {
 		return provisionedNamespaces[i].Name < provisionedNamespaces[j].Name
 	})
-	provisionedNamespaces[0].Type = "default"
+	provisionedNamespaces[0].Type = toolchainv1alpha1.NamespaceTypeDefault
 
 	nsTmplSet.Status.ProvisionedNamespaces = provisionedNamespaces
 	return r.Client.Status().Update(context.TODO(), nsTmplSet)
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/codeready-toolchain/member-operator
 
 require (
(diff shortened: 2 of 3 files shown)


mfrancisc (commented):
Looks good! 👍 

you may want to replace the constant also in this test: https://github.com/codeready-toolchain/member-operator/blob/34ee2faf5c96b7c6dbdbc3eb64909a1432df968f/controllers/nstemplateset/status_test.go#L72

alexeykazakov:
> you may want to replace the constant also in this test:

I actually prefer not to use runtime constants in the tests when it's used in API.
In this specific case the "default" string as the namespace type is our API. It's used by HAC for example. And now imagine we change the global constant to something like "defalt" by mistake. Out test wouldn't catch that bug which would break HAC if we use the same constant in the test.
I prefer using constants in the runtime code only and use strings (or tests constants) in the tests to catch such bugs if present.

controllers/nstemplateset/status_test.go
```diff
@@ -69,7 +69,7 @@ func TestUpdateStatus(t *testing.T) {
 			HasProvisionedNamespaces([]toolchainv1alpha1.SpaceNamespace{
 				{
 					Name: spacename + "-dev",
-					Type: "default", // check that default type is added to first NS in alphabetical order
+					Type: toolchainv1alpha1.NamespaceTypeDefault, // check that default type is added to first NS in alphabetical order
```
alexeykazakov, line 72: See my comment above.
```suggestion
					Type: "default", // check that default type is added to first NS in alphabetical order
```

alexeykazakov (approved):
Looks good. But please see my comment above.

rajivnathan:
@alexeykazakov yeah, I was thinking the same but I figured e2e tests would cover that problem but it's a good point nonetheless so I reverted it in the unit tests too. We can just stick to avoiding constants in tests whenever possible.

mfrancisc:
> It's used by HAC for example. And now imagine we change the global constant to something like "defalt" by mistake. Out test wouldn't catch that bug which would break HAC if we use the same constant in the test.

Makes sense 👍

mfrancisc (approved):

xcoulon (approved):

PR #465, opened 2023-07-25
add security context

ranakan19:
for https://app.snyk.io/org/red-hat-developer-sandbox/project/863d225b-9e36-48fe-aad7-ffa5cc7d17c7


config/default/manager_config_patch.yaml
```diff
@@ -1,20 +0,0 @@
-apiVersion: apps/v1
-kind: Deployment
-metadata:
-  name: controller-manager
-  namespace: system
```
MatousJobanek, line 5: could you please remove the lines (even though they are just comments) from the kustomize.yaml file? https://github.com/codeready-toolchain/member-operator/blob/ec800b33605d265754c7791b6345cc229f62a27e/config/default/kustomization.yaml#L28-L30

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -20,6 +26,12 @@ spec:
         - containerPort: 8443
           name: https
       - name: manager
+        securityContext :
+          allowPrivilegeEscalation: false
+          runAsNonRoot: true
+          capabilities:
+            drop:
+              - all
```
MatousJobanek, line 34: the manager has already securityContext defined here:
https://github.com/codeready-toolchain/member-operator/blob/ec800b33605d265754c7791b6345cc229f62a27e/config/manager/manager.yaml#L36-L37
could we move the config there?

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -11,6 +11,12 @@ spec:
       containers:
       - name: kube-rbac-proxy
         image: registry.redhat.io/openshift4/ose-kube-rbac-proxy:v4.7.0
+        securityContext:
+          allowPrivilegeEscalation: false
+          runAsNonRoot: true
```
MatousJobanek, line 16: btw, the `runAsNonRoot` is already defined at the pod level
https://github.com/codeready-toolchain/member-operator/blob/ec800b33605d265754c7791b6345cc229f62a27e/config/manager/manager.yaml#L27-L28
I guess that we don't have to re-define it also for containers separately

ranakan19: from the documentation, "Security settings that you specify for a Container apply only to the individual Container, and they override settings made at the Pod level when there is overlap." I interpreted that we need to re-define it at container level again to avoid vulnerability.

config/default/manager_auth_proxy_patch.yaml
ranakan19: same as the other comment, the security context of container overrides the pod level security context so i think we should keep it at container level to avoid vulnerability

config/default/manager_config_patch.yaml
ranakan19: thanks, done in [6f7182b](https://github.com/codeready-toolchain/member-operator/pull/465/commits/6f7182bd096ca266757a9725534e9969dcf9ba2a)

config/default/manager_auth_proxy_patch.yaml
alexeykazakov: > The security settings that you specify for a Pod apply to all Containers in the Pod

However if an individual container template also specifies the security context then that one from the container template will override the one specified for the pod.

But this specific `config/default/manager_auth_proxy_patch.yaml` template doesn't have either the pod or the individual container security contexts specified. I guess this is why snyk complains about it.

config/default/manager_auth_proxy_patch.yaml
```diff
@@ -11,6 +11,11 @@ spec:
       containers:
       - name: kube-rbac-proxy
         image: registry.redhat.io/openshift4/ose-kube-rbac-proxy:v4.7.0
+        securityContext:
```
mfrancisc, line 14: minor thing - I guess snyk ignores the fact that this file is a patch.

I think updating this section should suffice: https://github.com/codeready-toolchain/member-operator/blob/ec800b33605d265754c7791b6345cc229f62a27e/config/manager/manager.yaml#L27-L28
with:
```
      securityContext:
        runAsNonRoot: true
        allowPrivilegeEscalation: false
        capabilities:
            drop:
               - all
```
so that we define it only once and future containers/patches will inherit it.

config/default/manager_auth_proxy_patch.yaml
MatousJobanek: tbh, we could easily drop the patch and move the definition of  the auth proxy directly to the manifest of the Deployment - there is no benefit in having it separate. We just reused whatever was generated by operator-sdk. This could also make the management easier

config/default/manager_auth_proxy_patch.yaml
ranakan19: that would be nice indeed, but `allowPrivilegeEscalation` and `capabilities` can only be set at container level.

config/default/manager_auth_proxy_patch.yaml
ranakan19: done in [5e08ac7](https://github.com/codeready-toolchain/member-operator/pull/465/commits/5e08ac7f00d983ba2633997088a5f06111ae7c5c)

config/default/manager_auth_proxy_patch.yaml
mfrancisc: haa my bad 🤦‍♂️ , indeed seems like those 2 fields are available in the [SecurityContext](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#securitycontext-v1-core) struct but not in  [PodSecurityContext](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#podsecuritycontext-v1-core). While `runAsNonRoot` and others are present in both structs. Never mind , you do need both sections.

config/manager/manager.yaml
```diff
@@ -66,5 +69,30 @@ spec:
...
+      - name: manager
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - all
+        args:
+          - "--health-probe-bind-address=:8081"
+          - "--metrics-bind-address=127.0.0.1:8080"
+          - "--leader-elect"
```
MatousJobanek, line 96: there is already one `manager` container defined 
https://github.com/codeready-toolchain/member-operator/blob/5e08ac7f00d983ba2633997088a5f06111ae7c5c/config/manager/manager.yaml#L30-L35
no need to define it as a new one - could you just incorporate the params into the original `manager` container?

ranakan19: my bad, fixed in 50d0ca9

config/default/kustomization.yaml
```diff
@@ -9,22 +9,12 @@ namespace: toolchain-member-operator
 namePrefix: member-operator-
 
 # Labels to add to all resources and selectors.
-#commonLabels:
-#  someName: someValue
+#commonLl someValue
 
-bases:
```
ranakan19, line 15: renamed bases to resources because bases is now depricated

rajivnathan: I was wondering about that :) Is this the relevant issue you saw: https://github.com/kubernetes/kubectl/issues/891#issuecomment-731238572 Or did you see that somewhere else?

config/default/kustomization.yaml
```diff
@@ -9,22 +9,12 @@ namespace: toolchain-member-operator
 namePrefix: member-operator-
 
 # Labels to add to all resources and selectors.
-#commonLabels:
-#  someName: someValue
+#commonLl someValue
```
rajivnathan, line 12: Typo?

ranakan19: done in c753592

mfrancisc (approved):
Looks good 👍

MatousJobanek (approved):
Looks great :+1: 
Thanks a lot for doing the additional changes :pray:

PR #477, opened 2023-10-04
Vm webhook limits

rajivnathan:
JIRA story: https://issues.redhat.com/browse/SANDBOX-362

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/809


pkg/webhook/mutatingwebhook/types/virtualmachine.go
```diff
@@ -0,0 +1,127 @@
+package types
```
alexeykazakov, line 1: Wouldn't it be safer to depend on actual CNV API instead of duplicating it here?

pkg/webhook/mutatingwebhook/vm_mutate.go
```diff
@@ -0,0 +1,87 @@
...
+	if limits == nil {
+		limits = corev1.ResourceList{}
+	}
+
+	anyChanges := false
+	for _, r := range []corev1.ResourceName{"memory", "cpu"} {
+		_, isLimitDefined := limits[r]
+		_, isRequestDefined := requests[r]
+		if !isLimitDefined && isRequestDefined {
+			limits[r] = requests[r]
```
alexeykazakov, line 72: So the only thing we actually need to do is to set `vm.Spec.Template.Spec.Domain.Resources.Limits` to the value from `vm.Spec.Template.Spec.Domain.Resources.Requests`? Let's add it to the description of the function. We cal also reference to the CNV JIRA issue pointing that it's a temp workaround for that.

pkg/webhook/mutatingwebhook/types/virtualmachine.go
rajivnathan: Yes, we could do that if we're okay with adding the dependency. But when I tried there were a bunch of dependency conflicts so we'd need to resolve those.

alexeykazakov: Having a dependency to CNV API only would be fine. But this API depends on something which creates conflicts on our end then it's not good. OK let's keep it this way. I'm concern with some API changes which would break our webhook though. We won't be able to just update the dependency to verify if it works or not with a newer CNV :(
Can we at least comment it here why we are duplicating their API here?
Also, any idea how we could mitigate risks of breaking API changes?

pkg/webhook/mutatingwebhook/vm_mutate.go
```diff
@@ -0,0 +1,87 @@
...
+	// instead of changing the object we need to tell K8s how to change the object
+	vmPatchItems := []map[string]interface{}{}
+
+	// ensure limits are set
+	vmPatchItems = ensureLimits(vm, vmPatchItems)
+
+	patchContent, err := json.Marshal(vmPatchItems)
+	if err != nil {
+		vmLogger.Error(err, "unable to marshal patch items for VirtualMachine", "AdmissionReview", admReview, "Patch-Items", vmPatchItems)
+		return responseWithError(errors.Wrapf(err, "unable to marshal patch items for VirtualMachine - raw request object: %v", admReview.Request.Object.Raw))
```
mfrancisc, line 49: minor thing - seems like it's missing a test case, not sure if you want to add one so that it's reaches 100% coverage :)

deploy/webhook/member-operator-webhook.yaml
```diff
@@ -137,6 +145,30 @@ objects:
...
+    clientConfig:
+      caBundle: ${CA_BUNDLE}
+      service:
+        name: member-operator-webhook
+        namespace: ${NAMESPACE}
+        path: "/mutate-virtual-machines"
+        port: 443
+    matchPolicy: Equivalent
+    rules:
+    - operations: ["CREATE"]
```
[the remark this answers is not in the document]
rajivnathan: The webhook's primary purpose is to ensure a VM will start when they try it in the sandbox for the first time so that's why I only have it for CREATE since the limits need to be defined the first time to ensure the VM can start.

The situation with UPDATE is a little less clear. In order for the webhook to do anything for the UPDATE case it would mean that the user has removed the limits from the VM after creation and maybe there's some reason why they'd want to do that. So that's why I didn't include it for UPDATE.

pkg/webhook/mutatingwebhook/vm_mutate.go
rajivnathan: It's not straightforward to test since `vmPatchItems` is returned by the `ensureLimits` func and I don't see how it would return something that would cause the marshalling to fail. I suppose that means we probably don't need to check for error here but I'd rather have it there just as a safeguard in case we're missing something.

If we really want to test it I'd have to have a fake ensureLimits func or do some refactoring and that doesn't seem worth it to do just for this purpose. So I'm okay with less than 100% coverage and the user pods mutate tests have the same gap in testing :)

pkg/webhook/mutatingwebhook/vm_mutate.go
rajivnathan: Okay, as discussed offline, I've added comments to better describe what we're doing and included a link to the CNV JIRA.

pkg/webhook/mutatingwebhook/vm_mutate.go
alexeykazakov: Yeah. Such things is pretty tricky to test. IMO that's OK to skip it.
But I like your attitude to have 100% coverage when possible @mfrancisc! :)

alexeykazakov (approved):
Looks good!
Looking forward to advanced e2e-tests though and also your ideas around how we can mitigate our risks of undetected braking CNV API changes... I don't have much to suggest :(

pkg/webhook/mutatingwebhook/types/virtualmachine.go
mfrancisc: +1 for importing this package if possible, would make maintenance easier (let's imagine if they do some breaking changes and split this package into several packages in the future).

deploy/webhook/member-operator-webhook.yaml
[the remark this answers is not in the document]
mfrancisc: Thanks for explaining that, so this means that users will be able to UPDATE `virtualmachines` resources after creation and the webhook will not interfere with the updates ?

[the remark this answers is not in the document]
MatousJobanek: but don't we want to inject the SSH key in the future using the webhook as well? That would mean that user can remove it, which is something we don't want to let them do, right?

pkg/webhook/mutatingwebhook/types/virtualmachine.go
MatousJobanek: Have you tried a generic approach - without introducing the dependency on their library? 
We could either play with the `map[string]interface{}` representation, or use a library to get the value (and check if limits are already set), for example https://github.com/tidwall/gjson should do the job. I know that it's introducing another dependency as well, but it's a more generic one and shouldn't cause any conflict problems with other dependencies.

Another approach (without introducing the dependency on them) would be creating a very simplified representation of what we want to check/get there:
```go

type VM struct {
    Spec struct {
        Template struct {
            Spec struct {
                Domain struct {
                    Resources struct {
                        Limits v1.ResourceList `json:"limits,omitempty"`
                        Requests v1.ResourceList `json:"requests,omitempty"`
                    } `json:"resources,omitempty"`
                } `json:"domain,omitempty"`
            } `json:"spec,omitempty"`
        } `json:"template,omitempty"`
    } `json:"spec,omitempty"`
}
```

alexeykazakov (approved):

deploy/webhook/member-operator-webhook.yaml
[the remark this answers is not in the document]
rajivnathan: just for the record, we discussed this offline and decided to keep it as only CREATE for now and we can add UPDATE later if necessary

pkg/webhook/mutatingwebhook/types/virtualmachine.go
rajivnathan: Used unstructured.Unstructured type for now but will switch to gjson if it's better in later PRs

pkg/webhook/mutatingwebhook/vm_mutate.go
```diff
@@ -0,0 +1,101 @@
...
+	return resp
+}
+
+// ensureLimits ensures resource limits are set on the VirtualMachine if requests are set, this is a workaround for https://issues.redhat.com/browse/CNV-28746 (https://issues.redhat.com/browse/CNV-32069)
+// The issue is that if the namespace has LimitRanges defined and the VirtualMachine resource does not have resource limits defined then it will use the LimitRanges which may be less than requested
+// resources and the VirtualMachine will fail to start.
+// This should be removed once https://issues.redhat.com/browse/CNV-32069 is complete.
+func ensureLimits(unstructuredObj *unstructured.Unstructured, patchItems []map[string]interface{}) []map[string]interface{} {
+
+	requests, reqFound, err := unstructured.NestedStringMap(unstructuredObj.Object, "spec", "template", "spec", "domain", "resources", "requests")
```
MatousJobanek, line 60: I didn't know about this `unstructured.NestedStringMap` function, it is really neat :+1: 
It does exactly what we need :tada:

PR #478, opened 2023-10-05
Adding Pod name for idle email notification

fbm3307:


controllers/idler/idler_controller.go
```diff
@@ -145,7 +145,7 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
 					podLogger.Info("Pod deleted")
 				}
 				// By now either a pod has been deleted or scaled to zero by controller, idler Triggered notification should be sent
-				if err := r.createNotification(logger, idler); err != nil {
+				if err := r.createNotification(logger, idler, pod.Name); err != nil {
```
xcoulon, line 148: instead of passing the actual pod name to the `createNotification` func, I believe it would be better if we can pass the controller name (ie, the name parent Deployment, StatefulSet, etc.), as the notification sent to the user will be more explicit. See the call to `scaleControllerToZero` on line 136.

For example: 

> In accordance with the usage terms of Developer Sandbox, we have reduced the number of instances of your
        application my-app to zero (0)

and not something like

> In accordance with the usage terms of Developer Sandbox, we have reduced the number of instances of your
        application my-app-1a2b3c to zero (0)

MatousJobanek: yeah, but that's applicable only in those cases when the pod is managed by a controller (eg. Deployment), but there could be cases when there is a single pod (which is not managed by any controller) that just got deleted - in such situation we can provide the pod name only, nothing else. 
In other words:
* if the pod is managed by a controller, let's provide the name of the controller
* if the pod is not managed by a controller, then let's provide the name of the pod

xcoulon: yes, I forgot to mention that if the pod has no controller, then we should pass it as-is.

fbm3307: Okay , i understand,  will it be the same case for daemonset, job controller type also  ?

alexeykazakov: There are a few options.
1. We just call it an application as we do now. We just add a name of the controller or standalone pod:
```
In accordance with the usage terms of Developer Sandbox, we have reduced the number of instances of your
application {{.ApplicationName}} to zero (0)
```
2. We introduce two params. Application Type and Application Name. Something like:
```
In accordance with the usage terms of Developer Sandbox, we have reduced the number of instances of your
{{.ApplicationType}} {{.ApplicationName}} to zero (0)
```
Where `{{.ApplicationType}}` is `Deployment`, `ReplicaSet` or `Pod` (if no controllers). And `{{.ApplicationName}}` is the deployment, replicaset, pod, etc name.

I think the second option would be even better.

BTW, please update the template itself so the controller/pod name is used here: https://github.com/codeready-toolchain/host-operator/blob/master/deploy/templates/notificationtemplates/sandbox/idlertriggered/notification.html#L50

alexeykazakov: Just noticed the other PR: https://github.com/codeready-toolchain/host-operator/pull/878
Please add it to the description of this PR so it's harder to miss :)

fbm3307: Added the related PR here also 
 
Okay agree for second option, making the changes to the PR

fbm3307: Updated the PR

controllers/idler/idler_controller.go
```diff
@@ -197,10 +212,18 @@ func (r *Reconciler) createNotification(logger logr.Logger, idler *toolchainv1al
...
+		if len(appName) == 0 {
+			appName = ""
+		}
+		if len(appType) == 0 {
+			appType = ""
+		}
 		keysAndVals := map[string]string{
 			"Namespace": idler.Name,
+			"Appname":   appName,
+			"Apptype":   appType,
```
MatousJobanek, line 224: out of curiosity - is there any reason for not using the CamelCase 
```suggestion
			"AppName":   appName,
			"AppType":   appType,
```

fbm3307: No Specific reason.. I shall update it to camel case

controllers/idler/idler_controller.go
```diff
@@ -144,8 +144,23 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
...
+						owner.Kind == "StatefulSet" || owner.Kind == "DeploymentConfig" ||
+						owner.Kind == "ReplicationController") {
+						appName = owner.Name
+						appType = owner.Kind
+						break
+					} else {
+						appName = pod.Name
+						appType = ""
+					}
+				}
```
mfrancisc, line 161: Nothing problematic with your code , but looking at `scaleControllerToZero` seems like we already have some logic to recursively find the root controller . 

For example , nn the case we have this hierarchy `Deployment --> ReplicaSet --> Pod` , this logic will report the `ReplicaSet` which might be fine, not a big issue. But I was wondering if in this case we want the `Deployment` name instead. 

ReplicaSet might be: `host-operator-controller-manager-5d7756db97` while Deployment name is: `host-operator-controller-manager`.

This is already a great improvement, so feel free to leave it as is if it's too complicated to reuse some of that logic for the recursive search . T

controllers/idler/idler_controller.go
```diff
@@ -197,10 +212,18 @@ func (r *Reconciler) createNotification(logger logr.Logger, idler *toolchainv1al
 			// no email found, thus no email sent
 			return fmt.Errorf("no email found for the user in MURs")
 		}
-
+		if len(appName) == 0 {
+			appName = ""
+		}
+		if len(appType) == 0 {
+			appType = ""
+		}
```
mfrancisc, line 220: wondering if we really need this - if the len of the string is 0 isn't the string already blank or am I missing something ?

mfrancisc (commented):
Nice job! 

I left few minor comments only.

controllers/idler/idler_controller.go
```diff
@@ -144,8 +144,23 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
...
+					if owner.Controller != nil && (owner.Kind == "Deployment" || owner.Kind == "ReplicaSet" ||
+						owner.Kind == "StatefulSet" || owner.Kind == "DeploymentConfig" ||
+						owner.Kind == "ReplicationController") {
+						appName = owner.Name
+						appType = owner.Kind
+						break
+					} else {
+						appName = pod.Name
+						appType = ""
+					}
```
mfrancisc, line 160: Nothing problematic with your code , but looking at `scaleControllerToZero` seems like we already have some logic to recursively find the root controller . 

For example , nn the case we have this hierarchy `Deployment --> ReplicaSet --> Pod` , this logic will report the `ReplicaSet` which might be fine, not a big issue. But I was wondering if in this case we want the `Deployment` name instead. 

ReplicaSet might be: `host-operator-controller-manager-5d7756db97` while Deployment name is: `host-operator-controller-manager`.

This is already a great improvement, so feel free to leave it as is if it's too complicated to reuse some of that logic for the recursive search .

fbm3307: my understanding was that if its a replica we are going to scale to 0, we want the customer to know that we reduced the replica and not the main deployment..?

controllers/idler/idler_controller.go
fbm3307: i can remove this ,its not required, it was written previously so that it doesn't break something else but. i handled that scenario differently

controllers/idler/idler_controller.go
alexeykazakov: +1 to CamelCase.

controllers/idler/idler_controller.go
alexeykazakov: In case of Deployments, as @mfrancisc mentioned there is a ReplicaSet too. Which is just an implementation detail of the Deployment. This is why we always look for the top parent and scale that parent to zero. If we will tell the user that we scaled down to zero the ReplicaSet then it would be misleading. It's not what we scaled down. And what is more importantly if the user tries to scale up the ReplicaSet instead of the Deployment then that will be auto-reverted by the Deployment. The Deployment is the ultimate controller here.

I would extend our `scaleControllerToZero()` function so it also returns the type of the controller if any. We should also cover the standalone pod case (no controllers). Basically let's keep the type of the actual resource we scale down or delete and re-use it when creating a notification instead of re-calculating it again after we scaled it down/deleted.

controllers/idler/idler_controller.go
alexeykazakov: Let's delete this :) This code doesn't really do anything.

controllers/idler/idler_controller_test.go
```diff
@@ -594,6 +594,7 @@ func TestCreateNotification(t *testing.T) {
 		},
 		Spec: toolchainv1alpha1.IdlerSpec{TimeoutSeconds: 60},
 	}
+
```
alexeykazakov, line 597: The current tests are not enough to cover the new functionality.
I would add a test add use the `prepareReconcile()` function which already creates a tree of different workloads with different controllers. Then call the `scaleControllerToZero()` function for each type/case and check the outcome AppName/Type.

controllers/idler/idler_controller.go
```diff
@@ -262,47 +262,41 @@ func (r *Reconciler) getUserEmailsFromMURs(logger logr.Logger, hostCluster *clus
 // scaleControllerToZero checks if the object has an owner controller (Deployment, ReplicaSet, etc)
 // and scales the owner down to zero and returns "true".
 // Otherwise returns "false".
+// It Also returns Parent Controller Name and type.
```
alexeykazakov, line 265:
```suggestion
// It also returns the parent controller type and name or empty strings if there is no parent controller.
```

controllers/idler/idler_controller.go
```diff
@@ -144,8 +146,13 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
 					}
 					podLogger.Info("Pod deleted")
 				}
+
+				if appName == "" {
+					appName = pod.Name
```
alexeykazakov, line 151:
```suggestion
					appName = pod.Name
					appType = "Pod"
```

controllers/idler/idler_controller.go
```diff
@@ -132,8 +132,10 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
 			// Already tracking this pod. Check the timeout.
 			if time.Now().After(trackedPod.StartTime.Add(time.Duration(idler.Spec.TimeoutSeconds) * time.Second)) {
 				podLogger.Info("Pod running for too long. Killing the pod.", "start_time", trackedPod.StartTime.Format("2006-01-02T15:04:05Z"), "timeout_seconds", idler.Spec.TimeoutSeconds)
+				var appName string
```
ranakan19, line 135: this can be removed since it's defined in the following lines with :=

controllers/idler/idler_controller.go
```diff
@@ -132,8 +132,10 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
...
+				var appType string
```
ranakan19, line 136: this can be removed since it's defined in the following lines with :=

controllers/idler/idler_controller_test.go
```diff
@@ -583,7 +583,195 @@ func TestEnsureIdlingFailed(t *testing.T) {
...
+			},
+		},
+		Spec: toolchainv1alpha1.IdlerSpec{TimeoutSeconds: 60},
+	}
+	namespaces := []string{"dev", "stage"}
+	usernames := []string{"alex"}
+	nsTmplSet := newNSTmplSet(test.MemberOperatorNs, "alex", "advanced", "abcde11", namespaces, usernames)
+	mur := newMUR("alex")
+	reconciler, _, _, _ := prepareReconcile(t, idler.Name, getHostCluster, idler, nsTmplSet, mur)
+	payloads := preparePayloads(t, reconciler, idler.Name, "", time.Now())
```
ranakan19, line 602: could we use a different name of the variable instead of payloads since it shadows the struct in here https://github.com/codeready-toolchain/member-operator/blob/master/controllers/idler/idler_controller_test.go#L796

ranakan19:
Maybe we should a check here to confirm that the appName and appType are rightly passed to the notification as well - https://github.com/codeready-toolchain/member-operator/blob/master/controllers/idler/idler_controller_test.go#L326

ranakan19 (commented):
Looks good overall! left some comments and need to fix the cyclomatic complexity in test

controllers/idler/idler_controller_test.go
```diff
@@ -583,7 +583,195 @@ func TestEnsureIdlingFailed(t *testing.T) {
...
+
+	t.Run("Test AppName/AppType for 'Deployment' ", func(t *testing.T) {
+		//given
+		p := func() *corev1.Pod {
+			for _, pod := range payloads.controlledPods {
+				for _, owner := range pod.OwnerReferences {
+					if owner.Kind == "ReplicaSet" && owner.Name == fmt.Sprintf("%s-replicaset", payloads.deployment.Name) {
```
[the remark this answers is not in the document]
alexeykazakov: "ReplicaSet" kind is correct here because this test verifies the nested controller use case: Deployment -> ReplicaSet -> Pod.
So the scaling down the pods with a "ReplicaSet" owner should actually scale down the Deployment parent.
I would just add a comment to that test to explain it.

controllers/idler/idler_controller_test.go
```diff
@@ -583,7 +583,195 @@ func TestEnsureIdlingFailed(t *testing.T) {
...
+				}
+			}
+			return nil
+		}()
+		//when
+		appType, appName, deletedByController, err := reconciler.scaleControllerToZero(logf.FromContext(context.TODO()), p.ObjectMeta)
+		if appName == "" {
+			appName = p.Name
+			appType = "Pod"
+		}
```
alexeykazakov: We can't unfortunately test the individual pod case using `scaleControllerToZero()` function because it's not covered by that function. I'm afraid we need to test this case separately. You could prepare the reconcile by creating an individual Pod only (no other pods or controllers) and then reconcile it. Then check that the notification was created and that it's set to Pod kind and its name. It will also cover the functionality which actually sets the appName and appType in the notification object.

alexeykazakov (commented):
Looks good! I've reduced the code duplication in the tests though: https://github.com/codeready-toolchain/member-operator/pull/478/commits/8ddb790e52985e37d11937698b7627e3f515688c

The only thing is left is the "standalone Pod" use case. We need to cover it by another tests. Please see my comment. The rest looks good!

alexeykazakov:
> Maybe we should a check here to confirm that the appName and appType are rightly passed to the notification as well

Right. I suggest another test which would cover the standalone pod and the notification's appName & appType all together. See my comment above.

controllers/idler/idler_controller_test.go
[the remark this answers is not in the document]
ranakan19: I see, thanks for the explanation

fbm3307:
> Maybe we should a check here to confirm that the appName and appType are rightly passed to the notification as well - https://github.com/codeready-toolchain/member-operator/blob/master/controllers/idler/idler_controller_test.go#L326

yes updated it with new test case

alexeykazakov (approved):
Looks good! Thank you for addressing all the comments.

PR #485, opened 2023-10-19
No Notification for "Completed" and without controller Pods

fbm3307:
If the Pod is in complete state and doesn't have a controller, no notification should be sent to the user.

This relates to #478

### DIFF
--- a/controllers/idler/idler_controller.go
+++ b/controllers/idler/idler_controller.go
@@ -134,11 +134,9 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
 				podLogger.Info("Pod running for too long. Killing the pod.", "start_time", trackedPod.StartTime.Format("2006-01-02T15:04:05Z"), "timeout_seconds", idler.Spec.TimeoutSeconds)
 				var podreason string
 				podcondition := pod.Status.Conditions
-				if podcondition != nil {
-					for _, podtype := range podcondition {
-						if podtype.Type == "Ready" {
-							podreason = podtype.Reason
-						}
+				for _, podtype := range podcondition {
+					if podtype.Type == "Ready" {
+						podreason = podtype.Reason
 					}
 				}
 				// Check if it belongs to a controller (Deployment, DeploymentConfig, etc) and scale it down to zero.
--- a/controllers/idler/idler_controller.go
+++ b/controllers/idler/idler_controller.go
@@ -139,6 +139,7 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
 						podreason = podtype.Reason
 					}
 				}
+
 				// Check if it belongs to a controller (Deployment, DeploymentConfig, etc) and scale it down to zero.
 				appType, appName, deletedByController, err := r.scaleControllerToZero(podLogger, pod.ObjectMeta)
 				if err != nil {
@@ -156,8 +157,9 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
 					appName = pod.Name
 					appType = "Pod"
 				}
-				// Do not send notification if Pod Not managed by a controller and is in completed state
-				if podreason != "PodCompleted" && !deletedByController {
+				// Send notification if the deleted pod was managed by a controller or was a standalone pod that was not completed
+				// eg. If a build pod is in "PodCompleted" status then it was not running so there's no reason to send an idler notification
+				if podreason != "PodCompleted" || deletedByController {
 					// By now either a pod has been deleted or scaled to zero by controller, idler Triggered notification should be sent
 					if err := r.createNotification(logger, idler, appName, appType); err != nil {
 						logger.Error(err, "failed to create Notification")
--- a/controllers/idler/idler_controller.go
+++ b/controllers/idler/idler_controller.go
@@ -135,9 +135,11 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
 				var podreason string
 				podcondition := pod.Status.Conditions
 				if podcondition != nil {
-					podreason = podcondition[0].Reason
-				} else {
-					podreason = ""
+					for _, podtype := range podcondition {
+						if podtype.Type == "Ready" {
+							podreason = podtype.Reason
+						}
+					}
 				}
 				// Check if it belongs to a controller (Deployment, DeploymentConfig, etc) and scale it down to zero.
 				appType, appName, deletedByController, err := r.scaleControllerToZero(podLogger, pod.ObjectMeta)
--- a/controllers/idler/idler_controller.go
+++ b/controllers/idler/idler_controller.go
@@ -132,7 +132,13 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
 			// Already tracking this pod. Check the timeout.
 			if time.Now().After(trackedPod.StartTime.Add(time.Duration(idler.Spec.TimeoutSeconds) * time.Second)) {
 				podLogger.Info("Pod running for too long. Killing the pod.", "start_time", trackedPod.StartTime.Format("2006-01-02T15:04:05Z"), "timeout_seconds", idler.Spec.TimeoutSeconds)
-
+				var podreason string
+				podcondition := pod.Status.Conditions
+				if podcondition != nil {
+					podreason = podcondition[0].Reason
+				} else {
+					podreason = ""
+				}
 				// Check if it belongs to a controller (Deployment, DeploymentConfig, etc) and scale it down to zero.
 				appType, appName, deletedByController, err := r.scaleControllerToZero(podLogger, pod.ObjectMeta)
 				if err != nil {
@@ -150,13 +156,15 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
 					appName = pod.Name
 					appType = "Pod"
 				}
-
-				// By now either a pod has been deleted or scaled to zero by controller, idler Triggered notification should be sent
-				if err := r.createNotification(logger, idler, appName, appType); err != nil {
-					logger.Error(err, "failed to create Notification")
-					if err = r.setStatusIdlerNotificationCreationFailed(idler, err.Error()); err != nil {
-						logger.Error(err, "failed to set status IdlerNotificationCreationFailed")
-					} // not returning error to continue tracking remaining pods
+				// Do not send notification if Pod Not managed by a controller and is in completed state
+				if podreason != "PodCompleted" && !deletedByController {
+					// By now either a pod has been deleted or scaled to zero by controller, idler Triggered notification should be sent
+					if err := r.createNotification(logger, idler, appName, appType); err != nil {
+						logger.Error(err, "failed to create Notification")
+						if err = r.setStatusIdlerNotificationCreationFailed(idler, err.Error()); err != nil {
+							logger.Error(err, "failed to set status IdlerNotificationCreationFailed")
+						} // not returning error to continue tracking remaining pods
+					}
 				}
 
 			} else {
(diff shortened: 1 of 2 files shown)


controllers/idler/idler_controller.go
```diff
@@ -132,7 +132,13 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
 			// Already tracking this pod. Check the timeout.
 			if time.Now().After(trackedPod.StartTime.Add(time.Duration(idler.Spec.TimeoutSeconds) * time.Second)) {
 				podLogger.Info("Pod running for too long. Killing the pod.", "start_time", trackedPod.StartTime.Format("2006-01-02T15:04:05Z"), "timeout_seconds", idler.Spec.TimeoutSeconds)
-
+				var podreason string
+				podcondition := pod.Status.Conditions
+				if podcondition != nil {
+					podreason = podcondition[0].Reason
```
rajivnathan, line 138: Shouldn't we check the status of a specific condition? I guess the "Ready" condition?

controllers/idler/idler_controller.go
```diff
@@ -132,7 +132,13 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
...
+				} else {
+					podreason = ""
+				}
```
rajivnathan, line 141:
```suggestion
				}
```
The else case is redundant since the default value for strings is `""` so we can remove it

rajivnathan (commented):
Looks good, just a few questions. Also, we should add some unit tests for this.

controllers/idler/idler_controller.go
fbm3307: will do

controllers/idler/idler_controller.go
fbm3307: All the condition "type" will have a _reason_ as "Pod Completed" , so i didnt check that .. but can add to check the reason only for "ready" condition type

fbm3307:
> Also, we should add some unit tests for this.

Added the Unit test case for this.
Looks like i will have to update e2e too

controllers/idler/idler_controller.go
```diff
@@ -150,13 +156,15 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
...
 				}
-
-				// By now either a pod has been deleted or scaled to zero by controller, idler Triggered notification should be sent
-				if err := r.createNotification(logger, idler, appName, appType); err != nil {
-					logger.Error(err, "failed to create Notification")
-					if err = r.setStatusIdlerNotificationCreationFailed(idler, err.Error()); err != nil {
-						logger.Error(err, "failed to set status IdlerNotificationCreationFailed")
-					} // not returning error to continue tracking remaining pods
+				// Do not send notification if Pod Not managed by a controller and is in completed state
+				if podreason != "PodCompleted" && !deletedByController {
```
alexeykazakov, line 160: Shouldn't it be
```suggestion
				if podreason != "PodCompleted" || deletedByController {
```
instead?
Your current statement will result in creating a notification only in the case of standalone pods in a non-complete state. We still want to create notifications for pods scaled down to zero through their controllers.

And please also make sure that all those use cases are covered by the unit tests:
- Standalone pod with "PodCompleted" reason.
- Standalone pod with a non-complete reason.
- Pod managed by a controller with "PodCompleted" reason.
- Pod managed by a controller with a non-complete reason.

controllers/idler/idler_controller.go
```diff
@@ -150,13 +156,15 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
...
 					appType = "Pod"
 				}
-
-				// By now either a pod has been deleted or scaled to zero by controller, idler Triggered notification should be sent
-				if err := r.createNotification(logger, idler, appName, appType); err != nil {
-					logger.Error(err, "failed to create Notification")
-					if err = r.setStatusIdlerNotificationCreationFailed(idler, err.Error()); err != nil {
-						logger.Error(err, "failed to set status IdlerNotificationCreationFailed")
-					} // not returning error to continue tracking remaining pods
+				// Do not send notification if Pod Not managed by a controller and is in completed state
```
rajivnathan, line 159:
```suggestion
				// Send notification if the deleted pod was managed by a controller or was a standalone pod that was not completed eg. If a build pod is in "PodCompleted" status then it was not running so there's no reason to send an idler notification
```
I think adding a bit more context here can help when we look at this code in the future since this is not an obvious case

controllers/idler/idler_controller.go
```diff
@@ -132,7 +132,13 @@ func (r *Reconciler) ensureIdling(logger logr.Logger, idler *toolchainv1alpha1.I
...
 			if time.Now().After(trackedPod.StartTime.Add(time.Duration(idler.Spec.TimeoutSeconds) * time.Second)) {
 				podLogger.Info("Pod running for too long. Killing the pod.", "start_time", trackedPod.StartTime.Format("2006-01-02T15:04:05Z"), "timeout_seconds", idler.Spec.TimeoutSeconds)
-
+				var podreason string
+				podcondition := pod.Status.Conditions
+				for _, podtype := range podcondition {
+					if podtype.Type == "Ready" {
+						podreason = podtype.Reason
+					}
+				}
```
ranakan19, line 141:
```suggestion
		readyCondition, found := condition.FindConditionByType(pod.Status.Conditions, toolchainv1alpha1.ConditionReady)
               if found {
                    podReason := readyCondition.Reason
                }
```

could maybe do something like this using the common function - https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/condition/condition.go#L51

controllers/idler/idler_controller.go
fbm3307: Yes will change

controllers/idler/idler_controller.go
fbm3307: we want to send notification for individual pods which are without controllers and are non-complete state ..? If yes than it cannot be || condition, else for any pod without controller there will be no notification

alexeykazakov: > we want to send notification for individual pods which are without controllers and are non-complete state ..?

Yes. But we also want to send a notification for pods owned by a controller. No matter what state.

> If yes than it cannot be || condition, else for any pod without controller there will be no notification.

Not really. There will be no notification only if both conditions are met: pod is complete AND not owned by a controller (the "if" will be false in that case, so the notification is skipped).

fbm3307: will do

controllers/idler/idler_controller.go
fbm3307: Here i am not able to use this function as the variable pod.Status.condition,  the info on pod status we get it from corev1 and not toolchainv1alpha1

fbm3307:
I will update the test cases to reduce duplication

controllers/idler/idler_controller_test.go
```diff
@@ -682,51 +682,89 @@ func TestAppNameTypeForControllers(t *testing.T) {
...
-		assert.NoError(t, err)
-		assert.True(t, res.Requeue)
+	testpod := map[string]struct {
+		pcond           corev1.PodCondition
+		expectedAppType string
+	}{
+		"Individual-Completed-Pod": {
+			pcond:           corev1.PodCondition{Type: "Ready", Reason: "PodCompleted"},
+			expectedAppType: "Pod",
+		},
```
alexeykazakov, line 708: Instead of relying on the test name below:
```
if pt == "Individual-Completed-Pod" || pt == "Individual-NonCompleted-Pod" {
```
I would add a dedicated field to distinguish controlled owned pods from standalone ones. Something like this: 
```suggestion
		"Individual-Completed-Pod": {
			pcond:           corev1.PodCondition{Type: "Ready", Reason: "PodCompleted"},
			expectedAppType: "Pod",
			expectedNotificationCreated: true, // set to true if it's expected that a notification is created
			controllerOwned: false,
		},
```
And then use it:
```
if tcs.controllerOwned {
	preparePayloads(t, reconciler, idler.Name, "todelete-", idlerTimeoutPlusOneSecondAgo, tcs.pcond)
} else {
	preparePayloadsSinglePod(t, reconciler, idler.Name, "todelete-", idlerTimeoutPlusOneSecondAgo, tcs.pcond)
}
```
and
```
if tcs.expectedNotificationCreated {
	...
} else {
	...
}

File: controllers/idler/idler_controller_test.go [hunk 7]
```diff
@@ -682,51 +682,89 @@ func TestAppNameTypeForControllers(t *testing.T) {
...
+				Name:      "feny-stage-idled",
+			}, notification)
+
+			if pt == "Individual-Completed-Pod" {
+				require.EqualError(t, err, "notifications.toolchain.dev.openshift.com \"feny-stage-idled\" not found")
+			} else {
+				require.NoError(t, err)
+				require.Equal(t, "feny@test.com", notification.Spec.Recipient)
+				require.Equal(t, "idled", notification.Labels[toolchainv1alpha1.NotificationTypeLabelKey])
+				require.Equal(t, tcs.expectedAppType, notification.Spec.Context["AppType"])
```
alexeykazakov, line 763: What about checking the AppName?

controllers/idler/idler_controller_test.go
```diff
@@ -682,51 +682,89 @@ func TestAppNameTypeForControllers(t *testing.T) {
...
+			expectedAppType: "Pod",
+		},
+		"Controlled-Completed-Pod": {
+			pcond:           corev1.PodCondition{Type: "Ready", Reason: "PodCompleted"},
+			expectedAppType: "Deployment",
+		},
+		"Controlled-NonCompleted-Pod": {
+			pcond:           corev1.PodCondition{Type: "Ready", Reason: ""},
+			expectedAppType: "Deployment",
+		},
```
alexeykazakov, line 720: Let's cover two more use cases:
- There is no conditions at all.
- There are multiple conditions.

alexeykazakov (commented):
Looks good! Just a few minor comments for test improvement.

controllers/idler/idler_controller.go
ranakan19: you're right, its the podCondition here, not condition :)

controllers/idler/idler_controller_test.go
fbm3307: done

controllers/idler/idler_controller_test.go
fbm3307: i forgot to add it for individual pod when i refactored the test case, added the check now for individual pod

controllers/idler/idler_controller_test.go
fbm3307: @alexeykazakov the first "no conditions" i added, for 2nd "multiple condition" do you mean to pass a list of conditions or just another reason type.? 
My understanding is that pod can have a single condition status at a time. Not sure though

fbm3307: I have added the multiple conditions too in test

controllers/idler/idler_controller_test.go
```diff
@@ -682,51 +682,116 @@ func TestAppNameTypeForControllers(t *testing.T) {
...
+				Namespace: test.HostOperatorNs,
+				Name:      "feny-stage-idled",
+			}, notification)
+			if tcs.expectedNotificationCreated {
+				require.NoError(t, err)
+				require.Equal(t, "feny@test.com", notification.Spec.Recipient)
+				require.Equal(t, "idled", notification.Labels[toolchainv1alpha1.NotificationTypeLabelKey])
+				require.Equal(t, tcs.expectedAppType, notification.Spec.Context["AppType"])
+				if !tcs.controllerOwned {
+					require.Equal(t, pname, notification.Spec.Context["AppName"])
```
alexeykazakov, line 788: It would be nice to check the AppName for the controllers but I understand it's a bit tricky in this specific test. That's probably fine because AppNames is covered in the other tests for the controllers anyway.

alexeykazakov (approved):
Looks good. Thank you for addressing the comments.

controllers/idler/idler_controller_test.go
fbm3307: The Same thoughts and hence didn't add the tests for controller here 
Thanks for the review

PR #490, opened 2023-10-30
Update VM webhook to add cloud init user data

rajivnathan:
https://issues.redhat.com/browse/SANDBOX-363

- Adds logic to parse userData for both cloudInitNoCloud and cloudInitConfigDrive volumes and update the data
- Refactored and simplified the unit tests for mutating webhooks
- Temporarily hardcoding a fake key, will have a separate PR for retrieval from config

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/830

### DIFF
--- a/pkg/webhook/mutatingwebhook/mutate.go
+++ b/pkg/webhook/mutatingwebhook/mutate.go
@@ -7,10 +7,12 @@ import (
 	"net/http"
 
 	"github.com/go-logr/logr"
+	admissionv1 "k8s.io/api/admission/v1"
 	v1 "k8s.io/api/admission/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 var (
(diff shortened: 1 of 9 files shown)


pkg/webhook/mutatingwebhook/mutate.go
```diff
@@ -19,51 +20,60 @@ var (
...
-func mutate(logger logr.Logger, body []byte, mutator mutateHandler) []byte {
-	admReview := v1.AdmissionReview{}
-	if _, _, err := deserializer.Decode(body, nil, &admReview); err != nil {
-		logger.Error(err, "unable to deserialize the admission review object", "body", string(body))
-		admReview.Response = responseWithError(err)
+	// deserialize the request
+	admReview := admissionv1.AdmissionReview{}
+	if _, _, err := deserializer.Decode(admReviewBody, nil, &admReview); err != nil {
+		logger.Error(err, "unable to deserialize the admission review object", "body", string(admReviewBody))
+		writeResponse(logger, http.StatusBadRequest, w, []byte("unable to read the body of the request"))
```
alexeykazakov, line 45: I'm curious why you just write an http response here and return instead of using an admission response with not-allowed and an error?

rajivnathan: It's because the AdmissionResponse is added to `admReview.Response` then when we write the response the body is the full `admReview` object and until line 55 we haven't really validated `admReview`. I'm not sure if you can return the AdmissionResponse object by itself if that's what you're suggesting. I haven't tried that. The original code (before I added the vm mutator) returned the full `admReview` object too.

What I have:
```
// sets "Response"
admReview.Response = mutator(admReview)

// respBody is the full admReview
respBody, err := json.Marshal(admReview)
if err != nil {
	logger.Error(err, "unable to marshal the admission review with response", "admissionReview", admReview)
	writeResponse(logger, http.StatusInternalServerError, w, []byte("failed to marshal the adm review resposne"))
	return
}

// the full respBody is written as the response
writeResponse(logger, http.StatusOK, w, respBody)
```

Original code:
https://github.com/codeready-toolchain/member-operator/blob/fdca75cbca46b4d2edd41c4fef5eeb05a2704a86/pkg/webhook/mutatingwebhook/mutate.go#L82-L86

Looking at the original code the full `admReview` is returned but what's interesting is it set admReview.Response even if the decoding fails. I'm not sure that's the correct thing to do though.

rajivnathan: As discussed, I changed it to use `admReview.Response` in the error cases for readability and since that's how it was in the original code. See c1f6dff0eeeab396c4031454704a063a519dacd8

alexeykazakov (approved):
Looks good!

PR #503, opened 2023-11-10
upgrade golang to 1.20

ranakan19:


.github/workflows/ci-build.yml
```diff
@@ -46,7 +46,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v3
```
mfrancisc, line 47: not related to your PR, but I see we use a mix of `actions/setup-go@v4` and `actions/setup-go@v3`, maybe we could upgrade this too.

mfrancisc (approved):
Looks good 👍 

Thanks for updating Golang version!

alexeykazakov (approved):

rajivnathan (approved):

PR #507, opened 2023-11-23
refactored member operator to only use properties from propagated claims

sbryzak:
This PR is a total refactor of the UserAccountController to use the values stored in PropagatedClaims instead of any other property or annotation values that are derived from the user's SSO token.

Fixes https://issues.redhat.com/browse/SANDBOX-431

E2E tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/858

### DIFF
--- a/controllers/useraccount/useraccount_controller.go
+++ b/controllers/useraccount/useraccount_controller.go
@@ -244,8 +244,8 @@ func (r *Reconciler) ensureUser(ctx context.Context, config membercfg.Configurat
 		expectedIdentities = append(expectedIdentities, commonidentity.NewIdentityNamingStandard(userAcc.Spec.PropagatedClaims.OriginalSub, config.Auth().Idp()).IdentityName())
 	}
 
-	// Also if the sso-user-id annotation is set, then another additional identity is required if it is a different value to the Spec.UserID
-	if userAcc.Spec.PropagatedClaims.UserID != "" && userAcc.Spec.PropagatedClaims.UserID != userAcc.Spec.UserID {
+	// Also if the UserID property is set, then an additional identity is required if it is a different value to the Sub
+	if userAcc.Spec.PropagatedClaims.UserID != "" && userAcc.Spec.PropagatedClaims.UserID != userAcc.Spec.PropagatedClaims.Sub {
 		expectedIdentities = append(expectedIdentities, commonidentity.NewIdentityNamingStandard(
 			userAcc.Spec.PropagatedClaims.UserID, config.Auth().Idp()).IdentityName())
 	}
@@ -630,7 +630,7 @@ func newUser(userAcc *toolchainv1alpha1.UserAccount, config membercfg.Configurat
 	if userAcc.Spec.PropagatedClaims.OriginalSub != "" {
 		identities = append(identities, commonidentity.NewIdentityNamingStandard(userAcc.Spec.PropagatedClaims.OriginalSub, config.Auth().Idp()).IdentityName())
 	}
-	if userAcc.Spec.PropagatedClaims.UserID != "" && userAcc.Spec.PropagatedClaims.UserID != userAcc.Spec.PropagatedClaims.OriginalSub {
+	if userAcc.Spec.PropagatedClaims.UserID != "" && userAcc.Spec.PropagatedClaims.UserID != userAcc.Spec.PropagatedClaims.Sub {
 		identities = append(identities, commonidentity.NewIdentityNamingStandard(userAcc.Spec.PropagatedClaims.UserID, config.Auth().Idp()).IdentityName())
 	}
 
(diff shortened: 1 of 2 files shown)


controllers/useraccount/useraccount_controller.go
```diff
@@ -237,19 +237,17 @@ func (r *Reconciler) ensureUser(ctx context.Context, config membercfg.Configurat
...
+		expectedIdentities = append(expectedIdentities, commonidentity.NewIdentityNamingStandard(userAcc.Spec.PropagatedClaims.OriginalSub, config.Auth().Idp()).IdentityName())
 	}
 
 	// Also if the sso-user-id annotation is set, then another additional identity is required if it is a different value to the Spec.UserID
-	if val, ok := userAcc.Annotations[toolchainv1alpha1.SSOUserIDAnnotationKey]; ok {
-		if val != userAcc.Spec.UserID {
-			expectedIdentities = append(expectedIdentities, commonidentity.NewIdentityNamingStandard(
-				userAcc.Annotations[toolchainv1alpha1.SSOUserIDAnnotationKey], config.Auth().Idp()).IdentityName())
-		}
+	if userAcc.Spec.PropagatedClaims.UserID != "" && userAcc.Spec.PropagatedClaims.UserID != userAcc.Spec.UserID {
```
alexeykazakov, line 248: Should be the following instead, right?
```suggestion
	if userAcc.Spec.PropagatedClaims.UserID != "" && userAcc.Spec.PropagatedClaims.UserID != userAcc.Spec.PropagatedClaims.Sub {
```

controllers/useraccount/useraccount_controller.go
```diff
@@ -635,13 +627,11 @@ func (r *Reconciler) updateStatusConditions(ctx context.Context, userAcc *toolch
...
-	if userAcc.Spec.OriginalSub != "" {
-		identities = append(identities, commonidentity.NewIdentityNamingStandard(userAcc.Spec.OriginalSub, config.Auth().Idp()).IdentityName())
+	if userAcc.Spec.PropagatedClaims.OriginalSub != "" {
+		identities = append(identities, commonidentity.NewIdentityNamingStandard(userAcc.Spec.PropagatedClaims.OriginalSub, config.Auth().Idp()).IdentityName())
 	}
-	if val, ok := userAcc.Annotations[toolchainv1alpha1.SSOUserIDAnnotationKey]; ok {
-		if val != userAcc.Spec.UserID {
-			identities = append(identities, commonidentity.NewIdentityNamingStandard(userAcc.Annotations[toolchainv1alpha1.SSOUserIDAnnotationKey], config.Auth().Idp()).IdentityName())
-		}
+	if userAcc.Spec.PropagatedClaims.UserID != "" && userAcc.Spec.PropagatedClaims.UserID != userAcc.Spec.PropagatedClaims.OriginalSub {
```
alexeykazakov, line 633: If I'm reading it right it should be:
```suggestion
	if userAcc.Spec.PropagatedClaims.UserID != "" && userAcc.Spec.PropagatedClaims.UserID != userAcc.Spec.PropagatedClaims.Sub {
```

controllers/useraccount/useraccount_controller.go
MatousJobanek: you could update the unit tests so the original fields are empty to make sure that only the new ones are used

controllers/useraccount/useraccount_controller.go
MatousJobanek: I'm wondering why unit tests didn't fail for that.

controllers/useraccount/useraccount_controller.go
sbryzak: Good catch, fixed thanks.

controllers/useraccount/useraccount_controller.go
sbryzak: Fixed

alexeykazakov (approved):

controllers/useraccount/useraccount_controller.go
MatousJobanek: @sbryzak but I don't see that the tests would be updated so the original fields would be set to empty to make sure that they are not used at all

MatousJobanek:
@sbryzak you missed some occurrences of the usages of the email annotation?
https://github.com/codeready-toolchain/member-operator/blob/2931fb137d4c73385324160a4538eb0647f142d2/controllers/useraccount/useraccount_controller.go#L397
https://github.com/codeready-toolchain/member-operator/blob/2931fb137d4c73385324160a4538eb0647f142d2/controllers/useraccount/useraccount_controller_test.go#L1554
https://github.com/codeready-toolchain/member-operator/blob/2931fb137d4c73385324160a4538eb0647f142d2/controllers/useraccount/useraccount_controller_test.go#L91

So, as I mentioned in the other comment: https://github.com/codeready-toolchain/member-operator/pull/507#discussion_r1407753800 could you please update the test so it doesn't use the old fields and annotations that we are migrating from?
https://github.com/codeready-toolchain/member-operator/blob/2931fb137d4c73385324160a4538eb0647f142d2/controllers/useraccount/useraccount_controller_test.go#L1609-L1614

PR #535, opened 2024-02-26
Ensure sandbox toleration

rajivnathan:
https://issues.redhat.com/browse/SANDBOX-499

Updates the member webhook to ensure the sandbox toleration is set on VM resources so that the VM can be scheduled on the BM node.
```
      tolerations:
      - effect: NoSchedule
        key: sandbox-cnv
        operator: Exists
```

e2e tests: https://github.com/codeready-toolchain/toolchain-e2e/pull/898

### DIFF
--- a/pkg/webhook/mutatingwebhook/vm_mutate.go
+++ b/pkg/webhook/mutatingwebhook/vm_mutate.go
@@ -21,6 +21,12 @@ const (
 
 var vmLogger = logf.Log.WithName("virtual_machines_mutating_webhook")
 
+var sandboxToleration = map[string]interface{}{
+	"effect":   "NoSchedule",
+	"key":      "sandbox-cnv",
+	"operator": "Exists",
+}
+
 func HandleMutateVirtualMachines(w http.ResponseWriter, r *http.Request) {
 	handleMutate(vmLogger, w, r, vmMutator)
 }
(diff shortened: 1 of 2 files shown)


pkg/webhook/mutatingwebhook/vm_mutate.go
```diff
@@ -247,6 +256,22 @@ func ensureLimits(unstructuredObj *unstructured.Unstructured, patchItems []map[s
...
+func ensureTolerations(unstructuredRequestObj *unstructured.Unstructured, patchItems []map[string]interface{}) []map[string]interface{} {
+	// get existing tolerations, if any
+	tolerations, _, err := unstructured.NestedSlice(unstructuredRequestObj.Object, "spec", "template", "spec", "tolerations")
+	if err != nil {
+		vmLogger.Error(err, "unable to get tolerations from VirtualMachine", "VirtualMachine", unstructuredRequestObj)
+		return patchItems
+	}
+
+	// no need to check original contents of tolerations, if there were existing tolerations the sandbox one will be appended; if there were no tolerations then the patch will add the sandbox toleration
+	tolerations = append(tolerations, sandboxToleration)
```
alexeykazakov, line 269: What if there is already a sandbox toleration. We will have two copies of the same toleration with the same key:
```
{
	"effect":   "NoSchedule",
	"key":      "sandbox-cnv",
	"operator": "Exists",
},
{
	"effect":   "NoSchedule",
	"key":      "sandbox-cnv",
	"operator": "Exists",
}
```
right? Or I'm reading it wrong?

rajivnathan: yeah, I thought about adding a check for that case but
1. the webhook is on create, so the toleration needs to be in the original VM for that to happen which is pretty unlikely
2. it works correctly even if it adds a duplicate (I tested it)

I mostly ignored it because of 1...but now that I think about it again it's possible that a user copies the VM yaml and creates another VM then it'd add the toleration again which is ugly. So I'll add a check and avoid adding a toleration if there's an existing toleration with the `sandbox-cnv` key.

alexeykazakov: Yeah... and if in the future we start invoking the webhook on updates too (not just on create) then every update on the VM resource would result in another duplicate of the toleration.

rajivnathan: Done a0ee25a991e676c0d928f6c64cf0b289aed2aeb6 . Thanks

alexeykazakov (approved):
Looks good. Thanks!

PR #538, opened 2024-03-05
chore: keeping our GitHub Actions up to date with Dependabot

xcoulon:
see docs at https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

see https://github.com/codeready-toolchain/sandbox-sre/pull/1578

### DIFF
--- a/controllers/memberoperatorconfig/memberoperatorconfig_controller_test.go
+++ b/controllers/memberoperatorconfig/memberoperatorconfig_controller_test.go
@@ -187,13 +187,13 @@ func TestHandleAutoscalerDeploy(t *testing.T) {
 		ctx := log.IntoContext(context.TODO(), controller.Log)
 
 		// when
-		cl.(*test.FakeClient).MockGet = func(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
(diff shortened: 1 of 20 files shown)


alexeykazakov (approved):

controllers/memberoperatorconfig/memberoperatorconfig_controller_test.go
```diff
@@ -187,13 +187,13 @@ func TestHandleAutoscalerDeploy(t *testing.T) {
 		ctx := log.IntoContext(context.TODO(), controller.Log)
 
 		// when
-		cl.(*test.FakeClient).MockGet = func(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+		cl.(*test.FakeClient).MockGet = func(_ context.Context, _ client.ObjectKey, _ client.Object, _ ...client.GetOption) error {
```
fbm3307, line 190: This is for me to understand, i think `ctx,key` etc were removed as they might no longer be useful. but just curious as to how does it help here.?

xcoulon: @fbm3307 Golang CI linter was complaining that these parameters were not used in the func, so by naming them `_`  we acknowledge that we don't use them - but we need them to match the signature of the `MockGet` func

mfrancisc (approved):
Thanks!

PR #551, opened 2024-03-25
SANDBOX-559 K8s 1 27 member

ranakan19:
Updates k8s dependencies to 0.27 and controller-runtime to v0.15.
With the update to controller-runtime v0.15 (changes in release detailed [here](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0)), this PR has following changes to combat the breaking changes:

- carry forward changes from api and toolchain-common - adding context in event handlers, change from runtime.Object to client.Object, and SubStatusResources
- updated fakeClient behaviour - https://github.com/kubernetes-sigs/controller-runtime/pull/2316 
- handle finalizers in fake client - https://github.com/kubernetes-sigs/controller-runtime/pull/1399 
- Watches signature has changed, For source.Source, now use WatchesRawSource - https://github.com/kubernetes-sigs/controller-runtime/pull/2135

Changes wrt - https://issues.redhat.com/browse/SANDBOX-559
DO NOT MERGE (All the PRs related to version updates need to be merged at once, but these PR are for early feedback so that there are not a lot of changes to review at once)


controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -1372,74 +1372,31 @@ func TestDeleteNSTemplateSet(t *testing.T) {
 		require.Empty(t, result)
 	})
 
-	t.Run("delete when there is no finalizer", func(t *testing.T) {
```
ranakan19, line 1375: Deleted this testcase because it was no longer possible to test this scenario which required to observe the object in a state when it has a deletion timestamp but no finalizer, because the object would get deleted before it can be observed.
If we still need this test case - we can try looking at [Interceptors](https://github.com/kubernetes-sigs/controller-runtime/pull/2306) in the following PR.

controllers/idler/idler_controller_test.go
```diff
@@ -80,9 +80,11 @@ func TestReconcile(t *testing.T) {
 		idler := &toolchainv1alpha1.Idler{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:              "being-deleted",
+				Finalizers:        []string{"toolchain.dev.openshift.com"},
 				DeletionTimestamp: &now,
 			},
 			Spec: toolchainv1alpha1.IdlerSpec{TimeoutSeconds: 30},
+			//Status: toolchainv1alpha1.IdlerStatus{Conditions: make([]toolchainv1alpha1.Condition, 0)},
```
alexeykazakov, line 87:
```suggestion
```

controllers/nstemplateset/cluster_resources.go
```diff
@@ -247,6 +247,12 @@ func (r *clusterResourcesManager) deleteClusterResource(ctx context.Context, nsT
 	if err := r.setStatusUpdatingIfNotProvisioning(ctx, nsTmplSet); err != nil {
 		return false, err
 	}
+	//remove finalizer before deleting
+	toDelete.SetFinalizers(nil)
```
alexeykazakov, line 251: Why?

controllers/nstemplateset/nstemplateset_controller.go
```diff
@@ -58,14 +58,14 @@ func (r *Reconciler) SetupWithManager(mgr manager.Manager, allNamespaceCluster r
 	mapToOwnerByLabel := handler.EnqueueRequestsFromMapFunc(commoncontroller.MapToOwnerByLabel("", toolchainv1alpha1.SpaceLabelKey))
 	build := ctrl.NewControllerManagedBy(mgr).
 		For(&toolchainv1alpha1.NSTemplateSet{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
-		Watches(&source.Kind{Type: &corev1.Namespace{}}, mapToOwnerByLabel).
-		Watches(source.NewKindWithCache(&rbac.Role{}, allNamespaceCluster.GetCache()), mapToOwnerByLabel, builder.WithPredicates(commonpredicates.LabelsAndGenerationPredicate{})).
-		Watches(source.NewKindWithCache(&rbac.RoleBinding{}, allNamespaceCluster.GetCache()), mapToOwnerByLabel, builder.WithPredicates(commonpredicates.LabelsAndGenerationPredicate{}))
+		Watches(&corev1.Namespace{}, mapToOwnerByLabel).
+		WatchesRawSource(source.Kind(allNamespaceCluster.GetCache(), &rbac.Role{}), mapToOwnerByLabel, builder.WithPredicates(commonpredicates.LabelsAndGenerationPredicate{})).
```
alexeykazakov, line 62: Just curious what is new here and why replacing `Watches` with `WatchesRawSource`?

controllers/useraccount/useraccount_controller_test.go
```diff
@@ -416,16 +414,15 @@ func TestReconcile(t *testing.T) {
...
 		r, req, cl, _ := prepareReconcile(t, username, cfg, userAcc, preexistingUser, preexistingIdentity)
 
 		t.Run("first reconcile deletes identity", func(t *testing.T) {
 			// given
 			// Set the deletionTimestamp
-			now := metav1.NewTime(time.Now())
-			userAcc.DeletionTimestamp = &now
-			err = r.Client.Update(context.TODO(), userAcc)
+			//now := metav1.NewTime(time.Now())
+			//userAcc.DeletionTimestamp = &now
```
alexeykazakov, line 424:
```suggestion
```

controllers/useraccount/useraccount_controller_test.go
```diff
@@ -566,18 +563,18 @@ func TestReconcile(t *testing.T) {
 
 	t.Run("useraccount is being deleted and has no users or identities - delete calls returns not found - then it should just remove finalizer and be removed from the client", func(t *testing.T) {
 		// given
-		userAcc := newUserAccount(username, userID)
-		util.AddFinalizer(userAcc, toolchainv1alpha1.FinalizerName)
+		userAcc := newUserAccount(username, userID, withFinalizer())
+		//util.AddFinalizer(userAcc, toolchainv1alpha1.FinalizerName)
```
alexeykazakov, line 567:
```suggestion
```

controllers/useraccount/useraccount_controller_test.go
```diff
@@ -566,18 +563,18 @@ func TestReconcile(t *testing.T) {
...
 			if obj.GetObjectKind().GroupVersionKind().Kind == "UserAccount" {
 				return apierros.NewNotFound(toolchainv1alpha1.GroupVersion.WithResource("UserAccount").GroupResource(), userAcc.Name)
 			}
 			return cl.Client.Delete(ctx, obj, opts...)
 		}
-		now := metav1.NewTime(time.Now())
-		userAcc.DeletionTimestamp = &now
-		err = r.Client.Update(context.TODO(), userAcc)
+		//now := metav1.NewTime(time.Now())
+		//userAcc.DeletionTimestamp = &now
```
alexeykazakov, line 576:
```suggestion
```

controllers/useraccount/useraccount_controller_test.go
```diff
@@ -626,9 +623,9 @@ func TestReconcile(t *testing.T) {
 			Get()
 
 		// Set the deletionTimestamp
-		now := metav1.NewTime(time.Now())
-		userAcc.DeletionTimestamp = &now
-		err = r.Client.Update(context.TODO(), userAcc)
+		//now := metav1.NewTime(time.Now())
+		//userAcc.DeletionTimestamp = &now
```
alexeykazakov, line 627:
```suggestion
```

controllers/useraccount/useraccount_controller_test.go
```diff
@@ -705,9 +702,9 @@ func TestReconcile(t *testing.T) {
...
```
alexeykazakov, line 706:
```suggestion
```

controllers/useraccount/useraccount_controller_test.go
```diff
@@ -744,9 +741,9 @@ func TestReconcile(t *testing.T) {
...
```
alexeykazakov, line 745:
```suggestion
```

controllers/useraccount/useraccount_controller_test.go
```diff
@@ -783,9 +780,9 @@ func TestReconcile(t *testing.T) {
...
```
alexeykazakov, line 784:
```suggestion
```

controllers/useraccount/useraccount_controller_test.go
```diff
@@ -833,9 +830,9 @@ func TestReconcile(t *testing.T) {
...
```
alexeykazakov, line 834:
```suggestion
```

pkg/consoleplugin/consolepluginserver_test.go
```diff
@@ -51,7 +52,7 @@ func TestConsolePluginServer(t *testing.T) {
 func waitForReady(t *testing.T) {
 	cl := http.Client{}
 
-	err := wait.Poll(DefaultRetryInterval, DefaultTimeout, func() (done bool, err error) {
+	err := wait.PollUntilContextTimeout(context.TODO(), DefaultRetryInterval, DefaultTimeout, false, func(ctx context.Context) (done bool, err error) {
```
ranakan19, line 55: Poll is being deprecated and the linter was complaining. However to fix this there was a typo in the function name that was suggested as replacement, which was later fixed. 
PTAL here if the context is used correctly - I'm not a 100% sure.
For reference please check - https://github.com/kubernetes/apimachinery/issues/153 and the fix here - https://github.com/kubernetes/kubernetes/pull/117143/files

controllers/idler/idler_controller_test.go
ranakan19: I didn't have to comment out the status - uncommented in [0b5289e](https://github.com/codeready-toolchain/member-operator/pull/551/commits/0b5289e17f8337c2eeba756dd94a524b87163fd0) instead of deleting

controllers/useraccount/useraccount_controller_test.go
ranakan19: done in [0b5289e](https://github.com/codeready-toolchain/member-operator/pull/551/commits/0b5289e17f8337c2eeba756dd94a524b87163fd0), thanks

controllers/useraccount/useraccount_controller_test.go
ranakan19: done in [0b5289e](https://github.com/codeready-toolchain/member-operator/pull/551/commits/0b5289e17f8337c2eeba756dd94a524b87163fd0), thanks

controllers/useraccount/useraccount_controller_test.go
ranakan19: done in [0b5289e](https://github.com/codeready-toolchain/member-operator/pull/551/commits/0b5289e17f8337c2eeba756dd94a524b87163fd0) thanks

controllers/useraccount/useraccount_controller_test.go
ranakan19: done in [0b5289e](https://github.com/codeready-toolchain/member-operator/pull/551/commits/0b5289e17f8337c2eeba756dd94a524b87163fd0), thanks

controllers/useraccount/useraccount_controller_test.go
ranakan19: done in [0b5289e](https://github.com/codeready-toolchain/member-operator/pull/551/commits/0b5289e17f8337c2eeba756dd94a524b87163fd0), thanks

controllers/useraccount/useraccount_controller_test.go
ranakan19: done in [0b5289e](https://github.com/codeready-toolchain/member-operator/pull/551/commits/0b5289e17f8337c2eeba756dd94a524b87163fd0), thanks

controllers/useraccount/useraccount_controller_test.go
ranakan19: done in [0b5289e](https://github.com/codeready-toolchain/member-operator/pull/551/commits/0b5289e17f8337c2eeba756dd94a524b87163fd0), thanks

controllers/useraccount/useraccount_controller_test.go
ranakan19: done in [0b5289e](https://github.com/codeready-toolchain/member-operator/pull/551/commits/0b5289e17f8337c2eeba756dd94a524b87163fd0), thanks

controllers/nstemplateset/cluster_resources.go
ranakan19: because the object will not be deleted untill there is a finalizer. fakeClient.Delete now skips the deletion if a finalizer is present.

controllers/nstemplateset/nstemplateset_controller.go
ranakan19: please note that not all `Watches` are replaced with `WatchesRawSource`, only the one where there was cache. Because with the removal of Source as argument in the Watches signature, the closest replacement was `WatchesRawSource`.
If only metadata is enough, then we can use `WatchesMetadata` instead but i didn't see that to be the case, please let me know if you think otherwise. 
For reference please see this PR in controller-runtime - https://github.com/kubernetes-sigs/controller-runtime/pull/2135/files

controllers/nstemplateset/cluster_resources.go
alexeykazakov: Hm... but in the runtime, isn't the expected workflow is the following:
- If there is no finalizer then add one (if the resource is not already being deleted).
- If there is a finalizer present and the resource is being deleted (there is a deletion timestamp) then do all required cleanup and then delete the finalizer to unblock the actual resource deletion by the garbage collector. Otherwise the resource will be stuck in the "terminating" if a finalizesr is present.

Or I'm missing something?

ranakan19: As i understand it:
- it doesn't add a finalizer if not present, rather it errors out if a finalizer is not present when trying to create a resource with deletion timestamp.
- When a finalizer is present and Delete() is called - it adds a deletion timestamp and updates as [here](https://github.com/lleshchi/controller-runtime/blob/7a66d580c0c53504f5b509b45e9300cc18a1cc30/pkg/client/fake/client.go#L1068-L1073). And update actually deletes the resource only when there is no finalizer as [here](https://github.com/lleshchi/controller-runtime/blob/7a66d580c0c53504f5b509b45e9300cc18a1cc30/pkg/client/fake/client.go#L454-L456). Thus the action of removing the finalizer is to be done in our test. 
When i draw a parallel to the real client as well - we remove the finalizers in our controllers after all the required cleanup.

MatousJobanek: @ranakan19 but what you are changing here is the production code, not a test. We shouldn't remove any finalizers we are not responsible for - then only finalizer we can add/remove (in production code) is the toolchain one. Nothing else.
If you need to do some workarounds around finalizers to make FakeClient working, then do it in test files, not in the production code.

ranakan19: I mislooked that it not in the test, i'll make the necessary changes

ranakan19: @alexeykazakov @MatousJobanek removed in [b1816ef](https://github.com/codeready-toolchain/member-operator/pull/551/commits/b1816ef8297f6b49fba73cf142878768e514bae1)

controllers/nstemplateset/nstemplateset_controller_test.go
```diff
@@ -1372,74 +1372,31 @@ func TestDeleteNSTemplateSet(t *testing.T) {
...
-		stageNS := newNamespace("advanced", spacename, "stage", withTemplateRefUsingRevision("abcde11"))
+		devNS := newNamespace("advanced", spacename, "dev", withTemplateRefUsingRevision("abcde11"), withFinalizer())
+		stageNS := newNamespace("advanced", spacename, "stage", withTemplateRefUsingRevision("abcde11"), withFinalizer())
 
-		r, fakeClient := prepareController(t, nsTmplSet, devNS, stageNS)
+		r, _ := prepareController(t, nsTmplSet, devNS, stageNS)
 		req := newReconcileRequest(namespaceName, spacename)
 
-		// only add deletion timestamp, but not delete
-		fakeClient.MockDelete = func(ctx context.Context, obj client.Object, opts ...client.DeleteOption) error {
```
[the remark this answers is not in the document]
ranakan19: there is no problem with this test case, I am removing the MockDelete function here because it is not needed anymore. The fakeClient would do only add a deletionTimestamp now since adding the finalizer.

controllers/nstemplateset/nstemplateset_controller_test.go
ranakan19: Yes, precisely.

alexeykazakov (approved):

MatousJobanek (commented):
Looks good, but I have a few comments on the versions in go.mod file.

Also, is there a reason why you picked the `v0.27.2` specifically, and not the latest `v0.27.x` patch version that is available at the time when we are doing the update? This is just a detail since we are going to bump the version to the next minor one soon again.

PR #562, opened 2024-04-26
Rename member webhook cluster scoped objects

mfrancisc:
This PR contains:
- renaming of the cluster scoped objects in the webhook deployments. It appends the member namespace in all the cluster scoped resources.
- temporary `old-name` label used to delete the objects using the current name ( since we are redeploying with new name )
- temporary migration logic to delete current objects in favor of the ones with new name

The renaming of the cluster scoped objects is needed in order to be able to delete the member2 webhook configuration during e2e tests. 
We cannot avoid deploying the webhook on member2 anymore since we inverted the order of the make targets in https://github.com/codeready-toolchain/toolchain-e2e/pull/950/files#diff-163703aeb25751c1605d305fbfdf39f3a9dfb74d2ac9d8a03f5cf0169d7d41b7R266 , due to the fact that member operator is now deploying the toolchaincluster-sa and we need to verify and use the toolchaincluster-member SA in `setup-toolchainclusters` only after deploy the member operator.

Paired with: 
- toolchain-e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/950

see: https://redhat-internal.slack.com/archives/C06MJ2DVBU4/p1713534677202559


controllers/memberoperatorconfig/memberoperatorconfig_controller.go
```diff
@@ -105,9 +106,30 @@ func (r *Reconciler) handleWebhookDeploy(ctx context.Context, cfg membercfg.Conf
...
+		}
+		// TODO -- end temporary migration
 		logger.Info("(Re)Deployed users' pods webhook")
 	} else {
-		logger.Info("Skipping deployment of users' pods webhook")
+		memberNS := &corev1.Namespace{}
+		if err := r.Client.Get(ctx, types.NamespacedName{Name: namespace}, memberNS); err != nil {
+			return errs.Wrapf(err, "cannot get member namespace %s", memberNS)
+		}
+		if value, found := memberNS.Labels["toolchain.dev.openshift.com/webhook"]; found && value == "disabled" {
```
MatousJobanek, line 124: ok, I now see why you added the label in the e2e tests, but it doesn't answer why we need it here. Disabling of the webhook should be fully managed by ToolchainConfig/MemberConfig CR.

pkg/webhook/deploy/deployment.go
```diff
@@ -46,6 +53,51 @@ func Webhook(ctx context.Context, cl runtimeclient.Client, s *runtime.Scheme, na
 	return nil
 }
 
+// Delete deletes the webhook app if it's deployed. Does nothing if it's not.
+// Returns true if the app was deleted.
+func Delete(ctx context.Context, cl runtimeclient.Client, s *runtime.Scheme, namespace string, oldObjectOnly bool) (bool, error) {
+	objs, err := getTemplateObjects(s, namespace, "0", []byte{00000001})
```
MatousJobanek, line 59: it's probably just a detail, I know that it's not being applied anywhere - I wouldn't use zero `0` as a value for the image url just to be sure that is not expressed as a number while the image location should be a string
```suggestion
	objs, err := getTemplateObjects(s, namespace, "dummy-image", []byte{00000001})
```
apart from that, this looks more natural than the zero :-)

pkg/webhook/deploy/deployment.go
```diff
@@ -46,6 +53,51 @@ func Webhook(ctx context.Context, cl runtimeclient.Client, s *runtime.Scheme, na
...
+		return false, err
+	}
+
+	var deleted bool
+	for _, obj := range objs {
+		unst := &unstructured.Unstructured{}
+		unst.SetGroupVersionKind(obj.GetObjectKind().GroupVersionKind())
+		objName := obj.GetName()
+		_, doNotDeleteFound := obj.GetLabels()["toolchain.dev.openshift.com/no-deletion"]
+		if doNotDeleteFound {
```
MatousJobanek, line 70: I would really use an annotation for it - the purpose of annotations is to provide an additional context that can influence the execution/process of the resources. Label is used mainly for matching.
I believe that this is more closer to the first usage.

pkg/webhook/deploy/deployment.go
```diff
@@ -46,6 +53,51 @@ func Webhook(ctx context.Context, cl runtimeclient.Client, s *runtime.Scheme, na
...
+			// this object needs to stay
+			continue
+		}
+		// TODO --- temporary migration step to delete the objects by using the old name
+		if oldObjectOnly {
+			oldName, found := obj.GetLabels()["toolchain.dev.openshift.com/old-name"]
```
MatousJobanek, line 76: the key that is used in the manifests is `old-name` - without the `toolchain.dev.openshift.com/` prefix

pkg/webhook/deploy/deployment.go
```diff
@@ -46,6 +53,51 @@ func Webhook(ctx context.Context, cl runtimeclient.Client, s *runtime.Scheme, na
...
+			if !errors.IsNotFound(err) { // Ignore not found
+				return false, errs.Wrap(err, "cannot get webhook object")
+			}
+		} else {
+			log.Info(fmt.Sprintf("Deleting %s  name:%s namespace:%s", obj.GetObjectKind().GroupVersionKind(), objName, obj.GetNamespace()))
+			if err := cl.Delete(ctx, unst); err != nil {
+				return false, errs.Wrap(err, "cannot delete webhook object")
+			}
+			deleted = true
+		}
```
MatousJobanek, line 95: could be also:
```suggestion
			}
			continue
		}
		log.Info(fmt.Sprintf("Deleting %s  name:%s namespace:%s", obj.GetObjectKind().GroupVersionKind(), objName, obj.GetNamespace()))
		if err := cl.Delete(ctx, unst); err != nil {
			return false, errs.Wrap(err, "cannot delete webhook object")
		}
		deleted = true
```

pkg/webhook/deploy/deployment.go
```diff
@@ -46,6 +53,51 @@ func Webhook(ctx context.Context, cl runtimeclient.Client, s *runtime.Scheme, na
...
+		if oldObjectOnly {
+			oldName, found := obj.GetLabels()["toolchain.dev.openshift.com/old-name"]
+			if !found {
+				// this object needs to stay
+				continue
+			}
+			objName = oldName
+		}
+		// TODO --- end temporary migration step
+		log.Info(fmt.Sprintf("Seraching for object to delete: %s  name:%s namespace:%s", obj.GetObjectKind().GroupVersionKind(), objName, obj.GetNamespace()))
```
MatousJobanek, line 84:
```suggestion
		logger := logf.FromContext(ctx).WithName("webhook_deploy").WithValues("gvk", obj.GetObjectKind().GroupVersionKind(), "name", objName, "namespace", obj.GetNamespace())
		logger.Info("Searching for object to delete")
```

MatousJobanek (commented):
Could you please add unit tests for the deletion part of the code?

pkg/webhook/deploy/deployment.go
mfrancisc: yup , I agree with removing `else` statements when possible.

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
mfrancisc: good point, I've changed this in https://github.com/codeready-toolchain/member-operator/pull/562/commits/6f50ffab500165de806d03edc0b2b28807fd8d9c

pkg/webhook/deploy/deployment.go
mfrancisc: Makes sense, annotation is better in this case. Thanks!

I've converted those in https://github.com/codeready-toolchain/member-operator/pull/562/commits/6f50ffab500165de806d03edc0b2b28807fd8d9c

pkg/webhook/deploy/deployment.go
mfrancisc: Great catch 🤦‍♂️ 

I've added the prefix in all annotations.

mfrancisc:
> Could you please add unit tests for the deletion part of the code?

@MatousJobanek thanks for your review. I've added a couple of additional unit tests to cover:
- migration , when old objects are replaced by new ones
- e2e tests, when all the objects ( except the ones marked as no-deletion) are deleted

pkg/webhook/deploy/deployment.go
```diff
@@ -3,11 +3,19 @@ package deploy
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"testing"
```
rajivnathan, line 8: The test functions should be moved to the test file unless I'm missing something

rajivnathan (commented):
Can you please revert the VM-related changes for this PR to ensure it has no impact?

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
```diff
@@ -105,9 +104,24 @@ func (r *Reconciler) handleWebhookDeploy(ctx context.Context, cfg membercfg.Conf
...
 		logger.Info("(Re)Deployed users' pods webhook")
 	} else {
-		logger.Info("Skipping deployment of users' pods webhook")
+		deleted, err := deploy.Delete(ctx, r.Client, r.Client.Scheme(), namespace, false)
+		if err != nil {
+			return err
+		}
+		if deleted {
+			logger.Info("Deleted previously deployed webhook app")
+		}
```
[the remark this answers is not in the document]
mfrancisc: Yes, because we now deploy the member operator before we disable the configuration see: https://github.com/codeready-toolchain/toolchain-e2e/pull/950/files#diff-163703aeb25751c1605d305fbfdf39f3a9dfb74d2ac9d8a03f5cf0169d7d41b7R292 . 

See also my comment here : https://github.com/codeready-toolchain/toolchain-e2e/pull/950/files#r1580269009

pkg/webhook/deploy/deployment.go
mfrancisc: Good point! done in https://github.com/codeready-toolchain/member-operator/pull/562/commits/e07e190754ac98ee8781e514a63c1053cc16e628 thanks!

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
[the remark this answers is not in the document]
rajivnathan: Thanks for the explanation, I see your point now. I think the webhook provides some pretty important functionality that I'm not sure we'd want to delete unless it's explicitly meant to be deleted but I can't think of a reason why we'd disable the configuration and expect the webhook to stay as it was so I guess it's fine. I'd just suggest we update [the comment in the api](https://github.com/codeready-toolchain/api/blob/49462fdf229a6a392cae4bd4d4f4b551e8870f35/api/v1alpha1/memberoperatorconfig_types.go#L158) to make it clear what happens.

rajivnathan (approved):

controllers/memberoperatorconfig/memberoperatorconfig_controller.go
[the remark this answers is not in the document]
mfrancisc: > I'd just suggest we update [the comment in the api](https://github.com/codeready-toolchain/api/blob/49462fdf229a6a392cae4bd4d4f4b551e8870f35/api/v1alpha1/memberoperatorconfig_types.go#L158) to make it clear what happens.

good point, please see https://github.com/codeready-toolchain/api/pull/425/files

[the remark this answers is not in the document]
mfrancisc: > I think the webhook provides some pretty important functionality that I'm not sure we'd want to delete unless it's explicitly meant to be deleted but I can't think of a reason why we'd disable the configuration and expect the webhook to stay

I was discussing this with @MatousJobanek, a better approach IMO would be to decouple the deployment of the webhooks from member operator , this would allow to have more flexibility and would avoid such "workarounds" where we deploy something and then we need to delete it. But the decoupling of the webhooks is a bigger effort and not on the priority list atm.

rajivnathan (approved):

MatousJobanek (approved):
Nice :+1:

pkg/webhook/deploy/deployment.go
```diff
@@ -46,7 +59,53 @@ func Webhook(ctx context.Context, cl runtimeclient.Client, s *runtime.Scheme, na
...
+		// TODO --- end temporary migration step
+		logger := logf.FromContext(ctx).WithName("webhook_deploy").WithValues("gvk", obj.GetObjectKind().GroupVersionKind(), "name", objName, "namespace", obj.GetNamespace())
+		logger.Info("Searching for object to delete")
+		if err := cl.Get(ctx, types.NamespacedName{Namespace: obj.GetNamespace(), Name: objName}, unst); err != nil {
+			if !errors.IsNotFound(err) { // Ignore not found
+				return false, errs.Wrap(err, "cannot get webhook object")
+			}
+			continue
+		}
+		log.Info(fmt.Sprintf("Deleting %s  name:%s namespace:%s", obj.GetObjectKind().GroupVersionKind(), objName, obj.GetNamespace()))
```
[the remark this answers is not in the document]
mfrancisc: Good point! Thanks!

PR #568, opened 2024-05-14
remove console resources, replace uuid library

ranakan19:
fix wrt https://issues.redhat.com/browse/SANDBOX-605 
Adds removal of console resources - configmap, role and rolebinding - when user is removed.


controllers/useraccount/useraccount_controller.go
```diff
@@ -475,6 +484,13 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
 	logger := log.FromContext(ctx)
 	logger.Info("deleting the User resources")
 
+	//get the UID before deleting the user
+	// TO DO: this is a temporary solution to remove resources created for users logging in from OIDC which don't have owner references after 4.15
+	uid := userList[0].UID
+	logger.Info(fmt.Sprintf("checking for user settings resources with UID [%s] to be deleted", uid))
+	if deleted, err := r.deleteUserResources(ctx, string(uid)); err != nil || deleted {
+		return deleted, err
```
alexeykazakov, line 492: I don't think we can return here. The idea behind deleting the User resource first and stop the reconcile (by return an empty result here: https://github.com/codeready-toolchain/member-operator/pull/568/files#diff-bda4e43a8c3e78834044bcb623b8e1a57e184bec0d940e8dd0f534446796ab69R139 ) is the following:
- we delete the user and stop
- since the controller watches Users it will get triggered again when the User is deleted
- we delete the identities
- since the controller watches Identities it will get triggered again when the Identity is deleted

But it won't work with user setting resources because we do not watch them. So I think we should proceed with deleting the User resource here and return only in case of an error.

controllers/useraccount/useraccount_controller.go
```diff
@@ -484,6 +500,93 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
 	return true, nil
 }
 
+// Returns `true` if the associated resources (configMap, role and role-binding) created for a user by console is deleted, `false` otherwise with the underlying error.
```
alexeykazakov, line 503: It can happen that there is nothing to delete. And it's not an error. Correct?
Basically it's the following:
- `return true, nil`. Deleted w/o error.
- `return false, nil`. Nothing to delete.
- `return false, err`. Tried to delete existing resources but failed.

Right?
```suggestion
// Returns `true` if the associated resources (configMap, role and role-binding) created for a user by console are deleted, `false` otherwise.
```

controllers/useraccount/useraccount_controller.go
```diff
@@ -484,6 +500,93 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
...
+// This function only looks for these resources in the namespace - openshift-console-user-settings -
+func (r *Reconciler) deleteUserResources(ctx context.Context, userID string) (bool, error) {
+
+	// new user will have a label introduced which can help retrieve resources, otherwise older ones will have to rely on the name of the resource being of type `user-settings-<UID>`
+	// delete ConfigMap, Role and RoleBinding
```
alexeykazakov, line 508:
```suggestion
	// Users which were created in the cluster with the OCP versions which includes https://issues.redhat.com/browse/OCPBUGS-32321 fix
	// will have a label which will help to map the User to the User settings resources.
	// User which were created before that won't have that label and we have to rely on the name of the resource being of type `user-settings-<UID>`,
	// where <UID> is the User UID.

	// Delete ConfigMap, Role and RoleBinding
```

controllers/useraccount/useraccount_controller.go
```diff
@@ -484,6 +500,93 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
 	return true, nil
 }
 
+// Returns `true` if the associated resources (configMap, role and role-binding) created for a user by console is deleted, `false` otherwise with the underlying error.
+// This function only looks for these resources in the namespace - openshift-console-user-settings -
+func (r *Reconciler) deleteUserResources(ctx context.Context, userID string) (bool, error) {
```
alexeykazakov, line 505: Just to make it clear that it's not the SSO `UserID` but the `User.metadata.uid`.
```suggestion
func (r *Reconciler) deleteUserResources(ctx context.Context, userUID string) (bool, error) {
```

controllers/useraccount/useraccount_controller.go
```diff
@@ -484,6 +500,93 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
...
+	if deleted, err := r.deleteConfigMap(ctx, userID); err != nil || deleted {
+		return deleted, err
+	}
+	if deleted, err := r.deleteRole(ctx, userID); err != nil || deleted {
+		return deleted, err
+	}
+	if deleted, err := r.deleteRoleBinding(ctx, userID); err != nil || deleted {
+		return deleted, err
+	}
+	return false, nil
```
alexeykazakov, line 518: Same concern as above. We do not watch CMs, RoleBindings, and roles here. We can't delete them one by one. We either have to explicitly re-reconcile in this case or just delete them at once and do not stop the reconcile loop.
I think returning true/fails if any of the resources were deleted doesn't make much sense it this case. I would just return an error if something fails. Otherwise try to delete everything at the same loop.

controllers/useraccount/useraccount_controller.go
```diff
@@ -484,6 +500,93 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
...
+}
+
+func (r *Reconciler) deleteConfigMap(ctx context.Context, uid string) (bool, error) {
```
alexeykazakov, line 521:
```suggestion
func (r *Reconciler) deleteConfigMap(ctx context.Context, userUID string) (bool, error) {
```
Just in case it's not clear. `UID` means unique identifier. Not user ID :) Every kube resource has an UID. So we need to make it clear that it's the corresponding user's UID. Not the Config Map UID.

controllers/useraccount/useraccount_controller.go
```diff
@@ -31,6 +33,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+const (
+	ConsoleUserSettingsIdentifier = "console.openshift.io/user-settings"
+	ConsoleUserSettingsUID        = "console.openshift.io/user-settings-uid"
+	UserSettingNS                 = "openshift-console-user-settings"
+	ConsoleUserSettingsPrefix     = "user-settings-"
```
alexeykazakov, line 40:
```suggestion
	ConsoleUserSettingsResourceNamePrefix     = "user-settings-"
```

controllers/useraccount/useraccount_controller.go
```diff
@@ -675,3 +778,61 @@ func getUsersByOwnerName(ctx context.Context, cl client.Client, owner string) ([
...
+	rbList := &v1.RoleBindingList{}
+	labels := map[string]string{ConsoleUserSettingsIdentifier: "true", ConsoleUserSettingsUID: uid}
+	err := cl.List(ctx, rbList, client.MatchingLabels(labels), client.InNamespace(UserSettingNS))
+	if err != nil {
+		return []v1.RoleBinding{}, err
+	}
+	return rbList.Items, nil
+}
+
+func getObjectByName(ctx context.Context, cl client.Client, name string, object client.Object) (client.Object, error) {
```
alexeykazakov, line 812: Let's return `nil, nil` if the object not found (in this case the client will return a NotFound error which you just ignore). And `nil, err` if there is another error (not NotFound). Also add a description to the function explaining it.

Also let's rename it to something like `getConsoleUserSettingObjectByName()` or something. So it's clear that it's specific to the console user settings.
And maybe let's move all console user setting specific functions like getXXXObjects, delete...() and the constants to a different file? `usersettings.go` within the same package? So we don't have too big useraccount_controller.go file?

controllers/useraccount/useraccount_controller.go
```diff
@@ -484,6 +500,93 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
...
+func (r *Reconciler) deleteConfigMap(ctx context.Context, uid string) (bool, error) {
+	name := ConsoleUserSettingsPrefix + uid
+	logger := log.FromContext(ctx)
+	logger.Info(fmt.Sprintf("deleting configmap with name %s", name))
+	toDelete := &corev1.ConfigMap{}
+	toDelete.SetGroupVersionKind(schema.GroupVersionKind{
+		Kind: "ConfigMap",
+	})
+	configMap, err := getObjectByName(ctx, r.Client, name, toDelete)
+	if err != nil || configMap == nil {
```
alexeykazakov, line 530: I don't think we should hide the error here. We should ignore 404 errors (see above) but otherwise return the error here. And only if we didn't find the object to delete and there is no error (again, we ignore 404) we should try to use the label.

controllers/useraccount/useraccount_controller.go
```diff
@@ -484,6 +500,93 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
...
+		// try with label
+		if configMapList, err := getConfigMapByLabel(ctx, r.Client, uid); err != nil || len(configMapList) == 0 {
+			return false, err
```
alexeykazakov, line 533: And here too. Not found? No other error? Ignore and move on.
Another error? Stop. Return the error.
Found more then one CM with that label? Delete all of them???

controllers/useraccount/useraccount_controller.go
```diff
@@ -484,6 +500,93 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
...
+	} else {
+		toDelete = configMap.(*corev1.ConfigMap)
+	}
+	if err := r.Client.Delete(ctx, toDelete); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+func (r *Reconciler) deleteRole(ctx context.Context, uid string) (bool, error) {
```
alexeykazakov, line 546: Same as above.

controllers/useraccount/useraccount_controller.go
```diff
@@ -484,6 +500,93 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
...
+	} else {
+		toDelete = role.(*v1.Role)
+	}
+	if err := r.Client.Delete(ctx, toDelete); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+func (r *Reconciler) deleteRoleBinding(ctx context.Context, uid string) (bool, error) {
```
alexeykazakov, line 568: Same as above.

controllers/useraccount/useraccount_controller.go
```diff
@@ -475,6 +484,13 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
 	logger := log.FromContext(ctx)
 	logger.Info("deleting the User resources")
 
+	//get the UID before deleting the user
+	// TO DO: this is a temporary solution to remove resources created for users logging in from OIDC which don't have owner references after 4.15
```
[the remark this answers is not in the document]
ranakan19: fixed in [1bf005e](https://github.com/codeready-toolchain/member-operator/pull/568/commits/1bf005eeb31a8b6968271c667f5b5777c5f4c969)

controllers/useraccount/useraccount_controller.go
ranakan19: makes sense, thanks for explaining this. I've made the change here -[1bf005e](https://github.com/codeready-toolchain/member-operator/pull/568/commits/1bf005eeb31a8b6968271c667f5b5777c5f4c969) - to delete all the user resources in the same reconcile before deleting User, and only returns if there is an error.

controllers/useraccount/useraccount_controller.go
ranakan19: fixed the comment in afd558d

controllers/useraccount/useraccount_controller.go
ranakan19: done in [de45e6b](https://github.com/codeready-toolchain/member-operator/pull/568/commits/de45e6b31fcd13fe55a985f7614164dfb008a934)

controllers/useraccount/useraccount_controller.go
ranakan19: done in [de45e6b](https://github.com/codeready-toolchain/member-operator/pull/568/commits/de45e6b31fcd13fe55a985f7614164dfb008a934)

controllers/useraccount/useraccount_controller.go
ranakan19: done in [1bf005e](https://github.com/codeready-toolchain/member-operator/pull/568/commits/1bf005eeb31a8b6968271c667f5b5777c5f4c969)

controllers/useraccount/useraccount_controller.go
ranakan19: done in 3c748d9

controllers/useraccount/useraccount_controller.go
```diff
@@ -475,6 +475,13 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
 	logger := log.FromContext(ctx)
 	logger.Info("deleting the User resources")
 
+	//get the UID before deleting the user
+	// TO DO: this is a workaround for breaking change introduced by console with upgrade to 4.15. This remove resources created for users logging in from OIDC which don't have owner references after 4.15
```
alexeykazakov, line 479:
```suggestion
	// TO DO: this is a workaround for breaking change introduced by console with upgrade to 4.15. This remove resources created for users logging in from OIDC which don't have owner references starting from OCP 4.15 version.
```

controllers/useraccount/useraccount_controller.go
```diff
@@ -484,6 +491,28 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
 	return true, nil
 }
 
+// Returns `true` if the associated resources (configMap, role and role-binding) created for a user by console are deleted, `false` otherwise.
```
alexeykazakov, line 494: It returns an error, not bool :)

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,204 @@
...
+	} else if configMap != nil {
+		toDelete = append(toDelete, configMap)
+	} else {
+		// there was some kind of error in getting
+		return false, err
+	}
+	if err := deleteResources(ctx, cl, toDelete); err != nil {
+		return false, err
+	}
+	return true, nil
```
alexeykazakov, line 56: If there is no CMs (by name of by label) then there `toDelete()` will be empty. In this case do not return `true, nil`. Return `false, nil`.

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,204 @@
...
+	} else if role != nil {
+		toDelete = append(toDelete, role)
+	} else {
+		// there was some kind of error in getting
+		return false, err
+	}
+	if err := deleteResources(ctx, cl, toDelete); err != nil {
+		return false, err
+	}
+	return true, nil
```
alexeykazakov, line 90: Same here.

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,204 @@
...
+	} else if rb != nil {
+		toDelete = append(toDelete, rb)
+	} else {
+		// there was some kind of error in getting
+		return false, err
+	}
+	if err := deleteResources(ctx, cl, toDelete); err != nil {
+		return false, err
+	}
+	return true, nil
```
alexeykazakov, line 124: Same here.

controllers/useraccount/console_usersettings.go
alexeykazakov: Actually I see that you don't use that bool at all. If you don't use it in the controller why adding it here?
Return an error only. Don't return a bool too. And update the function description accordingly.

controllers/useraccount/useraccount_controller_test.go
```diff
@@ -477,6 +478,262 @@ func TestReconcile(t *testing.T) {
...
+					useraccount.AssertThatUserAccount(t, username, r.Client).
+						DoesNotExist()
+					require.Equal(t, 0, *mockCallsCounter)
+				})
+
+			})
+		})
+	})
+
+	//t.Run("delete useraccount with console resources having labels removes subsequent resources", func(t *testing.T) {
```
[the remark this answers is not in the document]
ranakan19: commented our test has been removed here [0fa343c](https://github.com/codeready-toolchain/member-operator/pull/568/commits/0fa343cdc02bb83e592d5d4423113147121f947d) . New test for `deleteUserResources` to come soon

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,204 @@
...
+	roleList := &v1.RoleList{}
+	labels := map[string]string{ConsoleUserSettingsIdentifier: "true", ConsoleUserSettingsUID: uid}
+	err := cl.List(ctx, roleList, client.MatchingLabels(labels), client.InNamespace(UserSettingNS))
+	if err != nil {
+		return []v1.Role{}, err
+	}
+	return roleList.Items, nil
+}
+
+func getRoleBindingsByLabel(ctx context.Context, cl client.Client, uid string) ([]v1.RoleBinding, error) {
```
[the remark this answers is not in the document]
ranakan19: done in [b18361f](https://github.com/codeready-toolchain/member-operator/pull/568/commits/b18361f20525f59ffaf7218e3798bcf1df0bee1e)

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,204 @@
...
+	configMapList := &corev1.ConfigMapList{}
+	labels := map[string]string{ConsoleUserSettingsIdentifier: "true", ConsoleUserSettingsUID: uid}
+	err := cl.List(ctx, configMapList, client.MatchingLabels(labels), client.InNamespace(UserSettingNS))
+	if err != nil {
+		return []corev1.ConfigMap{}, err
+	}
+	return configMapList.Items, nil
+}
+
+func getRolesByLabel(ctx context.Context, cl client.Client, uid string) ([]v1.Role, error) {
```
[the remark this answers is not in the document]
ranakan19: done in [b18361f](https://github.com/codeready-toolchain/member-operator/pull/568/commits/b18361f20525f59ffaf7218e3798bcf1df0bee1e)

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,204 @@
...
+		// there was some kind of error in getting
+		return false, err
+	}
+	if err := deleteResources(ctx, cl, toDelete); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+func getConfigMapByLabel(ctx context.Context, cl client.Client, uid string) ([]corev1.ConfigMap, error) {
```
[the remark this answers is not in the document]
ranakan19: done in [b18361f](https://github.com/codeready-toolchain/member-operator/pull/568/commits/b18361f20525f59ffaf7218e3798bcf1df0bee1e)

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,204 @@
...
+	ConsoleUserSettingsUID                = "console.openshift.io/user-settings-uid"
+	UserSettingNS                         = "openshift-console-user-settings"
+	ConsoleUserSettingsResourceNamePrefix = "user-settings-"
+)
+
+// deleteConfigMap deletes a ConfigMap associated with a user from console setting.
+// It first attempts to find the ConfigMap by name. If not found, it then looks for the ConfigMap by label.
+// If multiple ConfigMaps are found by label, all of them are deleted.
+// The function returns a boolean indicating whether the deletion was successful and an error if any occurred.
+func deleteConfigMap(ctx context.Context, cl client.Client, uid string) (bool, error) {
```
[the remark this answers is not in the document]
ranakan19: done in [b18361f](https://github.com/codeready-toolchain/member-operator/pull/568/commits/b18361f20525f59ffaf7218e3798bcf1df0bee1e)

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,204 @@
...
+		return false, err
+	}
+	return true, nil
+}
+
+// deleteRole deletes a Role associated with a user from console setting.
+// It first attempts to find the Role by name. If not found, it then looks for the Role by label.
+// If multiple Roles are found by label, all of them are deleted.
+// The function returns a boolean indicating whether the deletion was successful and an error if any occurred.
+func deleteRole(ctx context.Context, cl client.Client, uid string) (bool, error) {
```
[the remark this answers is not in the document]
ranakan19: done in [b18361f](https://github.com/codeready-toolchain/member-operator/pull/568/commits/b18361f20525f59ffaf7218e3798bcf1df0bee1e)

controllers/useraccount/console_usersettings.go
ranakan19: done in 08d0516

controllers/useraccount/console_usersettings.go
ranakan19: done in 08d0516

controllers/useraccount/console_usersettings.go
ranakan19: done in 08d0516

controllers/useraccount/useraccount_controller.go
ranakan19: great catch, fixed in [a224e71](https://github.com/codeready-toolchain/member-operator/pull/568/commits/a224e71fffd02f8735473c20088feb02a2eaa6b0)

controllers/useraccount/useraccount_controller.go
ranakan19: fixed in [a224e71](https://github.com/codeready-toolchain/member-operator/pull/568/commits/a224e71fffd02f8735473c20088feb02a2eaa6b0)

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,204 @@
...
+			for key := range configMapList {
+				toDelete = append(toDelete, &configMapList[key])
+			}
+		}
+	} else if configMap != nil {
+		toDelete = append(toDelete, configMap)
+	} else {
+		// there was some kind of error in getting
+		return err
+	}
```
mfrancisc, line 52: minor - another approach would be to check the error as first thing:
```
if err != nil {
    return err
}
// the other checks here ..

```

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,204 @@
...
+			for key := range roleList {
+				toDelete = append(toDelete, &roleList[key])
+			}
+		}
+	} else if role != nil {
+		toDelete = append(toDelete, role)
+	} else {
+		// there was some kind of error in getting
+		return err
+	}
```
mfrancisc, line 86: same as above

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,204 @@
...
+			for key := range rbList {
+				toDelete = append(toDelete, &rbList[key])
+			}
+		}
+	} else if rb != nil {
+		toDelete = append(toDelete, rb)
+	} else {
+		// there was some kind of error in getting
+		return err
+	}
```
mfrancisc, line 120: same here

controllers/useraccount/console_usersettings_test.go
```diff
@@ -0,0 +1,449 @@
...
+				TypeMeta: metav1.TypeMeta{Kind: "ConfigMap"},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "user-settings-johnsmith",
+					Namespace: UserSettingNS,
+				},
+			}
+			cl := fake.NewClientBuilder().WithObjects(cm).Build()
+
+			err := deleteConfigMap(ctx, cl, "johnsmith")
+			require.NoError(t, err)
```
mfrancisc, line 177: I saw that you check the abscence of these objects in the useraccount_controller_test but would it be worth checking that the config map is gone here as well ?

controllers/useraccount/console_usersettings_test.go
```diff
@@ -0,0 +1,449 @@
...
+					Namespace: UserSettingNS,
+					Labels: map[string]string{
+						ConsoleUserSettingsIdentifier: "true",
+						ConsoleUserSettingsUID:        "johnsmith",
+					},
+				},
+			}
+			cl := fake.NewClientBuilder().WithObjects(cm).Build()
+			err := deleteConfigMap(context.TODO(), cl, "johnsmith")
+			require.NoError(t, err)
```
mfrancisc, line 193: same here for checking all the config maps are not there anymore

controllers/useraccount/console_usersettings_test.go
```diff
@@ -0,0 +1,449 @@
...
+					Namespace: UserSettingNS,
+					Labels: map[string]string{
+						ConsoleUserSettingsIdentifier: "true",
+						ConsoleUserSettingsUID:        "johnsmith",
+					},
+				},
+			}
+			cl := fake.NewClientBuilder().WithObjects(cm1, cm2).Build()
+			err := deleteConfigMap(context.TODO(), cl, "johnsmith")
+			require.NoError(t, err)
```
mfrancisc, line 220: same here verifying all the cms were deleted

mfrancisc (commented):
Nice Job 👍 

I have only some minor comments/suggestions.

mfrancisc:
In addition to my comments, you could also add some `noise` objects and check that those are not deleted by the cleanup functions.

controllers/useraccount/console_usersettings_test.go
ranakan19: done in d4bc88c

controllers/useraccount/console_usersettings_test.go
ranakan19: done in d4bc88c

controllers/useraccount/console_usersettings_test.go
ranakan19: done in [d4bc88c](https://github.com/codeready-toolchain/member-operator/pull/568/commits/d4bc88cc583bfad7f84185b5fa341ffc508984e3)

controllers/useraccount/console_usersettings.go
ranakan19: thanks, simplified in [d4bc88c](https://github.com/codeready-toolchain/member-operator/pull/568/commits/d4bc88cc583bfad7f84185b5fa341ffc508984e3)

controllers/useraccount/console_usersettings.go
ranakan19: thanks, simplified in [d4bc88c](https://github.com/codeready-toolchain/member-operator/pull/568/commits/d4bc88cc583bfad7f84185b5fa341ffc508984e3)

controllers/useraccount/console_usersettings.go
ranakan19: thanks, simplified in [d4bc88c](https://github.com/codeready-toolchain/member-operator/pull/568/commits/d4bc88cc583bfad7f84185b5fa341ffc508984e3)

ranakan19:
> In addition to my comments, you could also add some `noise` objects and check that those are not deleted by the cleanup functions.



> In addition to my comments, you could also add some `noise` objects and check that those are not deleted by the cleanup functions.

Thanks, yeah added noise objects in the useraccount test in [d4bc88c](https://github.com/codeready-toolchain/member-operator/pull/568/commits/d4bc88cc583bfad7f84185b5fa341ffc508984e3)

ranakan19:
/retest 
to confirm the test is failing, there seem to be inconsistent time in reporting test failure and the test failing isn't related.

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,205 @@
...
+		return []v1.RoleBinding{}, err
+	}
+	return rbList.Items, nil
+}
+
+// function getConsoleUserSettingObjectByName looks for resources of type ConfigMap, Role and Rolebinding in the namespace openshift-console-user-settings by name.
+// it returns obj,nil when the object is found by name
+// return nil, nil if the object is not found, i.e ignore the not found error
+// return nil, err if there is any other kind of error, or if the object type is not supported
+func getConsoleUserSettingObjectByName(ctx context.Context, cl client.Client, name string, object client.Object) (client.Object, error) {
```
xcoulon, line 162: wouldn't it be simpler if we had 3 funcs, 1 per type of resource to find (cm, role, rolebinding), instead of this large `switch` block? For example: `getConsoleUserSettingConfigMap` + `getConsoleUserSettingRole` + `getConsoleUserSettingRoleBinding`?

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,205 @@
...
+			}
+		}
+	} else {
+		toDelete = append(toDelete, configMap)
+	}
+
+	if err := deleteResources(ctx, cl, toDelete); err != nil {
+		return err
+	}
+	return nil
```
MatousJobanek, line 57: as mentioned in the other comments, you don't need to duplicate the same code  - you can work with the generic `client.Object` or `unstructured.Unstructured` and  `unstructured.UnstructuredList` types. 
The pattern of the names as well as the labels seems to be the same for all resources, so you could theoretically have one generic function that would be called for all three types, just passing different GVK.
Please let me know if I'm missing some small detail.

MatousJobanek (commented):
Let's try to use generic types and avoid the code duplications

MatousJobanek:
@ranakan19 I just realized that we could significantly simplify it to a few lines of the code if you use the `Delete` and `DeleteAllOf` functions directly without trying to retrieve the objects upfront. Something like this should work, right?
```go
func deleteUserResources(ctx context.Context, cl client.Client, userUID string) error {
	if err := deleteResource(ctx, cl, userUID, &corev1.ConfigMap{}); err != nil {
		return err
	}
	if err := deleteResource(ctx, cl, userUID, &v1.Role{}); err != nil {
		return err
	}
	return deleteResource(ctx, cl, userUID, &v1.RoleBinding{})
}

func deleteResource(ctx context.Context, cl client.Client, userUID string, toDelete client.Object) error {

	name := ConsoleUserSettingsResourceNamePrefix + userUID
	toDelete.SetName(name)
	toDelete.SetNamespace(UserSettingNS)
	if err := cl.Delete(ctx, toDelete); err != nil {
		if errors.IsNotFound(err) {
			labels := map[string]string{ConsoleUserSettingsIdentifier: "true", ConsoleUserSettingsUID: userUID}
			return cl.DeleteAllOf(ctx, toDelete, client.MatchingLabels(labels), client.InNamespace(UserSettingNS))
		} else {
			return err
		}
	}
	return nil
}
```

ranakan19:
> @ranakan19 I just realized that we could significantly simplify it to a few lines of the code if you use the `Delete` and `DeleteAllOf` functions directly without trying to retrieve the objects upfront. Something like this should work, right?
> 
> ```go
> func deleteUserResources(ctx context.Context, cl client.Client, userUID string) error {
> 	if err := deleteResource(ctx, cl, userUID, &corev1.ConfigMap{}); err != nil {
> 		return err
> 	}
> 	if err := deleteResource(ctx, cl, userUID, &v1.Role{}); err != nil {
> 		return err
> 	}
> 	return deleteResource(ctx, cl, userUID, &v1.RoleBinding{})
> }
> 
> func deleteResource(ctx context.Context, cl client.Client, userUID string, toDelete client.Object) error {
> 
> 	name := ConsoleUserSettingsResourceNamePrefix + userUID
> 	toDelete.SetName(name)
> 	toDelete.SetNamespace(UserSettingNS)
> 	if err := cl.Delete(ctx, toDelete); err != nil {
> 		if errors.IsNotFound(err) {
> 			labels := map[string]string{ConsoleUserSettingsIdentifier: "true", ConsoleUserSettingsUID: userUID}
> 			return cl.DeleteAllOf(ctx, toDelete, client.MatchingLabels(labels), client.InNamespace(UserSettingNS))
> 		} else {
> 			return err
> 		}
> 	}
> 	return nil
> }
> ```

Thank you so much @MatousJobanek, this reduces complexity significantly.  done in [cefe5f3](https://github.com/codeready-toolchain/member-operator/pull/568/commits/cefe5f3f81dc73d4c681651eff2cc546c3e1bc18)

controllers/useraccount/console_usersettings_test.go
```diff
@@ -0,0 +1,285 @@
...
+	corev1 "k8s.io/api/core/v1"
+	rbac "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"testing"
+)
+
+func TestDeleteConsoleSettingObjects(t *testing.T) {
```
mfrancisc, line 17: minor - you could probably use _table driven tests_ for those , but feel free to ignore this or apply in follow up PR if you'd like to refactor the code

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,35 @@
...
+// Returns an error if the deletion operation fails, or nil if the operation is successful.
+func deleteResource(ctx context.Context, cl client.Client, userUID string, toDelete client.Object) error {
+
+	name := ConsoleUserSettingsResourceNamePrefix + userUID
+	toDelete.SetName(name)
+	toDelete.SetNamespace(UserSettingNS)
+	if err := cl.Delete(ctx, toDelete); err != nil {
+		if errors.IsNotFound(err) {
+			labels := map[string]string{ConsoleUserSettingsIdentifier: "true", ConsoleUserSettingsUID: userUID}
+			return cl.DeleteAllOf(ctx, toDelete, client.MatchingLabels(labels), client.InNamespace(UserSettingNS))
```
mfrancisc, line 29: just curious - I believe this would use the labels even if the name was set previously on `toDelete` object 🤔

mfrancisc (approved):
Nice Job 👍 
Looks good to me. I like the simplification you applied for `deleteResource`.

Thanks for the additional checks and tests!

controllers/useraccount/console_usersettings.go
MatousJobanek: from the comment of the method:

> // DeleteAllOf deletes all objects of the given type matching the given options.

so I guess that it shouldn't use the name nor namespace. 
But we could always create a deep-copy before setting the name and namespace.

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,35 @@
...
+//
+// userUID : The unique identifier of the user for whom the resource is being deleted.
+// Returns an error if the deletion operation fails, or nil if the operation is successful.
+func deleteResource(ctx context.Context, cl client.Client, userUID string, toDelete client.Object) error {
+
+	name := ConsoleUserSettingsResourceNamePrefix + userUID
+	toDelete.SetName(name)
+	toDelete.SetNamespace(UserSettingNS)
+	if err := cl.Delete(ctx, toDelete); err != nil {
+		if errors.IsNotFound(err) {
```
MatousJobanek, line 27: out of curiosity? is there any reason for deleting via a label only if the resource with the given name is not found? I mean, why couldn't we simply run both delete commands (by name and by a label) for every user that is being de-provisioned?

controllers/useraccount/console_usersettings_test.go
```diff
@@ -0,0 +1,285 @@
...
+				},
+			}
+			cl := fake.NewClientBuilder().WithObjects(cm).Build()
+
+			err := deleteResource(ctx, cl, "johnsmith", cm)
+			require.NoError(t, err)
+			// check that the configmap doesn't exist anymore
+			err = cl.Get(ctx, client.ObjectKey{Name: "user-settings-johnsmith", Namespace: UserSettingNS}, &corev1.ConfigMap{})
+			require.True(t, errors.IsNotFound(err))
+		})
```
MatousJobanek, line 35: Could you please add the missing `// given` `// when` `// then` comments to all tests? It's much easier to read them
```suggestion
		t.Run("Configmap found by name and deleted", func(t *testing.T) {
		        // given
			ctx := context.Background()
			cm := &corev1.ConfigMap{
				TypeMeta: metav1.TypeMeta{Kind: "ConfigMap"},
				ObjectMeta: metav1.ObjectMeta{
					Name:      "user-settings-johnsmith",
					Namespace: UserSettingNS,
				},
			}
			cl := fake.NewClientBuilder().WithObjects(cm).Build()
			
			// when
			err := deleteResource(ctx, cl, "johnsmith", cm)
			
			// then
			require.NoError(t, err)
			// check that the configmap doesn't exist anymore
			err = cl.Get(ctx, client.ObjectKey{Name: "user-settings-johnsmith", Namespace: UserSettingNS}, &corev1.ConfigMap{})
			require.True(t, errors.IsNotFound(err))
		})
```

controllers/useraccount/console_usersettings_test.go
```diff
@@ -0,0 +1,285 @@
...
+		t.Run("Configmap found by name and deleted", func(t *testing.T) {
+			ctx := context.Background()
+			cm := &corev1.ConfigMap{
+				TypeMeta: metav1.TypeMeta{Kind: "ConfigMap"},
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "user-settings-johnsmith",
+					Namespace: UserSettingNS,
+				},
+			}
+			cl := fake.NewClientBuilder().WithObjects(cm).Build()
```
MatousJobanek, line 28: is there any reason for not using our own FakeClient?

controllers/useraccount/console_usersettings_test.go
```diff
@@ -0,0 +1,285 @@
...
+		})
+		t.Run("No Error is returned when no configmap not found", func(t *testing.T) {
+			cl := test.NewFakeClient(t)
+			err := deleteResource(context.TODO(), cl, "johnsmith", &corev1.ConfigMap{})
+			require.NoError(t, err)
+		})
+
+	})
+
+	t.Run("Delete Role", func(t *testing.T) {
```
MatousJobanek, line 108: the `deleteResource` is generic now - it technically works with any type of the client.Object, so it doesn't matter if you test it with a ConfigMap, Role, or RoleBinding.
Let's pick one type, the rest looks like a duplication of the same. 
As it is written now, it's the responsibility of the caller (UserAccount controller) that should call the function with all three kinds/types.

controllers/useraccount/console_usersettings_test.go
```diff
@@ -0,0 +1,285 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "user-settings-name-no-match-second",
+					Namespace: UserSettingNS,
+					Labels: map[string]string{
+						ConsoleUserSettingsIdentifier: "true",
+						ConsoleUserSettingsUID:        "johnsmith",
+					},
+				},
+			}
+			cl := fake.NewClientBuilder().WithObjects(cm1, cm2).Build()
```
MatousJobanek, line 78: could you please add another object as a noise so we know that the label selection is applied correctly?

controllers/useraccount/useraccount_controller.go
```diff
@@ -484,6 +492,29 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
...
+	if err := deleteResource(ctx, r.Client, userUID, &corev1.ConfigMap{}); err != nil {
+		return err
+	}
+	if err := deleteResource(ctx, r.Client, userUID, &rbac.Role{}); err != nil {
+		return err
+	}
+	if err := deleteResource(ctx, r.Client, userUID, &rbac.RoleBinding{}); err != nil {
+		return err
+	}
+	return nil
```
MatousJobanek, line 515: just a detail
```suggestion
	return deleteResource(ctx, r.Client, userUID, &rbac.RoleBinding{})
```

controllers/useraccount/useraccount_controller_test.go
```diff
@@ -477,6 +478,159 @@ func TestReconcile(t *testing.T) {
...
+			assert.Equal(t, reconcile.Result{}, res)
+			require.NoError(t, err)
+
+			useraccount.AssertThatUserAccount(t, req.Name, cl).
+				HasConditions(notReady("Terminating", "deleting user/identity"))
+
+			// Check that the associated identity has been deleted
+			// when reconciling the useraccount with a deletion timestamp
+			assertIdentityNotFound(t, r, userAcc, config.Auth().Idp())
+			useraccount.AssertThatUserAccount(t, userAcc.Name, cl).HasConditions(terminating())
```
MatousJobanek, line 567: this seems to be a duplication of what is already tested in the previous test. Couldn't this be merged or refactored?

controllers/useraccount/useraccount_controller_test.go
```diff
@@ -477,6 +478,159 @@ func TestReconcile(t *testing.T) {
...
+				require.NoError(t, err)
+
+				useraccount.AssertThatUserAccount(t, req.Name, cl).
+					HasConditions(notReady("Terminating", "deleting user/identity"))
+
+				// Check that the associated configmap has been deleted
+				retrievedConfigMap := &corev1.ConfigMap{}
+				err := r.Client.Get(context.TODO(), types.NamespacedName{Name: resourceName, Namespace: UserSettingNS}, retrievedConfigMap)
+				require.Error(t, err)
+				assert.True(t, apierros.IsNotFound(err))
```
MatousJobanek, line 584: this (and the following lines) can be replaced with the usage of 
```go
AssertObjectNotFound(t, cl, UserSettingNS, resourceName, &corev1.ConfigMap{})
```

controllers/useraccount/useraccount_controller_test.go
```diff
@@ -477,6 +478,159 @@ func TestReconcile(t *testing.T) {
...
+				retrievedRb := &v1.RoleBinding{}
+				err = r.Client.Get(context.TODO(), types.NamespacedName{Name: resourceName, Namespace: UserSettingNS}, retrievedRb)
+				require.Error(t, err)
+				assert.True(t, apierros.IsNotFound(err))
+
+				// Check that the noise resources are not deleted
+				retrievedNoiseCm := &corev1.ConfigMap{}
+				err = r.Client.Get(context.TODO(), types.NamespacedName{Name: "noise-object", Namespace: UserSettingNS}, retrievedNoiseCm)
+				require.NoError(t, err)
+				assert.Equal(t, noiseCm, retrievedNoiseCm)
```
MatousJobanek, line 600: in a similar way, this can be replaced by:
```go
retrievedNoiseCm := &corev1.ConfigMap{}
AssertObject(t, cl, UserSettingNS, "noise-object", retrievedNoiseCm, func() {
	assert.Equal(t, noiseCm, retrievedNoiseCm)
})
```

controllers/useraccount/useraccount_controller_test.go
```diff
@@ -477,6 +478,159 @@ func TestReconcile(t *testing.T) {
...
+				// Check that the associated role has been deleted
+				retrievedRole := &v1.Role{}
+				err = r.Client.Get(context.TODO(), types.NamespacedName{Name: resourceName, Namespace: UserSettingNS}, retrievedRole)
+				require.Error(t, err)
+				assert.True(t, apierros.IsNotFound(err))
+				// Check that the associated rolebinding has been deleted
+				retrievedRb := &v1.RoleBinding{}
+				err = r.Client.Get(context.TODO(), types.NamespacedName{Name: resourceName, Namespace: UserSettingNS}, retrievedRb)
+				require.Error(t, err)
+				assert.True(t, apierros.IsNotFound(err))
```
MatousJobanek, line 594: and since the expected pattern of the names is the same, this can be replaced by
```go
for _, obj := range []client.Object{&corev1.ConfigMap{}, &v1.Role{}, &v1.RoleBinding{}} {
	AssertObjectNotFound(t, cl, UserSettingNS, resourceName, obj)
}
```

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,205 @@
...
+
+func getConfigMapByLabel(ctx context.Context, cl client.Client, userUID string) ([]corev1.ConfigMap, error) {
+	configMapList := &corev1.ConfigMapList{}
+	labels := map[string]string{ConsoleUserSettingsIdentifier: "true", ConsoleUserSettingsUID: userUID}
+	err := cl.List(ctx, configMapList, client.MatchingLabels(labels), client.InNamespace(UserSettingNS))
+	if err != nil {
+		return []corev1.ConfigMap{}, err
+	}
+	return configMapList.Items, nil
+}
```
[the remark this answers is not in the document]
ranakan19: resolving this as this function is no longer present

controllers/useraccount/console_usersettings.go
ranakan19: resolving this comment as this function is no longer present

controllers/useraccount/console_usersettings.go
ranakan19: resolving this comment, as this was taken care of with the other changes suggested and implemented in [cefe5f3](https://github.com/codeready-toolchain/member-operator/pull/568/commits/cefe5f3f81dc73d4c681651eff2cc546c3e1bc18)

controllers/useraccount/console_usersettings.go
ranakan19: In the ideal scenario it should be fine to just delete resources via label but since the label will only be present on newer objects ( once https://issues.redhat.com//browse/OCPBUGS-32321 gets implemented) - this function looks for the objects by name. 
The reason why it looks for deleting via label only after the resource name is not found by name is because it would be redundant trying to delete the object twice (since ideally there should not be more than one cm, role and rb for a user)

controllers/useraccount/console_usersettings_test.go
ranakan19: no, missed this one - fixing it now

controllers/useraccount/useraccount_controller_test.go
ranakan19: done in [be0c09f](https://github.com/codeready-toolchain/member-operator/pull/568/commits/be0c09fab987ae88941ce381ff2f841a94e594d7)

controllers/useraccount/useraccount_controller_test.go
ranakan19: done in [7b7bd25](https://github.com/codeready-toolchain/member-operator/pull/568/commits/7b7bd259bb5cb92d555a5043585a0d3c06875e8b)

controllers/useraccount/useraccount_controller_test.go
ranakan19: done in [7b7bd25](https://github.com/codeready-toolchain/member-operator/pull/568/commits/7b7bd259bb5cb92d555a5043585a0d3c06875e8b)

controllers/useraccount/useraccount_controller.go
ranakan19: done in [7b7bd25](https://github.com/codeready-toolchain/member-operator/pull/568/commits/7b7bd259bb5cb92d555a5043585a0d3c06875e8b)

controllers/useraccount/console_usersettings_test.go
ranakan19: fixed in [7b7bd25](https://github.com/codeready-toolchain/member-operator/pull/568/commits/7b7bd259bb5cb92d555a5043585a0d3c06875e8b)

controllers/useraccount/console_usersettings_test.go
ranakan19: done in [7b7bd25](https://github.com/codeready-toolchain/member-operator/pull/568/commits/7b7bd259bb5cb92d555a5043585a0d3c06875e8b)

controllers/useraccount/useraccount_controller_test.go
ranakan19: done in [d2b6226](https://github.com/codeready-toolchain/member-operator/pull/568/commits/d2b6226ae7d725105c82382ad27f43d0d7327e02)

controllers/useraccount/console_usersettings_test.go
ranakan19: done in [9fb025c](https://github.com/codeready-toolchain/member-operator/pull/568/commits/9fb025c2975f0a6a69b8c4936d252bb1c47be5cf)

controllers/useraccount/console_usersettings_test.go
ranakan19: done, removed test cases for Role and rolebinding and modified the remaining test case description with configmap to indiacte object

controllers/useraccount/console_usersettings_test.go
```diff
@@ -0,0 +1,144 @@
...
+			TypeMeta: metav1.TypeMeta{Kind: "ConfigMap"},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "user-settings-johnsmith",
+				Namespace: UserSettingNS,
+			},
+		}
+		cl := fake.NewClientBuilder().WithObjects(cm).Build()
+
+		// when
+		err := deleteResource(ctx, cl, "johnsmith", cm)
```
alexeykazakov, line 32: Don't pass the same `cm` here. Use `&corev1.ConfigMap{}` here instead. For proper testing.

controllers/useraccount/console_usersettings_test.go
```diff
@@ -0,0 +1,144 @@
...
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "user-settings-name-no-match-noise",
+				Namespace: UserSettingNS,
+				Labels: map[string]string{
+					ConsoleUserSettingsIdentifier: "true",
+				},
+			},
+		}
+		cl := test.NewFakeClient(t, cm, noiseObject)
```
alexeykazakov, line 67: Same here.

controllers/useraccount/console_usersettings_test.go
```diff
@@ -0,0 +1,144 @@
...
+				Labels: map[string]string{
+					ConsoleUserSettingsIdentifier: "true",
+					ConsoleUserSettingsUID:        "johnsmith",
+				},
+			},
+		}
+		cl := test.NewFakeClient(t, cm1, cm2)
+
+		// when
+		err := deleteResource(context.TODO(), cl, "johnsmith", &corev1.ConfigMap{})
```
alexeykazakov, line 109: And here.

controllers/useraccount/console_usersettings_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+		// then
+		require.Error(t, err)
+		require.Equal(t, "error in deleting configmap", err.Error())
+	})
+	t.Run("No Error is returned when no object is found", func(t *testing.T) {
+		// given
+		cl := test.NewFakeClient(t)
```
alexeykazakov, line 137: Add some noise here too and check it's still there.

alexeykazakov (approved):
Looks good! Just a few minor comments.

rsoaresd:
Hello, @ranakan19 🙂 

I was checking the latest runs of this PR and it seems that fail straight in `TestAfterMigration` (except for the runs that fail with `failed: failed to wait for importing imagestreamtags on ci-op-42k4p9q5/pipeline:ocp-4.12-upi-installer`).

I am not 100% sure that this PR causes it. I am still investigating. But from this [search query](https://search.dptools.openshift.org/?search=Unable+to+create+UserSignup+because+there+is+already+an+active+UserSignup+with+such+ID&maxAge=336h&context=1&type=build-log&name=.*codeready-toolchain.*&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job), we can see, for example, that error `Unable to create UserSignup because there is already an active UserSignup with such ID` only happened in this PR, for now.

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,35 @@
...
+	ConsoleUserSettingsUID                = "console.openshift.io/user-settings-uid"
+	UserSettingNS                         = "openshift-console-user-settings"
+	ConsoleUserSettingsResourceNamePrefix = "user-settings-"
+)
+
+// deleteResource deletes the specified resource associated with a user from console setting.
+// It first attempts to delete the resource by name, and if not found, it deletes all resources with matching labels.
+//
+// userUID : The unique identifier of the user for whom the resource is being deleted.
+// Returns an error if the deletion operation fails, or nil if the operation is successful.
```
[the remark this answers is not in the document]
ranakan19: done in [808948b](https://github.com/codeready-toolchain/member-operator/pull/568/commits/808948b9f9d8874182e4719eb287fe97f441bab5)

controllers/useraccount/useraccount_controller.go
```diff
@@ -484,6 +492,26 @@ func (r *Reconciler) deleteUser(ctx context.Context, userAcc *toolchainv1alpha1.
 	return true, nil
 }
 
+// deleteUserResources deletes the user settings resources (configmap, role and role-binding) associated with the specified user, created for a user by console.
```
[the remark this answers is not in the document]
ranakan19: done in [808948b](https://github.com/codeready-toolchain/member-operator/pull/568/commits/808948b9f9d8874182e4719eb287fe97f441bab5)

controllers/useraccount/console_usersettings_test.go
ranakan19: fixed in [808948b](https://github.com/codeready-toolchain/member-operator/pull/568/commits/808948b9f9d8874182e4719eb287fe97f441bab5)

controllers/useraccount/console_usersettings_test.go
ranakan19: I don't think this needs to be changed, since its adding the object to the fake client.

controllers/useraccount/console_usersettings_test.go
ranakan19: I don't think this needs to be changed, since its adding the object to the fake client.

controllers/useraccount/console_usersettings_test.go
ranakan19: done in 808948b

controllers/useraccount/console_usersettings_test.go
```diff
@@ -0,0 +1,155 @@
...
+		cl := fake.NewClientBuilder().WithObjects(cm).Build()
+
+		// when
+		err := deleteResource(ctx, cl, "johnsmith", &corev1.ConfigMap{})
+
+		// then
+		require.NoError(t, err)
+		// check that the configmap doesn't exist anymore
+		err = cl.Get(ctx, client.ObjectKey{Name: "user-settings-johnsmith", Namespace: UserSettingNS}, &corev1.ConfigMap{})
+		require.True(t, errors.IsNotFound(err))
```
MatousJobanek, line 39:
```suggestion
		AssertObjectNotFound(t, cl, UserSettingNS, "user-settings-johnsmith", &corev1.ConfigMap{})
```

controllers/useraccount/console_usersettings.go
```diff
@@ -0,0 +1,35 @@
...
+// Returns an error if the deletion operation fails. Returns nil if the operation is successful or there is nothing to delete.
+func deleteResource(ctx context.Context, cl client.Client, userUID string, toDelete client.Object) error {
+
+	name := ConsoleUserSettingsResourceNamePrefix + userUID
+	toDelete.SetName(name)
+	toDelete.SetNamespace(UserSettingNS)
+	if err := cl.Delete(ctx, toDelete); err != nil {
+		if errors.IsNotFound(err) {
+			labels := map[string]string{ConsoleUserSettingsIdentifier: "true", ConsoleUserSettingsUID: userUID}
+			return cl.DeleteAllOf(ctx, toDelete, client.MatchingLabels(labels), client.InNamespace(UserSettingNS))
```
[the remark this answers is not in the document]
MatousJobanek: taking back the comment about e2e tests - it doesn't seem to be easy to kick the creation of the resources unless we add some UI testing.

controllers/useraccount/console_usersettings_test.go
```diff
@@ -0,0 +1,155 @@
...
+		// given
+		ctx := context.Background()
+		cm := &corev1.ConfigMap{
+			TypeMeta: metav1.TypeMeta{Kind: "ConfigMap"},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "user-settings-johnsmith",
+				Namespace: UserSettingNS,
+			},
+		}
+		cl := fake.NewClientBuilder().WithObjects(cm).Build()
```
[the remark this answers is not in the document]
ranakan19: uh oh , thanks for catching that, removed in [45ad586](https://github.com/codeready-toolchain/member-operator/pull/568/commits/45ad586544a4cb59d77f4fe273b49369548ad134)

controllers/useraccount/console_usersettings_test.go
ranakan19: done in [45ad586](https://github.com/codeready-toolchain/member-operator/pull/568/commits/45ad586544a4cb59d77f4fe273b49369548ad134)

controllers/useraccount/console_usersettings.go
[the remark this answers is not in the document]
ranakan19: I talked to Vikram who is taking over console now, and he confirmed that this will worked upon - fixing the issue and adding the labels back - but yes there is no estimate as of now. I've asked in the ocp-forum channel to escalate this or atleast get an estimate

MatousJobanek (approved):
Looks good :+1: Thanks for the additional changes and for the simplification :rocket:

PR #580, opened 2024-06-25
Feature toggle annotations for NSTemplateSet

alexeykazakov:
It's the member operator part of https://issues.redhat.com/browse/SANDBOX-653

See the host operator part here: https://github.com/codeready-toolchain/host-operator/pull/1048

NSTemplateSet controller manages template objects with feature annotations according to the enabled features listed in the NSTemplate feature annotation.

See more details here: https://docs.google.com/document/d/1YWaR94JTrVBBVoMSjgIb8jb2FkqKiT1myVib9Qv0BzU/edit#bookmark=id.k3fvz630dsa6

**Note:** This PR introduces feature support of cluster scoped resources defined in the space templates only. Namespaced resources (for both space and space role templates) will be addressed later separately.

Paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/1008


rajivnathan (commented):
I took a broad look at the changes so far and it looks good overall. I wonder if we could accomplish the same thing if we alter the `shouldCreate()` func and make it a template filter function instead? The features act as a filter for the objects in the templates right? So maybe it's a good fit.

[tierTemplate.process()](https://github.com/codeready-toolchain/member-operator/blob/3525a39266b2d971348db2bbc100f08bfa1c8d89/controllers/nstemplateset/nstemplatetier.go#L87) is called to get the new objects in each of the cluster, namespaces and spacerole cases and in each call there are optional filters passed as params. eg. https://github.com/codeready-toolchain/member-operator/blob/3525a39266b2d971348db2bbc100f08bfa1c8d89/controllers/nstemplateset/cluster_resources.go#L133-L135

If we do it that way then the objects will be filtered at that point and we wouldn't need to alter any of the logic after that which should simplify the testing as well unless I'm missing something. WDYT?

alexeykazakov:
Interesting idea! I'm not sure it would cover all use cases but let me try.
I'm also not sure how it would simplify testing but maybe I'm missing something.

controllers/nstemplateset/namespaces.go
```diff
@@ -189,31 +191,50 @@ func (r *namespacesManager) ensureInnerNamespaceResources(ctx context.Context, n
...
-		if err != nil {
-			return r.wrapErrorWithStatusUpdate(ctx, nsTmplSet, r.setStatusUpdateFailed, err, "failed to process template for TierTemplate with name '%s'", currentRef)
+	// Check if we need to delete any obsolete resources
+	// First check if the namespace already has template reference meaning it's not freshly created and has not got any inner resources applied yet
+	currentRef, exists := namespace.Labels[toolchainv1alpha1.TemplateRefLabelKey]
+	if exists {
+		// Now check if there is any featured object to be deleted in case the feature annotation does not present in the NSTemplateSet anymore
+		toDeleteObsoleteFeaturedObjects := false
+		featureStr := nsTmplSet.Annotations[toolchainv1alpha1.FeatureToggleNameAnnotationKey]
+		features := strings.Split(featureStr, ",")
```
mfrancisc, line 201: should we check that the features annotations is not nil before splitting it ?

controllers/nstemplateset/namespaces_test.go
```diff
@@ -478,6 +480,89 @@ func TestEnsureNamespacesOK(t *testing.T) {
...
+				// Check that the rest of the resources for disabled features are not created
+				for _, feature := range allTierFeatures {
+					if slices.Contains(enabledFeatures, feature) {
+						nsAssertion.HasResource(fmt.Sprintf("%s-rb", feature), &rbacv1.RoleBinding{})
+					} else {
+						nsAssertion.HasNoResource(fmt.Sprintf("%s-rb", feature), &rbacv1.RoleBinding{})
+					}
+				}
+
+				AssertThatNamespace(t, spacename+"-stage", manager.Client).
```
mfrancisc, line 560: we could maybe add a comment here explaining why the stage namespace is not there. I guess it's because we provisioned only the `dev` namespace.

controllers/nstemplateset/namespaces_test.go
```diff
@@ -1024,10 +1120,14 @@ func TestUpdateNamespaces(t *testing.T) {
...
 			// given
-			nsTmplSet := newNSTmplSet(namespaceName, spacename, "advanced", withNamespaces("abcde12", "dev"))
+			nsTmplSet := newNSTmplSet(namespaceName, spacename, "advanced", withNamespaces("abcde12", "dev"), withNSTemplateSetFeatureAnnotation("feature-1"))
 			devNS := newNamespace("advanced", spacename, "dev", withTemplateRefUsingRevision("abcde11"))
 			ro := newRole(devNS.Name, "exec-pods", spacename)
 			rb := newRoleBinding(devNS.Name, "crtadmin-pods", spacename)
+			featuredRB := newRoleBinding(devNS.Name, "feature-1-rb", spacename) // existing featured RB we need to make sure gets deleted after updating since the new template doesn't have this featured object
+			featuredRB.Annotations = map[string]string{
+				toolchainv1alpha1.FeatureToggleNameAnnotationKey: "feature-1-rb",
+			}
```
mfrancisc, line 1130: shouldn't this be passed to the `prepareNamespacesManager` function or am I missing something ?

mfrancisc (commented):
Nice Job 👍 

I have only few minor comments

controllers/nstemplateset/namespaces.go
MatousJobanek, line 201: this should happen only if `currentRef != tierTemplate.templateRef`, right? Otherwise, it would be triggered everytime when there is some object with a feature not enabled for the given NSTemplateSet

controllers/nstemplateset/namespaces.go
```diff
@@ -189,31 +191,50 @@ func (r *namespacesManager) ensureInnerNamespaceResources(ctx context.Context, n
...
+		for _, obj := range newObjs {
+			objFeature, f := obj.GetAnnotations()[toolchainv1alpha1.FeatureToggleNameAnnotationKey]
+			if f && !slices.Contains(features, objFeature) {
+				// This object represents a disabled feature. We need to delete obsolete objects.
+				toDeleteObsoleteFeaturedObjects = true
+				break
+			}
```
MatousJobanek, line 208: related to the other comment I made before, you cannot only take into account the current/previous templateRef, but also the current & previous set of enabled features for the NSTemplateSet to know if there is something to delete or not

controllers/nstemplateset/nstemplateset_controller.go
```diff
@@ -221,17 +221,18 @@ func (r *Reconciler) deleteNSTemplateSet(ctx context.Context, nsTmplSet *toolcha
...
+func deleteObsoleteObjects(ctx context.Context, client runtimeclient.Client, currentObjs []runtimeclient.Object, newObjects []runtimeclient.Object, nsTmplSet *toolchainv1alpha1.NSTemplateSet) error {
 	logger := log.FromContext(ctx)
 	logger.Info("looking for obsolete objects", "count", len(currentObjs))
 Current:
 	for _, currentObj := range currentObjs {
 		objectLogger := logger.WithValues("objectName", currentObj.GetObjectKind().GroupVersionKind().Kind+"/"+currentObj.GetName())
 		objectLogger.Info("checking obsolete object", "object_namespace", currentObj.GetNamespace(), "object_name", currentObj.GetObjectKind().GroupVersionKind().Kind+"/"+currentObj.GetName())
 		for _, newObj := range newObjects {
-			if commonclient.SameGVKandName(currentObj, newObj) {
+			if shouldCreate(newObj, nsTmplSet) && commonclient.SameGVKandName(currentObj, newObj) {
```
MatousJobanek, line 235: the same thing again - we need to know which which feature toggles were enabled last time to make the proper delta between `currentObjs` and `newObjects`

MatousJobanek (commented):
some suggestions for an improvement and I figured out that it doesn't take into account the difference between feature toggles enabled before and after the update of the NSTemplateSet, so disabling a feature toggle won't get properly propagated to the namespace-scoped resources.

controllers/nstemplateset/namespaces.go
alexeykazakov: Not really. If the map is nil then `nsTmplSet.Annotations[toolchainv1alpha1.FeatureToggleNameAnnotationKey]` would return an empty string.

alexeykazakov: You can "read" a nil map. You can't "write" to a nil map.
```
m := nil
v := m["key"] // always returns ""
m["key"] = "value" // panics
```

controllers/nstemplateset/namespaces.go
```diff
@@ -352,7 +386,7 @@ func getNamespaceName(request reconcile.Request) (string, error) {
 
 // isUpToDateAndProvisioned checks if the obj has the correct Template Reference Label.
 // If so, it processes the tier template to get the expected roles and rolebindings and then checks if they are actually present in the namespace.
-func (r *namespacesManager) isUpToDateAndProvisioned(ctx context.Context, ns *corev1.Namespace, tierTemplate *tierTemplate) (bool, error) {
+func (r *namespacesManager) isUpToDateAndProvisioned(ctx context.Context, ns *corev1.Namespace, tierTemplate *tierTemplate, nsTmplSet *toolchainv1alpha1.NSTemplateSet) (bool, error) {
```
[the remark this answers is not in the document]
alexeykazakov: The current implementation for the namespaced feature-enabled resources is limited to RoleBindings and Roles only (as the only namespaced resources we are currently truly restoring if deleted). And unless I'm missing something the PR already covers the case of enabling features for existing NSTemplateSets.

What was missing is proper feature-disabling for existing NSTemplateSets. There was a bug in the `isUpToDateAndProvisioned()` function. I actually had a test for that specific use case but there was a bug in the test too so both bugs gave me a false negative result :)
This bug should be fixed here: https://github.com/codeready-toolchain/member-operator/pull/580/commits/3679b4ce0cdaedc8a5bb64369857ea5f0cae5eb0
I also extended the tests to cover both cases better.

Basically what the current code (in master) does: it always checks if the existing Roles &RBs match what is in the templates even if there is no changes in the reference (last applied templates). The current PR just adds an additional check if they also match the desired NSTemplate features. And unless I'm missing something there is no need for storing last applied features for that.

Alternative solution would be, as you said, storing the information about last applied features AND filtering out objects for disabled features in `tierTemplate.process()` (Rajiv's idea). That could potentially allows us to support all namespaced objects. Not just Roles and RBs.
I tried to avoid adding more annotations like "last-applied-features" and take advantage of our existing Role & RoleBinding handling. I hoped that a simple extension to the current Role & RB handling logic would be trivial but it turned out to be more complicated than expected. So I will give it a try and see if it's worth it.

controllers/nstemplateset/namespaces.go
alexeykazakov: If we do it in case the remplate ref was changed only then it wouldn't cover the cases of disabled/enabled feature for existing NSTemplateSets. 
So, you are correct that it triggers every time there is a feature object without the corresponding feature in the NSTemplateSet.

But it doesn't do too much. What it actually triggers:
- another template processing to get the list of objects defined in templates. It's already done multiple times (!) by this point. So we are doing it one more time.
- attempt to delete a feature object even if it doesn't actually exist. Should be fine because of the cache though. We already list all existing Roles and RBs. Every time. So it should not a big deal with the attempt deletion.

Is it perfect/optimal? No, it's not.
Does it make any significant performance impact? Unless I'm missing something, it should not make a noticeable difference.

controllers/nstemplateset/namespaces.go
alexeykazakov: See https://github.com/codeready-toolchain/member-operator/pull/580#discussion_r1667058340

controllers/nstemplateset/nstemplateset_controller.go
alexeykazakov: See https://github.com/codeready-toolchain/member-operator/pull/580#discussion_r1667058340

controllers/nstemplateset/feature_test.go
```diff
@@ -0,0 +1,80 @@
+package nstemplateset
```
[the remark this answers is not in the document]
alexeykazakov: Good catch. Renamed: https://github.com/codeready-toolchain/member-operator/pull/580/commits/0ec95d83c15c03ea738927660548f426a73e3e71

controllers/nstemplateset/feature_test.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"testing"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestShouldCreate(t *testing.T) {
+	t.Run("return namespace whose revision is not set", func(t *testing.T) {
```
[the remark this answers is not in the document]
alexeykazakov: Left overs from a copy-paste I forgot to cleanup :)
Thanks! https://github.com/codeready-toolchain/member-operator/pull/580/commits/8cdd546f331abfbd83042a7fe1b5a91cd300a17a

controllers/nstemplateset/cluster_resources_test.go
```diff
@@ -178,22 +180,110 @@ func TestEnsureClusterResourcesOK(t *testing.T) {
...
+			t.Run(testRun.name, func(t *testing.T) {
+				// given
 
-		// when
-		createdOrUpdated, err := manager.ensure(ctx, nsTmplSet)
+				// Create a NSTemplate referring to a tier with four ClusterResourceQuota objects
+				// The first one is with no feature annotation.
+				// The other three represent feature-1, feature-2 and feature-3.
+				allTierFeatures := []string{"feature-1", "feature-2", "feature-3"}
+				nsTmplSet = newNSTmplSet(namespaceName, spacename, "advanced", withNamespaces("abcde11", "dev"), withClusterResources("abcde11"))
```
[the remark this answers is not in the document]
alexeykazakov: No we don't! Thanks: https://github.com/codeready-toolchain/member-operator/pull/580/commits/2a2e040b36aefc46c875ff4b01636ebc946dcaad

controllers/nstemplateset/namespaces_test.go
alexeykazakov: Correct. We probably don't need this assertion at all because we don't care about the stage namespace in this test but anyway... I added the comment: https://github.com/codeready-toolchain/member-operator/pull/580/commits/4dbe286e0f90e4541e10878c21c7d94aba501db5

controllers/nstemplateset/namespaces_test.go
alexeykazakov: We should! Thank you: https://github.com/codeready-toolchain/member-operator/pull/580/commits/4dbe286e0f90e4541e10878c21c7d94aba501db5
Otherwise the test doesn't really test what it's supposed to.

controllers/nstemplateset/features.go
```diff
@@ -0,0 +1,30 @@
...
+// which means it's a regular object, and it's not managed by any feature toggle and should be always created.
+// Otherwise, returns false.
+func shouldCreate(toCreate runtimeclient.Object, nsTmplSet *toolchainv1alpha1.NSTemplateSet) bool {
+	feature, found := toCreate.GetAnnotations()[toolchainv1alpha1.FeatureToggleNameAnnotationKey]
+	if !found {
+		return true // This object is a regular object and not managed by a feature toggle. Always create it.
+	}
+	// This object represents a feature. Let's check if this feature is among winners in the NSTemplateSet.
+	winners, found := nsTmplSet.GetAnnotations()[toolchainv1alpha1.FeatureToggleNameAnnotationKey]
+	if !found {
```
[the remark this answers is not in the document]
alexeykazakov: I think `string.Split("", ",")` should really return an empty slice instead of confusing `[""]` :)
But anyway. Addressed here: https://github.com/codeready-toolchain/member-operator/pull/580/commits/48c03389ee23952d6d4ceb9ebc8b4f2dafda8693

controllers/nstemplateset/features.go
```diff
@@ -0,0 +1,30 @@
...
+		return false // No feature winners in the NSTemplateSet at all. Skip this object.
+	}
+	for _, winner := range strings.Split(winners, ",") {
+		if winner == feature {
+			return true
+		}
+	}
+	return false
```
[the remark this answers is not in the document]
alexeykazakov: Addressed here: https://github.com/codeready-toolchain/member-operator/pull/580/commits/48c03389ee23952d6d4ceb9ebc8b4f2dafda8693

controllers/nstemplateset/cluster_resources.go
```diff
@@ -269,6 +273,11 @@ NewObjects:
 				continue NewObjects
 			}
 		}
+		// Check if the new object is associated with a feature toggle.
+		// If yes then ignore this object if it represents a feature which is not enabled for this NSTemplateSet
+		if !shouldCreate(newObject, nsTmplSet) {
+			continue NewObjects
+		}
```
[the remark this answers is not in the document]
alexeykazakov: https://github.com/codeready-toolchain/member-operator/pull/580/commits/b75e2775c3ba57ea2f2fa04bf95f0c6015ffcbf6

controllers/nstemplateset/cluster_resources_test.go
```diff
@@ -178,22 +180,110 @@ func TestEnsureClusterResourcesOK(t *testing.T) {
...
+								}
+								// Check that the rest of the features are not (yet) created
+								for _, tierFeature := range allTierFeatures {
+									if !slices.Contains(expectedFeatureToBeAlreadyCreated, tierFeature) {
+										clusterAssertion.HasNoResource(fmt.Sprintf("%s-for-%s", tierFeature, spacename), &quotav1.ClusterResourceQuota{})
+									}
+								}
+							})
+						}
+					}
```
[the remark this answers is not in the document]
alexeykazakov: Good point.

controllers/nstemplateset/feature_test.go
```diff
@@ -0,0 +1,80 @@
...
+				objFeature:            "feature-2",
+				nsTemplateSetFeatures: "feature-1,feature-2,feature-3",
+				expectedToBeCreated:   true,
+			},
+		}
+		for _, testRun := range tests {
+			t.Run(testRun.name, func(t *testing.T) {
+				// given
+				obj := objectWithFeature(testRun.objFeature)
+				nsTmplSet := newNSTmplSet("na", "na", "na")
```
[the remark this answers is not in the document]
alexeykazakov: https://github.com/codeready-toolchain/member-operator/pull/580/commits/acc28dd5ee1dc6f836fc38680b15258bfff9864c

controllers/nstemplateset/namespaces.go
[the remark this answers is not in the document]
MatousJobanek: ok, so in case of namespace-scoped resources, you want to support feature-enabled resources for Roles/RoleBindings only for now, is that correct? And if the feature-enabled resources don't contain either Role or RoleBinding, then the controller will not apply the change for the already-existing NSTemplateSets.

I'm fine with that, let's just create a story to make sure that this is handled properly when we are doing the refactoring of the NSTemplateSet controller. cc @mfrancisc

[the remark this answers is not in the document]
MatousJobanek: This also means that the changes of the feature-toggles won't be reflected if the resources are defined in the SpaceRole templates. We update the SpaceRole templates only if there is some change between:
1. `toolchain.dev.openshift.com/last-applied-space-roles` annotation
2. current content of `NSTemplateSet.Spec.SpaceRoles`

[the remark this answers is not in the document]
MatousJobanek: That being said, since the support of the feature-enabled namespace-scoped resources is very limited in your PR and since we need to support only ClusterRoleBinding to be able to enable OLS in DevSandbox, don't you want to simplify the code and support cluster-scoped resources only for now?

We can take care of the broad/generic support of the feature-enabled resources when we will work on the changes of the NSTemplateSet controller. We will need to store the "last updated features" in status and re-apply everything if it's not matching with the current one in the annotation.

controllers/nstemplateset/namespaces.go
MatousJobanek: ok, now I understand why you did it - to make sure that we always delete all feature-disabled resources. 

No, it shouldn't make it much worse. Mainly when we know that we won't have any feature-related resources in DevSandbox templates any time soon. I hope that the next implementation of the controller will make the logic simpler and more effective.

controllers/nstemplateset/features_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+func TestReallySplit(t *testing.T) {
+	assert.Empty(t, splitCommaSeparatedList(""))
+	assert.Equal(t, strings.Split("1,2,3", ","), splitCommaSeparatedList("1,2,3"))
+	assert.Equal(t, strings.Split("1", ","), splitCommaSeparatedList("1"))
+}
+
+func p(s string) *string {
+	return &s
+}
```
MatousJobanek, line 109: you could use a generic version of this:
```suggestion
func p[T any](v T) *T {
	return &v
}
```
export it and move it somewhere to make it re-usable

controllers/nstemplateset/nstemplateset_controller.go
MatousJobanek, line 235: I know that the order doesn't matter, but it makes more sense to run `shouldCreate` only if the `currentObj` and `newObj` are the same
```suggestion
			if commonclient.SameGVKandName(currentObj, newObj) && shouldCreate(newObj, nsTmplSet) {
```

controllers/nstemplateset/space_roles.go
```diff
@@ -69,7 +69,7 @@ func (r *spaceRolesManager) ensure(ctx context.Context, nsTmplSet *toolchainv1al
 			return false, r.wrapErrorWithStatusUpdate(lctx, nsTmplSet, r.setStatusNamespaceProvisionFailed, err, "failed to provision namespace '%s' with space roles", ns.Name)
 		}
 
-		if err := deleteObsoleteObjects(lctx, r.Client, lastAppliedSpaceRoleObjs, spaceRoleObjs); err != nil {
+		if err := deleteObsoleteObjects(lctx, r.Client, lastAppliedSpaceRoleObjs, spaceRoleObjs, nsTmplSet); err != nil {
 			return false, r.wrapErrorWithStatusUpdate(lctx, nsTmplSet, r.setStatusUpdateFailed, err, "failed to delete redundant objects in namespace '%s'", ns.Name)
 		}
```
MatousJobanek, line 74: As I see, you don't support feature-enabled resources for SpaceRole objects at all - you don't filter them during creation. So, do you really still want to keep the support of the namespace-scoped resources when it's so limited?

MatousJobanek (approved):
Looks good, let's just consider if we really want to keep the limited support of the namespace-scoped resources, or if it wouldn't make more sense to drop it for now and address this in a generic way when we will work on the bigger changes of the controller.

controllers/nstemplateset/namespaces.go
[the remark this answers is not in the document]
alexeykazakov: I deleted all changes for namespaced resources. The PR supports the cluster scoped resource only now.

controllers/nstemplateset/namespaces.go
alexeykazakov: I deleted all changes for namespaced resources. The PR supports the cluster scoped resource only now.

controllers/nstemplateset/space_roles.go
alexeykazakov: I deleted all changes for namespaced resources. The PR supports the cluster scoped resource only now.

controllers/nstemplateset/nstemplateset_controller.go
alexeykazakov: I deleted all changes for namespaced resources. The PR supports the cluster scoped resource only now.

controllers/nstemplateset/features_test.go
alexeykazakov: I was thinking about moving it to common but it's just one-line function... Not sure if it's worth it?

MatousJobanek (approved):
Nice, thanks for simplifying the code :+1:

controllers/nstemplateset/features_test.go
MatousJobanek: Maybe later, it's fine. I just wanted to let you know about the generic version as well.

TBH, I would expect such a function to be provided by Go library itself.

mfrancisc (approved):
Great job 👍

PR #597, opened 2024-09-19
KUBESAW-187: Have a unique label for our operators

fbm3307:
Have a unique label instead of default controller-manger, so that it becomes easier to perform operations to only our operators
this is a predecessor for PR 
KSCTL - https://github.com/kubesaw/ksctl/pull/79 

Similar PR 
 Host-Operator - https://github.com/codeready-toolchain/host-operator/pull/1087

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -2,7 +2,8 @@ apiVersion: v1
 kind: Namespace
 metadata:
   labels:
-    control-plane: kubesaw-controller-manager
+    control-plane: controller-manager
+    kubesaw-control-plane: kubesaw-controller-manager
   name: system
 ---
 apiVersion: apps/v1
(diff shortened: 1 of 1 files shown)


config/manager/manager.yaml
```diff
@@ -11,18 +11,18 @@ metadata:
   name: controller-manager
   namespace: system
   labels:
-    control-plane: controller-manager
+    control-plane: kubesaw-controller-manager
```
MatousJobanek, line 14: TBH, I'm not sure if this labels is not used somewhere else (eg in sanbox-sre). Let's keep the label there and add a new one instead

fbm3307: yes you are right, added new ones now.

MatousJobanek (approved):
/lgtm

PR #614, opened 2025-01-16
Fix DeploymentConfig idling

rajivnathan:
fixes https://issues.redhat.com/browse/SANDBOX-942

updated e2e test: https://github.com/codeready-toolchain/toolchain-e2e/pull/1096

### DIFF
--- a/test/idler_assertion.go
+++ b/test/idler_assertion.go
@@ -197,6 +197,7 @@ func (a *IdleablePayloadAssertion) DeploymentConfigScaledDown(deployment *opensh
 	err := a.client.Get(context.TODO(), types.NamespacedName{Name: deployment.Name, Namespace: deployment.Namespace}, d)
 	require.NoError(a.t, err)
 	assert.Equal(a.t, int32(0), d.Spec.Replicas)
+	assert.False(a.t, d.Spec.Paused) // DeploymentConfig should be unpaused when scaling down so that the replicas update can be rolled out
 	return a
 }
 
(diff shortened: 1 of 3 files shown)


test/idler_assertion.go
```diff
@@ -197,6 +197,7 @@ func (a *IdleablePayloadAssertion) DeploymentConfigScaledDown(deployment *opensh
 	err := a.client.Get(context.TODO(), types.NamespacedName{Name: deployment.Name, Namespace: deployment.Namespace}, d)
 	require.NoError(a.t, err)
 	assert.Equal(a.t, int32(0), d.Spec.Replicas)
+	assert.False(a.t, d.Spec.Paused) // DeploymentConfig should be unpaused when scaling down so that the replicas update can be rolled out
```
alexeykazakov, line 200: Shouldn't we set it to true and then check it was reset to false? Otherwise the test would pass even if we don't reset it, right?

xcoulon (approved):

alexeykazakov (commented):
Thanks for fixing it!

test/idler_assertion.go
xcoulon, line 200: should it be set to `true` in the tests, before running the idler?

rajivnathan: sure, added 956344941b7a76c9a36581f3444a3fe94773a6f8

rajivnathan: sure, added 956344941b7a76c9a36581f3444a3fe94773a6f8

alexeykazakov (approved):
Thanks!

PR #621, opened 2025-02-05
remove logic of deletion of console setting resources using label

ranakan19:
Since labels were never added by console and ownerReferences have now been added back in 4.17 - we can delete the logic to delete console setting resources using labels.
Will keep the part of deleting console setting resources using name for a while before deleting.

slack thread for reference - https://redhat-internal.slack.com/archives/CHK0J6HT6/p1738683125581909


rajivnathan (approved):

alexeykazakov (approved):

MatousJobanek (approved):
/lgtm
thanks

PR #495, opened 2023-11-07
fix: forward context in idler reconciler

filariow:
Signed-off-by: Francesco Ilario <filario@redhat.com>

PR #630, opened 2025-02-20
SANDBOX-967: check restartCounts and idle crashlooping pods

ranakan19:
This PR
- introduces functionality to idle crashlooping pods if restartCount > 50
- change requeue time to 5 mins when nothing to idle.
- send idler notification when idles crashlooping pod ; need to make changes in the messaging of idler notification to not include 12 hrs - host PR - https://github.com/codeready-toolchain/host-operator/pull/1147

wrt - https://issues.redhat.com/browse/SANDBOX-967


controllers/idler/idler_controller.go
```diff
@@ -34,6 +34,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	RestartThreshold     = 50
+	RequeueTimeThreshold = 300
```
alexeykazakov, line 39:
```suggestion
	RequeueTimeThresholdSeconds = 300
```

controllers/idler/idler_controller.go
```diff
@@ -110,7 +115,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	// Find the earlier pod to kill and requeue. Otherwise, use the idler timeoutSeconds to requeue.
 	nextTime := nextPodToBeKilledAfter(logger, idler)
 	if nextTime == nil {
-		after := time.Duration(idler.Spec.TimeoutSeconds) * time.Second
+		after := time.Duration(RequeueTimeThreshold) * time.Second
```
alexeykazakov, line 118: But it would use the 300 second timeout only if nextTime == nil, right? So if it's set to 7 hours for example then we will wait 7h. I think we should use nextTime only if it's shorter than 300 seconds.

controllers/idler/idler_controller.go
```diff
@@ -198,6 +181,56 @@ func (r *Reconciler) ensureIdling(ctx context.Context, idler *toolchainv1alpha1.
...
+			} // not returning error to continue tracking remaining pods
+		}
+	}
+	return nil
+}
+
+func isRestartingOften(podstatus corev1.PodStatus) (bool, int32) {
+	var restartCount int32
+	for _, status := range podstatus.ContainerStatuses {
+		restartCount += status.RestartCount
```
alexeykazakov, line 229: So there can be multiple restart counters (one per container)? If it's the case then I guess we just need to pick the biggest number here. And not sum them up. Otherwise if there is a pod with 10 containers which all restart then just 3 pod restarts will cross the 30 restart threshold. Or I'm reading it wrong?

controllers/idler/idler_controller_test.go
```diff
@@ -143,8 +143,9 @@ func TestEnsureIdling(t *testing.T) {
 		reconciler, req, cl, allCl, dynamicClient := prepareReconcile(t, idler.Name, getHostCluster, idler, nsTmplSet, mur)
 
 		podsTooEarlyToKill := preparePayloads(t, reconciler, idler.Name, "", freshStartTimes(idler))
-
+		podCrashLooping := preparePayloadCrashloopingPod(t, reconciler, idler.Name, "restartCount-")
```
alexeykazakov, line 146: Let's also add pods with some restart count which is lower than the threshold to make sure they are not deleted. And add some crashlooping controller (in addition to the standalone pods). Like a Deployment. So we can cover more use cases.

alexeykazakov (commented):
Looks good overall!

controllers/idler/idler_controller.go
ranakan19: yes, you're reading that right. Let me make the change to look for the max restartCount if multiple containers are crashlooping.
I will also include this in one of the test cases.

controllers/idler/idler_controller.go
ranakan19: right. Okay, let me make that change.

ranakan19: done in [a9b2750](https://github.com/codeready-toolchain/member-operator/pull/630/commits/a9b27504aaa54ce3fd7aa66dad69abca3e0bf8f9)
To clarify - this would affect all workloads not just crashlooping ones

controllers/idler/idler_controller.go
ranakan19: done in [a9b2750](https://github.com/codeready-toolchain/member-operator/pull/630/commits/a9b27504aaa54ce3fd7aa66dad69abca3e0bf8f9)

controllers/idler/idler_controller.go
ranakan19: sorry, fixed the condition now in d40b75b

alexeykazakov: Yes, it will affect all the pods. So we are basically now requeueing every 5 minutes (or faster) for all pods. Which I think we have to do. Otherwise if the pod starts crash-loping, let's say in 10 minutes after the start we won't catch it because we will reconcile only in 12h.

controllers/idler/idler_controller.go
```diff
@@ -109,8 +114,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	}
 	// Find the earlier pod to kill and requeue. Otherwise, use the idler timeoutSeconds to requeue.
 	nextTime := nextPodToBeKilledAfter(logger, idler)
-	if nextTime == nil {
-		after := time.Duration(idler.Spec.TimeoutSeconds) * time.Second
+	after := time.Duration(RequeueTimeThresholdSeconds) * time.Second
+	if nextTime == nil || (nextTime != nil && *nextTime > after) {
```
alexeykazakov, line 118:
```suggestion
	if nextTime == nil || *nextTime > after {
```
You don't really need to check nextTime != nil in the second condition because if it's nil it won't get there. If the first condition is true (nextTie == nil) then it does't execute the second one (this is how `||` works).

controllers/idler/idler_controller.go
ranakan19: agreed, I know we discussed this, i just wanted to clarify in the PR for others reviewing as well.

controllers/idler/idler_controller.go
ranakan19: I had the same understanding. I wasn't checking for `!= nil` initially, but later fixed it in [d40b75b](https://github.com/codeready-toolchain/member-operator/pull/630/commits/d40b75bd609843b384e6dd0b6397ca9cda4f15ca) because I was getting segmentation  fault error when i was testing it in go playground. I don't have have the previous playground to look up what i wrote but I guess I was just doing something wrong before - https://go.dev/play/p/pPIV6OuIeDG
reverting the condition now. Thanks!

ranakan19: reverted in [c230065](https://github.com/codeready-toolchain/member-operator/pull/630/commits/c2300653882c00715be041bc59c395a5db195305)

controllers/idler/idler_controller_test.go
ranakan19: added in [576b44c](https://github.com/codeready-toolchain/member-operator/pull/630/commits/576b44c40a3fbd763c7792142a84d8a245b83b23)

controllers/idler/idler_controller.go
ranakan19: done in [576b44c](https://github.com/codeready-toolchain/member-operator/pull/630/commits/576b44c40a3fbd763c7792142a84d8a245b83b23)

controllers/idler/idler_controller_test.go
```diff
@@ -240,13 +245,14 @@ func TestEnsureIdling(t *testing.T) {
 					StatefulSetScaledDown(podsRunningForTooLong.statefulSet).
 					StatefulSetScaledUp(podsTooEarlyToKill.statefulSet).
 					StatefulSetScaledUp(noise.statefulSet).
+					StatefulSetScaledUp(podsCrashLoopingWithinThreshold.statefulSet).
 					VMStopped(podsRunningForTooLong.virtualmachine).
 					VMRunning(podsTooEarlyToKill.virtualmachine).
 					VMRunning(noise.virtualmachine)
 
-				// Still tracking all pods. Even deleted ones.
+				// Only tracks pods that have not been processed in this reconcile
```
alexeykazakov, line 253: Not been processed or rater not deleted yet?

controllers/idler/idler_controller_test.go
```diff
@@ -1324,32 +1330,91 @@ func preparePayloadsSinglePod(t *testing.T, r *Reconciler, namespace, namePrefix
...
+		ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("%s%s-statefulset", namePrefix, namespace), Namespace: namespace},
+		Spec:       appsv1.StatefulSetSpec{Replicas: &replicas},
+	}
+	err := r.AllNamespacesClient.Create(context.TODO(), sts)
+	require.NoError(t, err)
+	for i := 0; i < int(replicas); i++ {
+		pod := &corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("%s-pod-%d", sts.Name, i), Namespace: sts.Namespace},
+			Status: corev1.PodStatus{StartTime: &startTime, ContainerStatuses: []corev1.ContainerStatus{
+				{RestartCount: 30},
```
alexeykazakov, line 1389: maybe worth moving all replica count numbers to consts in case we want to change the threshold later. It will be easier to update the tests.

alexeykazakov (approved):
Looks good! Just a couple of minor comments.

controllers/idler/idler_controller_test.go
ranakan19: updated in [82f8118](https://github.com/codeready-toolchain/member-operator/pull/630/commits/82f8118a99f81e4c34d39486a26b46c419193af3),  have also added a comment in the third reconcile loop where we track controlledpods again

controllers/idler/idler_controller_test.go
ranakan19: done in [82f8118](https://github.com/codeready-toolchain/member-operator/pull/630/commits/82f8118a99f81e4c34d39486a26b46c419193af3)

controllers/idler/idler_controller.go
```diff
@@ -198,6 +184,58 @@ func (r *Reconciler) ensureIdling(ctx context.Context, idler *toolchainv1alpha1.
...
+		}
+		podLogger.Info("Pod deleted")
+	}
+	if appName == "" {
+		appName = pod.Name
+		appType = "Pod"
+	}
+
+	// If a build pod is in "PodCompleted" status then it was not running so there's no reason to send an idler notification
+	if podreason != "PodCompleted" || deletedByController || isCrashlooping {
```
rajivnathan, line 214: Do we really need the `isCrashlooping` boolean? There's no special treatment right?

controllers/idler/idler_controller_test.go
```diff
@@ -1317,12 +1327,94 @@ func preparePayloadsSinglePod(t *testing.T, r *Reconciler, namespace, namePrefix
 	}
 }
 
-func createPods(t *testing.T, r *Reconciler, owner metav1.Object, startTime metav1.Time, podsToTrack []*corev1.Pod, conditions ...corev1.PodCondition) []*corev1.Pod {
+func preparePayloadCrashloopingPod(t *testing.T, r *Reconciler, namespace, namePrefix string) payloads {
+	standalonePods := make([]*corev1.Pod, 0, 1)
+	startTime := metav1.Now()
+	replicas := int32(3)
+	// Create a standalone pod with no owner which has restart count sum > 50
```
rajivnathan, line 1340: Just an update to the comment since it's no longer the sum of restarts right?
```suggestion
	// Create a standalone pod with no owner which has at least one container with restart count > 50
```

controllers/idler/idler_controller.go
```diff
@@ -35,8 +35,11 @@ import (
 )
 
 const (
-	RestartThreshold            = 50
-	RequeueTimeThresholdSeconds = 300
+	RestartThreshold                      = 50
+	RequeueTimeThresholdSeconds           = 300
+	RestartCountWithinThresholdContainer1 = 30
+	RestartCountWithinThresholdContainer2 = 24
+	RestartCountOverThreshold             = 52
```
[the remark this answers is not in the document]
ranakan19: right, thanks, moved in [b38555a](https://github.com/codeready-toolchain/member-operator/pull/630/commits/b38555a279fa6b54b980b2ab98707159af6690b2)

controllers/idler/idler_controller_test.go
```diff
@@ -1317,12 +1333,94 @@ func preparePayloadsSinglePod(t *testing.T, r *Reconciler, namespace, namePrefix
 	}
 }
 
-func createPods(t *testing.T, r *Reconciler, owner metav1.Object, startTime metav1.Time, podsToTrack []*corev1.Pod, conditions ...corev1.PodCondition) []*corev1.Pod {
+func preparePayloadCrashloopingPod(t *testing.T, r *Reconciler, namespace, namePrefix string) payloads {
```
rajivnathan, line 1336:
```suggestion
func preparePayloadCrashloopingAboveThreshold(t *testing.T, r *Reconciler, namespace, namePrefix string) payloads {
```

rajivnathan (commented):
Looks good! Just some minor suggestions

controllers/idler/idler_controller.go
ranakan19: you're right, we can remove the boolean

controllers/idler/idler_controller.go
```diff
@@ -34,6 +34,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	RestartThreshold                      = 50
+	RequeueTimeThresholdSeconds           = 300
+	RestartCountWithinThresholdContainer1 = 30
+	RestartCountWithinThresholdContainer2 = 24
+	RestartCountOverThreshold             = 52
```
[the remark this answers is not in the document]
ranakan19: already done in [b38555a](https://github.com/codeready-toolchain/member-operator/pull/630/commits/b38555a279fa6b54b980b2ab98707159af6690b2)

controllers/idler/idler_controller_test.go
ranakan19: done in [a147e63](https://github.com/codeready-toolchain/member-operator/pull/630/commits/a147e636496395427a349188b8cd8a91bfc020ea)

controllers/idler/idler_controller_test.go
ranakan19: done in [a147e63](https://github.com/codeready-toolchain/member-operator/pull/630/commits/a147e636496395427a349188b8cd8a91bfc020ea)

controllers/idler/idler_controller.go
ranakan19: done in [a147e63](https://github.com/codeready-toolchain/member-operator/pull/630/commits/a147e636496395427a349188b8cd8a91bfc020ea)

controllers/idler/idler_controller.go
```diff
@@ -109,8 +114,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	}
 	// Find the earlier pod to kill and requeue. Otherwise, use the idler timeoutSeconds to requeue.
```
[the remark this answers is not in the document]
ranakan19: done in [a147e63](https://github.com/codeready-toolchain/member-operator/pull/630/commits/a147e636496395427a349188b8cd8a91bfc020ea)

controllers/idler/idler_controller.go
```diff
@@ -133,9 +138,19 @@ func (r *Reconciler) ensureIdling(ctx context.Context, idler *toolchainv1alpha1.
 	newStatusPods := make([]toolchainv1alpha1.Pod, 0, 10)
 	for _, pod := range podList.Items {
 		pod := pod // TODO We won't need it after upgrading to go 1.22: https://go.dev/blog/loopvar-preview
-		logger := log.FromContext(ctx)
-		podLogger := logger.WithValues("pod_name", pod.Name, "pod_phase", pod.Status.Phase)
+		podLogger := log.FromContext(ctx).WithValues("pod_name", pod.Name, "pod_phase", pod.Status.Phase)
 		if trackedPod := findPodByName(idler, pod.Name); trackedPod != nil {
+			// check the restart count for the trackedPod
+			crashlooping, restartCount := isRestartingOften(pod.Status)
+			if crashlooping {
```
MatousJobanek, line 145: I got really confused by reading two returned values that look like that they represent some different number. 
let's either rename the first var to `reachedThreshold` or change it to 
```go
restartCount := getHighestRestartCount(pod.Status)
if restartCount > RestartThreshold {

File: controllers/idler/idler_controller.go [hunk 15]
```diff
@@ -198,6 +181,58 @@ func (r *Reconciler) ensureIdling(ctx context.Context, idler *toolchainv1alpha1.
 	return r.updateStatusPods(ctx, idler, newStatusPods)
 }
 
+// Check if the pod belongs to a controller (Deployment, DeploymentConfig, etc) and scale it down to zero.
+// if it is a standalone pod, delete it.
+// Send notification if the deleted pod was managed by a controller, was a standalone pod that was not completed or was crashlooping
+func deletePodsAndCreateNotification(ctx context.Context, podLogger logr.Logger, pod corev1.Pod, r *Reconciler, idler *toolchainv1alpha1.Idler) error {
```
MatousJobanek, line 187: you should pass only the context, there is no need to provide the logger as a parameter of function

controllers/idler/idler_controller.go
```diff
@@ -198,6 +181,58 @@ func (r *Reconciler) ensureIdling(ctx context.Context, idler *toolchainv1alpha1.
...
+	lctx := log.IntoContext(ctx, podLogger)
```
MatousJobanek, line 188: This is needed only when you need to propagate some values from the logger to the context, but I don't see anything like that here.
If you want to propagate the values set in the logger before, then it should be done outside of this function

controllers/idler/idler_controller.go
```diff
@@ -198,6 +181,58 @@ func (r *Reconciler) ensureIdling(ctx context.Context, idler *toolchainv1alpha1.
...
+		if podCond.Type == "Ready" {
+			podreason = podCond.Reason
+		}
+	}
+	appType, appName, deletedByController, err := r.scaleControllerToZero(lctx, pod.ObjectMeta)
+	if err != nil {
+		return err
+	}
+	if !deletedByController { // Pod not managed by a controller. We can just delete the pod.
+		log.FromContext(ctx).Info("Deleting pod without controller")
```
MatousJobanek, line 202: no need to reinstantiate the logger

controllers/idler/idler_controller.go
```diff
@@ -198,6 +181,58 @@ func (r *Reconciler) ensureIdling(ctx context.Context, idler *toolchainv1alpha1.
 	return r.updateStatusPods(ctx, idler, newStatusPods)
 }
 
+// Check if the pod belongs to a controller (Deployment, DeploymentConfig, etc) and scale it down to zero.
+// if it is a standalone pod, delete it.
+// Send notification if the deleted pod was managed by a controller, was a standalone pod that was not completed or was crashlooping
+func deletePodsAndCreateNotification(ctx context.Context, podLogger logr.Logger, pod corev1.Pod, r *Reconciler, idler *toolchainv1alpha1.Idler) error {
+	lctx := log.IntoContext(ctx, podLogger)
+	logger := log.FromContext(ctx)
+	var podreason string
```
MatousJobanek, line 190:
```suggestion
	var podReason string
```

controllers/idler/idler_controller.go
ranakan19: lctx is used in the scaleControllerToero function. It was called outside this function before. I also see that this change was purposely made in https://github.com/codeready-toolchain/member-operator/pull/495.
I've now passed two contexts to the function deletePodsAndCreateNotification in sync with the changes of the mentioned PR. Let me know what you think

controllers/idler/idler_controller.go
ranakan19: done in [9164efe](https://github.com/codeready-toolchain/member-operator/pull/630/commits/9164efed1c4569ec9780def2f3ef716fc2da1abe)

controllers/idler/idler_controller.go
ranakan19: fixed in [9164efe](https://github.com/codeready-toolchain/member-operator/pull/630/commits/9164efed1c4569ec9780def2f3ef716fc2da1abe) - thanks for catching this

controllers/idler/idler_controller.go
ranakan19: two contexts passed in - [9164efe](https://github.com/codeready-toolchain/member-operator/pull/630/commits/9164efed1c4569ec9780def2f3ef716fc2da1abe)

controllers/idler/idler_controller.go
ranakan19: removed podLogger as parameter in [9164efe](https://github.com/codeready-toolchain/member-operator/pull/630/commits/9164efed1c4569ec9780def2f3ef716fc2da1abe)

controllers/idler/idler_controller.go
ranakan19: change done in [9164efe](https://github.com/codeready-toolchain/member-operator/pull/630/commits/9164efed1c4569ec9780def2f3ef716fc2da1abe)

controllers/idler/idler_controller_test.go
```diff
@@ -41,6 +41,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	RestartCountWithinThresholdContainer1 = 30
+	RestartCountWithinThresholdContainer2 = 24
+	RestartCountOverThreshold             = 52
+	TestIdlerTimeOutSeconds               = 540
```
ranakan19, line 48: I wanted to keep the `testIdlerTimeoutSeconds` to be different from `RequeueTimeThresholdSeconds`. Making it shorter was making the VMs idle too fast and assertions to fail at the count of pods being tracked. Keeping the `IdlerTimeoutSeconds` greater keeps the existing flow of tests with the max timeout of 5m.

controllers/idler/idler_controller.go
```diff
@@ -112,20 +112,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
-		logger.Info("requeueing for next pod to check", "after_seconds", after.Seconds())
-		return reconcile.Result{
-			Requeue:      true,
-			RequeueAfter: after,
-		}, r.setStatusReady(ctx, idler)
-	}
-	logger.Info("requeueing for next pod to kill", "after_seconds", nextTime.Seconds())
+	// Requeue in idler.Spec.TimeoutSeconds or RequeueTimeThresholdSeconds, whichever is sooner.
+	nextPodToKillAfter := nextPodToBeKilledAfter(logger, idler)
+	maxRequeDuration := time.Duration(RequeueTimeThresholdSeconds) * time.Second
```
MatousJobanek, line 117: typo
```suggestion
	maxRequeueDuration := time.Duration(RequeueTimeThresholdSeconds) * time.Second
```

controllers/idler/idler_controller.go
```diff
@@ -112,20 +112,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+	idlerTimeoutDuration := time.Duration(idler.Spec.TimeoutSeconds) * time.Second
+	after := findShortestDuration(nextPodToKillAfter, &maxRequeDuration, &idlerTimeoutDuration)
```
MatousJobanek, line 119: I like the idea of the `findShortestDuration` function, but wouldn't make more sense to move 
```
nextPodToKillAfter := nextPodToBeKilledAfter(logger, idler)
maxRequeDuration := time.Duration(RequeueTimeThresholdSeconds) * time.Second
idlerTimeoutDuration := time.Duration(idler.Spec.TimeoutSeconds) * time.Second
```
into `findShortestDuration` function? the variable don't seem to be used anywhere else in this method...

controllers/idler/idler_controller.go
```diff
@@ -635,6 +631,19 @@ func nextPodToBeKilledAfter(log logr.Logger, idler *toolchainv1alpha1.Idler) *ti
 	return &d
 }
 
+// findShortestDuration finds the shortest duration the given durations
+func findShortestDuration(durations ...*time.Duration) *time.Duration {
```
MatousJobanek, line 635: we expect that the function always returns some non-nil value, so let's return it as  non-pointer value, there is no need to keep the pointer in there.

controllers/idler/idler_controller.go
```diff
@@ -133,9 +134,20 @@ func (r *Reconciler) ensureIdling(ctx context.Context, idler *toolchainv1alpha1.
...
-		podLogger := logger.WithValues("pod_name", pod.Name, "pod_phase", pod.Status.Phase)
+		podLogger := log.FromContext(ctx).WithValues("pod_name", pod.Name, "pod_phase", pod.Status.Phase)
 		if trackedPod := findPodByName(idler, pod.Name); trackedPod != nil {
+			// check the restart count for the trackedPod
+			restartCount := getHighestRestartCount(pod.Status)
+			if restartCount > RestartThreshold {
+				podLogger.Info("Pod is restarting too often. Killing the pod", "restart_count", restartCount)
+				lctx := log.IntoContext(ctx, podLogger)
+				// Check if it belongs to a controller (Deployment, DeploymentConfig, etc) and scale it down to zero.
+				err := deletePodsAndCreateNotification(ctx, lctx, pod, r, idler)
```
MatousJobanek, line 145: could you please clarify what is the point of using two types of context? Why not using one?
Also, I would rename the context to make it clear that it's related to the pod
```suggestion
				podCtx := log.IntoContext(ctx, podLogger)
				// Check if it belongs to a controller (Deployment, DeploymentConfig, etc) and scale it down to zero.
				err := deletePodsAndCreateNotification(podCtx, pod, r, idler)
```

controllers/idler/idler_controller.go
```diff
@@ -144,43 +156,12 @@ func (r *Reconciler) ensureIdling(ctx context.Context, idler *toolchainv1alpha1.
...
-				var podreason string
-				podCondition := pod.Status.Conditions
-				for _, podCond := range podCondition {
-					if podCond.Type == "Ready" {
-						podreason = podCond.Reason
-					}
-				}
-
-				// Check if it belongs to a controller (Deployment, DeploymentConfig, etc) and scale it down to zero.
 				lctx := log.IntoContext(ctx, podLogger)
```
MatousJobanek, line 159: you can define the pod context only once - when you defined the pod logger

controllers/idler/idler_controller.go
```diff
@@ -144,43 +156,12 @@ func (r *Reconciler) ensureIdling(ctx context.Context, idler *toolchainv1alpha1.
...
-					if err := r.createNotification(ctx, idler, appName, appType); err != nil {
-						logger.Error(err, "failed to create Notification")
-						if err = r.setStatusIdlerNotificationCreationFailed(ctx, idler, err.Error()); err != nil {
-							logger.Error(err, "failed to set status IdlerNotificationCreationFailed")
-						} // not returning error to continue tracking remaining pods
-					}
-				}
 
 			} else {
 				newStatusPods = append(newStatusPods, *trackedPod) // keep tracking
```
MatousJobanek, line 167: only a small detail, another option (to keep it consistent with the previous section) would be 
```go
				continue
			}
			newStatusPods = append(newStatusPods, *trackedPod) // keep tracking
```
but feel free to keep it as it is

alexeykazakov (commented):
Looks good to me. Besides Matou's comments :)

controllers/idler/idler_controller.go
ranakan19: I tried answering it here - https://github.com/codeready-toolchain/member-operator/pull/630#discussion_r1970555406
To add though, podCtx would have the pod_name and pod_phase which is useful to have in the logs when scalling the controller and pods down. instead of passing it in the logs everytime. 
Below is the snippet of logs with normal ctx and podCtx. I'd keep the pod context - its helpful to see which pod is referenced in the log itself.
i'll rename the variable to podCtx though to make it clearer - thanks!
```
2025-02-26T13:52:38-05:00	INFO	Scaling controller to zero	{"pod_name": "restartCount-alex-stage-pod-fail", "pod_phase": "", "name": "restartCount-alex-stage-pod-fail"}
2025-02-26T13:52:38-05:00	INFO	Deleting pod without controller
2025-02-26T13:52:38-05:00	INFO	Pod deleted
```

ranakan19: we could just use podCtx for the entire function, but it would be an overkill imo - for example: pod_name and pod_phase is not needed in the logs when creating a notification, it'll be good to have sure but i think we can get by without it too.

controllers/idler/idler_controller.go
ranakan19: fixed in [bfc3cda](https://github.com/codeready-toolchain/member-operator/pull/630/commits/bfc3cdab10e2b06e72dbf287053a7f7188fad488)

controllers/idler/idler_controller.go
ranakan19: makes sense, moved it into the function and renamed it to `findShortestRequeueDuration` in [bfc3cda](https://github.com/codeready-toolchain/member-operator/pull/630/commits/bfc3cdab10e2b06e72dbf287053a7f7188fad488)
also removed the log.Info in `nextPodToKillAfter` - thus only logs the shortest requeue duration in `findShortestRequeueDuration`

controllers/idler/idler_controller.go
ranakan19: done in [bfc3cda](https://github.com/codeready-toolchain/member-operator/pull/630/commits/bfc3cdab10e2b06e72dbf287053a7f7188fad488)

controllers/idler/idler_controller.go
ranakan19: done in [bfc3cda](https://github.com/codeready-toolchain/member-operator/pull/630/commits/bfc3cdab10e2b06e72dbf287053a7f7188fad488)

controllers/idler/idler_controller.go
ranakan19: I've kept it as it is for now - making the change had some test failures in the number of pods being tracked - i didn't debug further for now

controllers/idler/idler_controller.go
```diff
@@ -34,6 +34,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+const (
+	RestartThreshold            = 50
+	RequeueTimeThresholdSeconds = 300
```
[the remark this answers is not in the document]
ranakan19: done in [bfc3cda](https://github.com/codeready-toolchain/member-operator/pull/630/commits/bfc3cdab10e2b06e72dbf287053a7f7188fad488)

controllers/idler/idler_controller.go
MatousJobanek: I'm completely fine with using `podCtx`, which is also what I proposed, but I don't see any reason for passing and using the original `ctx` context - this is what I'm still missing in your explanation. 

> we could just use podCtx for the entire function, but it would be an overkill imo - for example: pod_name and pod_phase is not needed in the logs when creating a notification, it'll be good to have sure but i think we can get by without it too.

overkill is using two types of contexts if not really needed for any cancellation reasons or anything like that. 
This whole function `deletePodsAndCreateNotification` is executed in the context of the specific pod, so it's completely valid (and TBH also expected) to include the pod information in the logs. From your snippet, you can also see that the lines:
```
2025-02-26T13:52:38-05:00	INFO	Deleting pod without controller
2025-02-26T13:52:38-05:00	INFO	Pod deleted
```
are missing which pod it is referring to, which is more a bug than a feature. 
Providing the context values is the best practice we do everywhere else, just look at the controller context and the logger - it contains the request metadata of the object the reconcile was triggered for in the whole reconcile loop.

MatousJobanek (commented):
Thanks for addressing my comments. Looks good :+1:

PR #631, opened 2025-02-28
Ansible Automation Platform Idler

alexeykazakov:
It turned out that we can't rely on the "Running" condition in the AAP CRs. It doesn't really properly reflect the status of AAPs. So we can't just use a simple cron job to idle AAPs.

This PR extends our idler so it can now handle AAPs too.
- The idling timeout for AAPs is shorten (comparing to the regular timeout from the Idler.Spec) so the regular Pod idler does not interfere with the AAP idler. AAP Idler should kick in first.
- I also tried to separate the AAP idler code from the main pod idler code.

TODO:
- [x] tests


controllers/idler/app_idler.go
```diff
@@ -0,0 +1,218 @@
+package idler
```
metlos, line 1: IMHO, the file should be called `aap_idler.go`. Typo?

alexeykazakov: Yes! :)

controllers/idler/idler_controller.go
```diff
@@ -111,6 +114,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(ctx, idler, r.setStatusFailed, err,
 			"failed to ensure idling '%s'", idler.Name)
 	}
+	if err := r.ensureAnsiblePlatformIdling(ctx, idler); err != nil {
```
rajivnathan, line 117: Couldn't we also call the AAP idler logic from the scaleControllerToZero function only for pods that have the AAP owner reference like how we do for VMs? https://github.com/codeready-toolchain/member-operator/blob/fc363451b7190fd597f62cb968f9d6bf74be5a35/controllers/idler/idler_controller.go#L321

controllers/idler/app_idler.go
```diff
@@ -0,0 +1,218 @@
...
+var aapGVR = schema.GroupVersionResource{Group: "aap.ansible.com", Version: "v1alpha1", Resource: "ansibleautomationplatforms"}
+
+// ensureAnsiblePlatformIdling checks if there is any long-running pod belonging to an AAP resource and if yes, then it idles the AAP
+// and sends a notification to the user.
+func (r *Reconciler) ensureAnsiblePlatformIdling(ctx context.Context, idler *toolchainv1alpha1.Idler) error {
+
+	idledAAPs := []string{}
+
+	// Check if there is any AAP CRs in the namespace
+	aapList, err := r.DynamicClient.Resource(aapGVR).Namespace(idler.Name).List(ctx, metav1.ListOptions{})
```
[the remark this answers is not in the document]
alexeykazakov: There are two reasons to list aaps:
1. A working AAP instance has a lot of pods (like 20 or so). Which have a complex ownership structure (for example some of them look like this: AAP -> Another AAP CR -> yet another AAP CR -> Deployment -> ReplicaSet -> Pod). And in most cases we expect users to have just one (or none) AAP (vs. many of them). Our UI will work with a single AAP instance only (it will create, show, un-idle just one AAP with a hardcoded name).
So, in case there is just one AAP then in order to avoid going through all of this ownership chain for every pod (20 times) for a single AAP instance we just list all AAPs first and then if the first AAP is already idled (after going through the ownership chain for the fist pod) then we don't need to handle the rest of the pods (19 AAP pods and any other one not related to AAP the user might have).
In other words in most cases we will need to handle a single pod instead of 20 of them.
2. Since we want to list all AAPs for the reason above anyway, why not do not skip checking the pods all together if there is no AAP in the namespace?

It's basically all about optimization. And a pod watcher doesn't change it, does it?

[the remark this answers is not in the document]
alexeykazakov: I'm not sure tbh. Most likely it will return an error (unavailable API). I will need to check it. If it's an error then we need to ignore it. Thanks for raising it.

controllers/idler/app_idler.go
alexeykazakov: Renamed in https://github.com/codeready-toolchain/member-operator/pull/631/commits/20fa50f5aaf38e97a80ee572071d3c219bd10914

controllers/idler/app_idler.go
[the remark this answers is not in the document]
alexeykazakov: Should be addressed here now: https://github.com/codeready-toolchain/member-operator/pull/631/commits/b82908248f62706183c39b123f6fefa30f859d1b
It now checks the AAP API availability before checking the AAP instances.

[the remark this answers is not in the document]
rajivnathan: That makes sense, thanks for the explanation. I think that info would be good to capture in a comment as well

controllers/idler/aap_idler.go
```diff
@@ -0,0 +1,252 @@
...
+
+const oneHour = 60 * 60 // in seconds
+
+func (r *Reconciler) aapTimeoutSeconds(idler *toolchainv1alpha1.Idler) int32 {
+	// Check if the idler timeout is less than one hour and if so, set it to half of the timeout.
+	// Otherwise, subtract one hour from the timeout.
+	// This is to ensure that the AAP idler kicks in before the main idler.
+	timeoutSeconds := idler.Spec.TimeoutSeconds
+	if timeoutSeconds <= oneHour {
+		timeoutSeconds = timeoutSeconds / 2
```
mfrancisc, line 146: unless I'm wrong, this would not work as of now. The AAP provisioning takes 30mins best case scenario, so this would just keep idling the instance , in case the timeout is less than one hour I think we should log an error . 

Alternatively , we could just have the timeout hardcoded in the idler for the time being.

controllers/idler/idler_controller.go
mfrancisc, line 117: shouldn't we move this before the `ensureIdling` ? 

In this way we first check if there's any pod associated with AAP and only after that we idle remaining pods.

controllers/idler/aap_idler.go
```diff
@@ -0,0 +1,252 @@
...
+	} else {
+		timeoutSeconds = timeoutSeconds - oneHour
+	}
+	return timeoutSeconds
+}
+
+// ensureAAPIdled checks if the long-running or crash-looping pod belongs to an AAP instance and if so, ensures that the AAP is idled.
+// Returns the AAP resource name in case it was idled, or an empty string if it was not idled.
+func (r *Reconciler) ensureAAPIdled(ctx context.Context, pod corev1.Pod, aapAPI aapAPI) (string, error) {
+	podCondition := pod.Status.Conditions
```
mfrancisc, line 156: we should probably check that the Status and the Conditions are not nil

controllers/idler/aap_idler.go
```diff
@@ -0,0 +1,252 @@
...
+	if err != nil {
+		return nil, err
+	}
+	// Get the owner object
+	ownerObject, err := r.DynamicClient.Resource(*gvr).Namespace(obj.GetNamespace()).Get(ctx, owner.Name, metav1.GetOptions{})
+	if errors.IsNotFound(err) { // Ignore not found errors. Can happen if the parent controller has been deleted. The Garbage Collector should delete the pods shortly.
+		logger.Info("Owner not found", "kind", owner.Kind, "name", owner.Name)
+		return nil, nil
+	}
+	if ownerObject.GetKind() == "AnsibleAutomationPlatform" {
```
mfrancisc, line 206: we could  reuse  the constant here
```suggestion
	if ownerObject.GetKind() == aapKind {
```

controllers/idler/aap_idler.go
```diff
@@ -0,0 +1,252 @@
...
+			// The AAP has been Idled
+			// A notification should be sent
+			if err := r.createNotification(podCtx, idler, aapName, "Ansible Automation Platform"); err != nil {
+				podLogger.Error(err, "failed to create Notification")
+				if err = r.setStatusIdlerNotificationCreationFailed(podCtx, idler, err.Error()); err != nil {
+					podLogger.Error(err, "failed to set status IdlerNotificationCreationFailed")
+				} // not returning error to continue tracking remaining pods
+			}
+
+			idledAAPs = append(idledAAPs, aapName)
```
mfrancisc, line 108: If I'm seeing this correctly, we potentially keep appending the same aapName multiple times - all pods will have same aapName.

controllers/idler/aap_idler.go
```diff
@@ -0,0 +1,252 @@
...
+				return nil
+			}
+		}
+	}
+
+	return nil
+}
+
+func checkIfAllAAPsIdled(allAAPs *unstructured.UnstructuredList, idledAAPs []string) bool {
+	if len(allAAPs.Items) != len(idledAAPs) {
```
mfrancisc, line 120: I think the len will be different since we append the same aapName multiple times, see previous comment. We could maybe use a map for this, and this would also simplify the loops below.

mfrancisc (commented):
Looks good overall!

controllers/idler/idler_controller.go
MatousJobanek: are we sure that there is really no way (label, annotation, some spec field) we could decide if the pod is created/owned by the APP instance?

I'm wondering if it wouldn't make sense to give it a try and incorporate it into the main logic - something like this:
* if controller is not owned by someone else then scale it down 
* if controller is owned by someone else, then:
    * if it's app, then go in the chain up as you do using `getAAPOwner` function
    * if it's owned by an unknown instance, then try to scale it down (expecting that the owner reverts the change)

in other words, incorporate the knowledge and usage of `getAAPOwner` into `scaleControllerToZero` or any of the functions called from it. it shouldn't be so hard and it would be very similar to what you do in the `aap_idler.go`.

controllers/idler/app_idler.go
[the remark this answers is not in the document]
alexeykazakov: Yes, the main idler logic and the aap idler logic kinda compete with each other. This is why I shorten the AAP idling timeout so it kicks in one hour earlier than the main idler: https://github.com/codeready-toolchain/member-operator/pull/631/files#diff-79b76c6fb335e6e635f582d5c79bbd4cadbb780830fa6af6e2e936e78b83253eR140-R143
>	// Check if the idler timeout is less than one hour and if so, set it to half of the timeout.
	// Otherwise, subtract one hour from the timeout.
	// This is to ensure that the AAP idler kicks in before the main idler.

So in theory the main idler should never try to idle controllers which are managed by the aap idler because the aap idler would idle it one hour earlier.

controllers/idler/aap_idler.go
alexeykazakov: Our production uses 12h as the timeout. This timeout /2 thing is just for unit tests, really.

controllers/idler/idler_controller.go
alexeykazakov: I was thinking about it but too but it doesn't really matter, does it? We can't rely on the order inside the reconcile loop anyway. Even if the ensureAAPIdling is executed first and does idle the AAP CRs, the pods actually would be deleted with a noticeable delay, so ensureIdling will be executed and finished before the pods are deleted by the AAP operator. So it wouldn't solve the problem with two idler functions (the aap and the main one) trying to idle the same aap pods. Instead it relies on the shorten AAP idling timeout. See https://github.com/codeready-toolchain/member-operator/pull/631#discussion_r1980090798 for details.

controllers/idler/aap_idler.go
alexeykazakov: Status can't be nil. It's not a pointer. And range can handle nil. So no need to check for nil here. However I got rid of the podCondition var since we don't really need it: https://github.com/codeready-toolchain/member-operator/pull/631/commits/b1f628febc20e71fabc22ad7495a6541d8f36d74

controllers/idler/aap_idler.go
alexeykazakov: Yep. You are right. While it still should work but it's better to use a map. Will fix it. Thanks.

controllers/idler/aap_idler.go
alexeykazakov: Great catch! Will fix it.

controllers/idler/idler_controller.go
mfrancisc: Correct the logic shouldn't rely on the order we execute the functions, but still I was thinking it might help with processing less pods in `ensureIdling`. The pods won't be deleted immediately right, but there are chances some of them might be gone by the time we start the `ensureIdling` , or at least they should be in `Terminating` state thus the function should ignore them. 

So basically this would be just a performance improvement.

controllers/idler/aap_idler.go
mfrancisc: Makes sense! Thanks!

controllers/idler/idler_controller.go
alexeykazakov: My experience tells me that it takes at least seconds to get the pod in the terminating state. So it doesn't look like there is a good chance that the pods will be gone by the time of the ensureIdling execution. Also there is actually another reason for keeping ensureIdling first. We still need the pods to be tracked in the Idler.Status which is done ensureIdling.

controllers/idler/aap_idler.go
alexeykazakov: Fixed in https://github.com/codeready-toolchain/member-operator/pull/631/commits/7c6e0e94ec34da50b737268382f9b3cc2c72b3a9

controllers/idler/aap_idler.go
alexeykazakov: Fixed in https://github.com/codeready-toolchain/member-operator/pull/631/commits/7c6e0e94ec34da50b737268382f9b3cc2c72b3a9

controllers/idler/aap_idler.go
```diff
@@ -0,0 +1,231 @@
...
+	resourceLists, err := r.DiscoveryClient.ServerPreferredResources()
+	if err != nil {
+		return err
+	}
+
+	// Check if the AAP API is even available/installed
+	aapGVR, err := findGVRForKind(aapKind, aapAPIVersion, resourceLists)
+	if err != nil {
+		return err
+	}
```
MatousJobanek, line 44: this could be done only once  - either as part of the controller initialization or using `sync.Once` logic

controllers/idler/aap_idler.go
```diff
@@ -0,0 +1,231 @@
...
+	// Get the owner object
+	ownerObject, err := r.DynamicClient.Resource(*gvr).Namespace(obj.GetNamespace()).Get(ctx, owner.Name, metav1.GetOptions{})
+	if errors.IsNotFound(err) { // Ignore not found errors. Can happen if the parent controller has been deleted. The Garbage Collector should delete the pods shortly.
+		logger.Info("Owner not found", "kind", owner.Kind, "name", owner.Name)
+		return nil, nil
+	}
+	if ownerObject.GetKind() == aapKind {
+		// Found the top AAP owner. Return it.
+		return obj, nil
```
MatousJobanek, line 187: shouldn't it return the `ownerObject` instead?
```suggestion
		// Found the top AAP owner. Return it.
		return ownerObject, nil
```

controllers/idler/aap_idler.go
alexeykazakov: I was thinking about that but then if the API is changed (installed/uninstalled operators) then we will have to restart the member operator manually. The requests seems to be quick enough but we could move it to the initialization (or once) and keep this limitation with installing new operators while the member operator is running in mind.

alexeykazakov: The most realistic risk here if the AAP operator is updated and it introduces some API changes which could break the idler (like introducing another AAP CRD which could be in the middle of the owner chain).

controllers/idler/aap_idler.go
alexeykazakov: Yes! Great catch.

controllers/idler/idler_controller.go
mfrancisc: ok makes sense!

 Out of curiosity why do we need to pods to be tracked in Idler.Status ?

alexeykazakov: We don't have to. It was the original design (so we just have something in the status which could help us with debugging problems with the idler) but now it's used in the main idler. We can get rid of it but it would requires some refactoring.

controllers/idler/aap_idler.go
```diff
@@ -0,0 +1,231 @@
...
+
+// ensureAAPIdled checks if the long-running or crash-looping pod belongs to an AAP instance and if so, ensures that the AAP is idled.
+// Returns the AAP resource name in case it was idled, or an empty string if it was not idled.
+func (r *Reconciler) ensureAAPIdled(ctx context.Context, pod corev1.Pod, aapAPI aapAPI) (string, error) {
+	for _, podCond := range pod.Status.Conditions {
+		if podCond.Type == "PodCompleted" {
+			// Pod is in the completed state, no need to idle
+			return "", nil
+		}
+	}
```
MatousJobanek, line 141: this code doesn't seem to be aap-specific, this should rather go into the main logic, correct?

MatousJobanek: We actually want to delete the pod (which we do), but using the logic from the main idler.
I would simply drop it from here as it brings confusion why we have a special handling of these pods in the aap idler.

MatousJobanek:
/retest 
flakiness https://storage.googleapis.com/test-platform-results/pr-logs/pull/codeready-toolchain_member-operator/631/pull-ci-codeready-toolchain-member-operator-master-e2e/1900445170155917312/build-log.txt

controllers/idler/aap_idler.go
```diff
@@ -0,0 +1,266 @@
...
+	timeoutSeconds := aapTimeoutSeconds(idler.Spec.TimeoutSeconds)
+	requeueAfter := time.Duration(timeoutSeconds) * time.Second
+	for _, pod := range podList.Items {
+		startTime := pod.Status.StartTime
+		if startTime == nil {
+			continue
+		}
+		var idledAAPName string
+		pod := pod // TODO We won't need it after upgrading to go 1.22: https://go.dev/blog/loopvar-preview
```
rsoaresd, line 88:
```suggestion
```
We are now using go 1.22, so we can remove it. Nice catch!

controllers/idler/aap_idler.go
```diff
@@ -0,0 +1,266 @@
...
+	// Get all pods running in the namespace
+	podList := &corev1.PodList{}
+	if err := i.allNamespacesClient.List(ctx, podList, client.InNamespace(idler.Name)); err != nil {
+		return 0, err
+	}
+	timeoutSeconds := aapTimeoutSeconds(idler.Spec.TimeoutSeconds)
+	requeueAfter := time.Duration(timeoutSeconds) * time.Second
+	for _, pod := range podList.Items {
+		startTime := pod.Status.StartTime
+		if startTime == nil {
```
rsoaresd, line 84: I think crash looping pods do not have `StartTime`. Do we care about them? I mean, a pod can be progressing for a long time (asking for learning purposes)

controllers/idler/aap_idler.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// getAAPOwner returns the top level owner of the given object if it is an AAP instance.
+func (i *aapIdler) getAAPOwner(ctx context.Context, obj metav1.Object) (metav1.Object, error) {
+	owners := obj.GetOwnerReferences()
+	var owner metav1.OwnerReference
+	if len(owners) == 0 {
+		return nil, nil // No owner
+	}
+	owner = owners[0] // Multiple owners, use the first one
```
rsoaresd, line 205: Is there a particular reason why we pick the first one?

rsoaresd (commented):
Great job 🚀 Just minor questions:

controllers/idler/aap_idler.go
MatousJobanek: they do. See the comment at the field
```
// RFC 3339 date and time at which the object was acknowledged by the Kubelet.
// This is before the Kubelet pulled the container image(s) for the pod.
// +optional
StartTime *metav1.Time `json:"startTime,omitempty" protobuf:"bytes,7,opt,name=startTime"`
```

controllers/idler/aap_idler.go
MatousJobanek: simplicity :-) 
in most of the cases, there is only one owner (TBH, I haven't seen a resources with multiple owners yet)

rsoaresd: Thanksss!

controllers/idler/aap_idler.go
rsoaresd: Thank you!! For some reason, I thought not 😅

controllers/idler/aap_idler.go
MatousJobanek: Thanks for the reminder - removed in fc469abc

controllers/idler/aap_idler.go
```diff
@@ -0,0 +1,265 @@
...
+			// it is already idled - nothing to do
+			return "", nil
+		}
+	}
+
+	// Patch the aap resource by setting spec.idle_aap to true in order to idle it
+	patch := []byte(`{"spec":{"idle_aap":true}}`)
+	_, err = i.dynamicClient.Resource(*i.aapGVR).Namespace(pod.Namespace).Patch(ctx, aap.GetName(), types.MergePatchType, patch, metav1.PatchOptions{})
+	if err != nil {
+		return "", err
```
alexeykazakov, line 189: I wonder if we actually want to return an error here... I'm just a bit concern about potential problems with AAP idling (changed API and/or some new webhook preventing us to patch it or something like this) which would generate an error and potentially (partially) break the entire idler... Maybe it would be safer to log the error but do not return it? WDYT?
I don't have any strong opinion.

controllers/idler/aap_idler.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+// getAAPOwner returns the top level owner of the given object if it is an AAP instance.
+func (i *aapIdler) getAAPOwner(ctx context.Context, obj metav1.Object) (metav1.Object, error) {
+	owners := obj.GetOwnerReferences()
+	var owner metav1.OwnerReference
+	if len(owners) == 0 {
+		return nil, nil // No owner
+	}
+	owner = owners[0] // Multiple owners, use the first one
```
alexeykazakov, line 204:
```suggestion
	owner = owners[0] // In case of multiple owners, use the first one
```

controllers/idler/aap_idler_test.go
```diff
@@ -0,0 +1,563 @@
...
+
+		// then
+		require.NoError(t, err)
+		assert.Greater(t, requeueAfter, (time.Duration(aapTimeoutSeconds(idler.Spec.TimeoutSeconds)/2)-1)*time.Second)
+		assert.Less(t, requeueAfter, (time.Duration(aapTimeoutSeconds(idler.Spec.TimeoutSeconds)/2)+1)*time.Second)
+		interceptedNotify.assertThatCalled(t, runningAAP.GetName())
+		assertAAPsIdled(t, aapIdler, idler.Name, idledAAP.GetName(), runningAAP.GetName())
+	})
+
+	t.Run("all idled", func(t *testing.T) {
```
alexeykazakov, line 204: Maybe I missed it but you always idle one (or none) AAPs in your tests, right? Can we add another AAP to idle to make sure we idle two/multiple (or none)? To test the case when there is more than one AAP to idle.

alexeykazakov (commented):
I can't approve my "own" PR so I just comment on it :) You can approve it for me ;)
The PR looks great! It's a "go" from me. Just a few minor comments/questions.

controllers/idler/idler_controller.go
MatousJobanek: >  We can get rid of it but it would requires some refactoring.

I was thinking about the same, that the list of pods is actually not needed and could be easily removed. As for the refactoring, it would be really small and it would actually simplify the logic a bit. 

I've created a story for it https://issues.redhat.com/browse/KUBESAW-269

controllers/idler/aap_idler_test.go
MatousJobanek: This test was actually supposed to idle the long-running AAP as well as the crashlooping one (so two AAP instances), but there was a bug. Fixed in: [3761ed5](https://github.com/codeready-toolchain/member-operator/pull/631/commits/3761ed51e26c532e3eb6076599a1fd8d8773a5d3)

controllers/idler/aap_idler.go
MatousJobanek: It's a good point. I was thinking about the same,  but I completely forgot to address it. 
At the first glance, any error returned from the AAP idler shouldn't affect the main-idler logic (as it's executed after it), but if the AAP idler keeps returning errors the controller, then it could negatively affect the reconciles scheduled by the main idler. 

addressed in c9725fd8
* it "ignores" errors within the idler (eg. when tracking the owners or trying to patch an AAP instance), simply to give a chance to idle the other pods/AAP instances. It only collects the errors and returns them at the end.
* it doesn't return the errors returned from the AAP idler - it only logs them and updates the Ready condition of the Idler CR

controllers/idler/idler_controller.go
```diff
@@ -122,24 +122,31 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+		}
 	}
+
 	logger.Info("requeueing for next pod to check", "after_seconds", requeueAfter.Seconds())
 	result := reconcile.Result{
 		Requeue:      true,
 		RequeueAfter: requeueAfter,
 	}
+	if aapIdlerFailed {
+		return result, nil
```
alexeykazakov, line 148:
```suggestion
		// Do not return any errors if the AAP idler failed. We just log the error (above), update the Idler CR status and move on.
		return result, nil
```

alexeykazakov (commented):
Looks good! Feel free to merge it.

controllers/idler/idler_controller.go
```diff
@@ -117,11 +122,32 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+		}
+	}
+
 	logger.Info("requeueing for next pod to check", "after_seconds", requeueAfter.Seconds())
 	result := reconcile.Result{
 		Requeue:      true,
 		RequeueAfter: requeueAfter,
 	}
+	if aapIdlerFailed {
+		// Do not return any errors if the AAP idler failed. We just log the error (above), update the Idler CR status and move on.
```
mfrancisc, line 148: maybe I've missed this but I don't see where the update of the idler CR status happens for aap failures.

mfrancisc (approved):
Amazing job integrating the AAP logic and adding the unit tests 🚀 

I only have some minor comments/questions, feel free to ignore them as they're not mandatory for merging this.

In general I found it a bit tricky to follow the unit tests, I think this is due to the high number of dependencies and setup that the tests have.

Not sure if it will help, but I was thinking if splitting the idler into multiple controllers would improve maintainability and readability:
- one controller for "standards" workloads
- one for vms
- one for AAPs
extracting the common functionality into separate packages. 

Also considering that we don't have e2e tests for this it's not difficult to miss some condition that can happen at runtime.

controllers/idler/aap_idler_test.go
```diff
@@ -0,0 +1,580 @@
...
+		aapIdler, interceptedNotify := prepareAAPIdler(t, idler, idledAAP, runningAAP, runningNoSpecAAP, noiseAAP)
+
+		isOwningSomething := false
+		preparePayloadsForAAPIdler(t, aapIdler, func(kind schema.GroupVersionKind, object client.Object) {
+			if kind.Kind == "Deployment" {
+				if !isOwningSomething {
+					require.NoError(t, controllerutil.SetOwnerReference(runningAAP, object, scheme.Scheme))
+					isOwningSomething = true
+				}
+			}
```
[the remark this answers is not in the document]
MatousJobanek: > I was trying to understand why do we need to check this isOwningSomething ?

because we want the AAP instance to own one single deployment, not all of them created by the function `preparePayloadsForAAPIdler`

> Could it just be a switch statement with all kinds inside the function where it is actually needed or set the ownership on those objects upfront ? 

If we want to reuse the already existing functions, then there is not other way. This approach avoids a tons of code duplication and keeps the code flexible enough.

controllers/idler/aap_idler_test.go
```diff
@@ -0,0 +1,580 @@
...
+		interceptedNotify.assertThatCalled(t)
+		assertAAPsIdled(t, aapIdler, idler.Name, idledAAP.GetName())
+	})
+
+	t.Run("with running AAPs, nothing to idle", func(t *testing.T) {
+		// given
+		aapIdler, interceptedNotify := prepareAAPIdler(t, idler, idledAAP, runningAAP, runningNoSpecAAP, noiseAAP)
+
+		isOwningSomething := false
+		preparePayloadsForAAPIdler(t, aapIdler, func(kind schema.GroupVersionKind, object client.Object) {
```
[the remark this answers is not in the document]
MatousJobanek: fixed f169057f

controllers/idler/aap_idler_test.go
```diff
@@ -0,0 +1,580 @@
...
+
+func TestAAPIdler(t *testing.T) {
+	// given
+	idler := &toolchainv1alpha1.Idler{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "john-dev",
+		},
+		Spec: toolchainv1alpha1.IdlerSpec{TimeoutSeconds: oneHour},
+	}
+	idledAAP := newAAP(t, true, "idled-test", idler.Name)
```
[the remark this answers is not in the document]
MatousJobanek: two reasons:
* to have a noise
* and mainly to verify that the logic ignores the idled AAPs: https://github.com/codeready-toolchain/member-operator/pull/631/files#diff-79b76c6fb335e6e635f582d5c79bbd4cadbb780830fa6af6e2e936e78b83253eR140-R157

controllers/idler/aap_idler_test.go
```diff
@@ -0,0 +1,580 @@
...
+					require.NoError(t, controllerutil.SetOwnerReference(runningAAP, object, scheme.Scheme))
+					isOwningSomething = true
+				}
+			}
+		}, idler.Name, "long-", expiredStartTimes(aapTimeoutSeconds(idler.Spec.TimeoutSeconds)))
+		preparePayloadsForAAPIdler(t, aapIdler, func(kind schema.GroupVersionKind, object client.Object) {
+			if kind.Kind == "ReplicaSet" && len(object.GetOwnerReferences()) == 0 {
+				require.NoError(t, controllerutil.SetOwnerReference(runningNoSpecAAP, object, scheme.Scheme))
+			}
+		}, idler.Name, "crashlooping-", freshStartTimes(aapTimeoutSeconds(idler.Spec.TimeoutSeconds)))
```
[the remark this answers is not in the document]
MatousJobanek: good catch - renamed: 5570e35c

controllers/idler/aap_idler_test.go
```diff
@@ -0,0 +1,580 @@
...
+
+		// when
+		requeueAfter, err := aapIdler.ensureAnsiblePlatformIdling(context.TODO(), idler)
+
+		// then
+		require.NoError(t, err)
+		assert.Greater(t, requeueAfter, (time.Duration(aapTimeoutSeconds(idler.Spec.TimeoutSeconds)/2)-1)*time.Second)
+		assert.Less(t, requeueAfter, (time.Duration(aapTimeoutSeconds(idler.Spec.TimeoutSeconds)/2)+1)*time.Second)
+		interceptedNotify.assertThatCalled(t, runningAAP.GetName())
+		assertAAPsIdled(t, aapIdler, idler.Name, idledAAP.GetName(), runningAAP.GetName())
```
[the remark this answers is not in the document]
MatousJobanek: in the previous one, the related pod was short-lived, so it didn't reach the timeout

controllers/idler/aap_idler_test.go
```diff
@@ -0,0 +1,580 @@
...
+			aapIdler, interceptedNotify := prepareAAPIdler(t, idler, idledAAP, runningAAP, runningNoSpecAAP, noiseAAP)
+			preparePayloadsForAAPIdler(t, aapIdler, func(kind schema.GroupVersionKind, object client.Object) {
+				if kind.Kind == "Deployment" {
+					require.NoError(t, controllerutil.SetOwnerReference(runningAAP, object, scheme.Scheme))
+				}
+			}, idler.Name, "long-", expiredStartTimes(aapTimeoutSeconds(idler.Spec.TimeoutSeconds)))
+
+			dynamicCl := aapIdler.dynamicClient.(*fakedynamic.FakeDynamicClient)
+			originalReactions := make([]clienttest.Reactor, len(dynamicCl.ReactionChain))
+			copy(originalReactions, dynamicCl.ReactionChain)
```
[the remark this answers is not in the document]
MatousJobanek: to be able to reset the original logic in the sub-tests; otherwise, the sub-tests would affect each other.

controllers/idler/idler_controller.go
MatousJobanek: here https://github.com/codeready-toolchain/member-operator/blob/f169057f4bc8373535fcf67c7403ef25295a0754/controllers/idler/idler_controller.go#L127
and here https://github.com/codeready-toolchain/member-operator/blob/f169057f4bc8373535fcf67c7403ef25295a0754/controllers/idler/idler_controller.go#L133

controllers/idler/aap_idler_test.go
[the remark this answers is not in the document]
MatousJobanek: There is no change in the running AAP resources, the difference is in the pods they own

controllers/idler/aap_idler_test.go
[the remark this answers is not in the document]
MatousJobanek: > I was trying to understand why do we need to check this isOwningSomething ?

I added comments to the code 325ce185

controllers/idler/aap_idler_test.go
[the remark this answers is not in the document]
MatousJobanek: I updated the names of the tests a8468beb

mfrancisc (approved):
Thanks for the additional changes! 🙏

alexeykazakov:
/retest

Same flaky test?
```
    init.go:214: 
        	Error Trace:	/tmp/toolchain-e2e/testsupport/init.go:214
        	            				/usr/local/go/src/sync/once.go:74
        	            				/usr/local/go/src/sync/once.go:65
        	            				/tmp/toolchain-e2e/testsupport/init.go:192
        	            				/tmp/toolchain-e2e/test/migration/verify/verify_migration_test.go:27
        	Error:      	Received unexpected error:
        	            	Get "https://member-operator-metrics-service-toolchain-member-20121156.apps.ci-op-z295klwr-887d4.devsandboxci.devcluster.openshift.com/metrics": EOF
        	Test:       	TestAfterMigration
        	Messages:   	failed while setting up or waiting for the route to the 'member-operator-metrics' service to be available
``

MatousJobanek:
@mfrancisc I completely forgot to reply to your message in the header of the review:
> In general I found it a bit tricky to follow the unit tests, I think this is due to the high number of dependencies and setup that the tests have.

I understand. I was trying to find a balance between reusing the code to avoid duplication and keeping the readability. It would definitely deserve a deeper refactoring and improved API. 

> Not sure if it will help, but I was thinking if splitting the idler into multiple controllers would improve maintainability and readability:
    one controller for "standards" workloads
    one for vms
    one for AAPs
    extracting the common functionality into separate packages.

This is an interesting idea. Technically speaking, it should be possible. However before doing that I would drop the list of the tracked pods from the Idler.status. 
I believe that we could achieve a similar result with one controller, but with a better API in the code. 
But both options would require a significant refactoring of the whole code to avoid code duplication and improve the code readability.

PR #633, opened 2025-03-03
fix requeue in idler controller

MatousJobanek:
Added watcher on pods with a predicate that should trigger the reconcile only if:
* pod exists in users namespace
* either the startTime was set in the pod's status, or it reached the restart threshold.

two fixes of scheduling the next requeue of the idler controller:
1. drops the `RequeueTimeThreshold` - the original value caused that the controller kept reconciling as crazy in instances with a high number of users 
2. take into account that VMs should live only one twelfth of the idler timeout - the next requeue should be scheduled accordingly.

### DIFF
--- a/controllers/idler/idler_controller.go
+++ b/controllers/idler/idler_controller.go
@@ -125,7 +125,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		}
 	} else {
 		// if nothing is being tracked then only wait for the next reconcile triggered
-		// by starting the pod
+		// by a pod starting
 		logger.Info("no pods being tracked")
 	}
 
--- a/controllers/idler/idler_controller.go
+++ b/controllers/idler/idler_controller.go
@@ -37,7 +37,7 @@ import (
 )
 
 const (
-	RestartThreshold = 50
+	restartThreshold = 50
 )
 
 var SupportedScaleResources = map[schema.GroupVersionKind]schema.GroupVersionResource{
@@ -125,7 +125,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		}
 	} else {
 		// if nothing is being tracked then only wait for the next reconcile triggered
-		// by starting (or crashlooping) the pod
+		// by starting the pod
 		logger.Info("no pods being tracked")
 	}
 
@@ -153,7 +153,7 @@ func (r *Reconciler) ensureIdling(ctx context.Context, idler *toolchainv1alpha1.
 		if trackedPod := findPodByName(idler, pod.Name); trackedPod != nil {
 			// check the restart count for the trackedPod
 			restartCount := getHighestRestartCount(pod.Status)
-			if restartCount > RestartThreshold {
+			if restartCount > restartThreshold {
 				podLogger.Info("Pod is restarting too often. Killing the pod", "restart_count", restartCount)
 				// Check if it belongs to a controller (Deployment, DeploymentConfig, etc) and scale it down to zero.
 				err := deletePodsAndCreateNotification(podCtx, pod, r, idler)
--- a/controllers/idler/idler_controller.go
+++ b/controllers/idler/idler_controller.go
@@ -35,7 +35,7 @@ import (
 
 const (
 	RestartThreshold     = 50
-	RequeueTimeThreshold = 300 * time.Second
+	RequeueTimeThreshold = 6 * time.Hour
 )
 
 var SupportedScaleResources = map[schema.GroupVersionKind]schema.GroupVersionResource{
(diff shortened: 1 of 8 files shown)


metlos (commented):
Nice simplification of the code 👍🏼

controllers/idler/idler_controller.go
```diff
@@ -35,7 +35,7 @@ import (
 
 const (
 	RestartThreshold     = 50
-	RequeueTimeThreshold = 300 * time.Second
+	RequeueTimeThreshold = 6 * time.Hour
```
alexeykazakov, line 38: I wonder if we can make it a bit shorter like 3h? Basically crash-lopping pods will be killed after the threshold is reached (50 restarts) plus up to `RequeueTimeThreshold`. With this change it's up to 6h instead of 5m extra time after the threshold. Which is fine. But potentially could be shorter (3h???) if it's not too often for the idler with many namespaces.
But again, it's minor. We can keep it as 6h for now and shorten it after we switch to a pod watcher.

controllers/idler/idler_controller.go
```diff
@@ -107,30 +107,36 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
 	}
+	requeueAfter := shorterDuration(RequeueTimeThreshold, time.Duration(idler.Spec.TimeoutSeconds)*time.Second)
 	newStatusPods := make([]toolchainv1alpha1.Pod, 0, 10)
 	for _, pod := range podList.Items {
 		pod := pod // TODO We won't need it after upgrading to go 1.22: https://go.dev/blog/loopvar-preview
 		podLogger := log.FromContext(ctx).WithValues("pod_name", pod.Name, "pod_phase", pod.Status.Phase)
 		podCtx := log.IntoContext(ctx, podLogger)
+		timeoutSeconds := idler.Spec.TimeoutSeconds
+		if isOwnedByVM(pod.ObjectMeta) {
+			// use 1/12th of the timeout for VMs
```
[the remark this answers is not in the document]
alexeykazakov: The reason is that we want to be more aggressive with VM idling. Ideally we should extend our Idlier API and introduce a new field there. Spec.VMTimeoutSeconds or something like this. Having 1/12 of the Spec.TimeoutSecond for VM was just a quick way for us (for having more aggressive VM Idler).

alexeykazakov (approved):

controllers/idler/idler_controller.go
ranakan19: +1

ranakan19 (commented):
Thanks for the changes!

Couple of thought/comments/questions:
1. Sorry, could you please clarify the issue with shorter VM idling mentioned in point 2 of PR description? Was it not functioning as expected before, aside from the 5m requeue time being too aggressive causing issues?
2. The current implementation checks restartCount against the threshold but doesn’t tightly couple requeueTime to it. Previously also the logic for restartCount wasn't tied to timeout, but shorter requeue time would almost instantly idle the pod once it reaches restartThreshold. For instance, reaching a restartCount of 50 takes about 4 hours, yet idling a crash-looping pod with requeueTimeThreshold set to 6 hours can take anywhere between 6 to 12 hours. As Alexey pointed out its still better than not idling crashlooping pods at all, but a requeueTimethreshold shorter than 6 hr should help too.

controllers/idler/mapper.go
```diff
@@ -0,0 +1,19 @@
...
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// MapPodToIdler maps the pod to the idler
+func MapPodToIdler(_ context.Context, obj runtimeclient.Object) []reconcile.Request {
+	return []reconcile.Request{{
+		// the idler should have the same name as the user's namespace
+		NamespacedName: types.NamespacedName{
+			Name: obj.GetNamespace(),
```
metlos, line 16: Does not specifying the namespace of the idler resource work correctly with the controller manager?

controllers/idler/predicate.go
```diff
@@ -0,0 +1,52 @@
...
+// and if either the highest restart count is higher than the threshold
+// or the startTime was newly set in the new version of the pod
+func (p PodIdlerPredicate) Update(event runtimeevent.UpdateEvent) bool {
+	isUserPod, newPod := isUserPod(event.ObjectNew)
+	if !isUserPod {
+		return false
+	}
+	if oldPod, ok := event.ObjectOld.(*corev1.Pod); ok {
+		startTimeNewlySet := oldPod.Status.StartTime == nil && newPod.Status.StartTime != nil
+		return startTimeNewlySet || getHighestRestartCount(newPod.Status) > RestartThreshold
```
metlos, line 23: Maybe have the `RestartThreshold` as a field of the predicate instead of using the global variable? That'll a) make the code slightly more local and therefore IMHO more readable and b) open it up to be user-configurable.

MatousJobanek: it should use the same value as is used in idler_controller.go. There is nothing to be configurable for now, if we wanted to have it configurable, then it should go into the ToolchainConfig, but this is not really needed.
I unexported the constant to make it clear that it's not a global variable, but specific to the controller. 2001124d

alexeykazakov (approved):

ranakan19:
> I was mainly looking at the logic where it calculated the requeue time from the list of tracked pods, and this logic didn't take into account the pods owned by VM

Ah because only the timeoutsecond variable was updated within the tracledPods loop, while the requeuetime calculation relied on idler spec! I see now, thank you so much :)

ranakan19 (approved):
Great improvements! predicate looks real nice.
minor - the testIdlertimeoutseconds const in idler_controller_test can be a lower value now

main.go
```diff
@@ -248,7 +249,7 @@ func main() {
 		DynamicClient:       dynamicClient,
 		GetHostCluster:      cluster.GetHostCluster,
 		Namespace:           namespace,
-	}).SetupWithManager(mgr); err != nil {
+	}).SetupWithManager(mgr, allNamespacesCluster); err != nil {
```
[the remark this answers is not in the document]
MatousJobanek: it's actually the same. 
https://github.com/codeready-toolchain/member-operator/blob/fc363451b7190fd597f62cb968f9d6bf74be5a35/main.go#L199
https://github.com/codeready-toolchain/member-operator/blob/fc363451b7190fd597f62cb968f9d6bf74be5a35/main.go#L338-L345
The code could be a bit refactored to make it clear

controllers/idler/idler_controller_test.go
```diff
@@ -107,13 +107,13 @@ func TestEnsureIdling(t *testing.T) {
 
 	logf.SetLogger(zap.New(zap.UseDevMode(true)))
 
-	t.Run("No pods in namespace managed by idler", func(t *testing.T) {
+	t.Run("No pods in namespace managed by idler, idler timeout is longer than requeue-time-threshold", func(t *testing.T) {
```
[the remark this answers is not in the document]
MatousJobanek: of course :-) 
good catch, thanks,  fixed in 4e807fb0

MatousJobanek:
>  minor - the testIdlertimeoutseconds const in idler_controller_test can be a lower value now

it's a unit-test, it's just a number. TBH, it's sometimes even better to keep it higher to have a room for playing with the age of the workload and other numbers, so keeping it as it is

rajivnathan (approved):
Nice! 👍

PR #634, opened 2025-03-06
use SSA to apply the template objects

metlos:
This is a minimal change in the member operator to start using the SSA to persist the changes to the templates.

I didn't do this change in `toolchain-common` to limit the "impact" this change would have at runtime across the ecosystem (i.e. not impact host-operator or ksctl).

The future steps for this would be to drastically simplify the template apply logic in `toolchain-common`.

As such, this is still more of an experiment than a definitive solution. At best, it is the first step in the larger effort to simplify the deployment of NSTemplateSets.

### DIFF
--- a/controllers/nstemplateset/client.go
+++ b/controllers/nstemplateset/client.go
@@ -2,12 +2,10 @@ package nstemplateset
 
 import (
 	"context"
-	"strings"
+	"fmt"
 
 	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
-	applycl "github.com/codeready-toolchain/toolchain-common/pkg/client"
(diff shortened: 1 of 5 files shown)


MatousJobanek:
Thanks a lot  for the PR :+1: 

> I didn't do this change in toolchain-common to limit the "impact" this change would have at runtime across the ecosystem (i.e. not impact host-operator or ksctl).

I would simply create a new function "next" to the existing one (in a separate file/package) and gradually propagate the change to all places in our code. 
In this way, it would be better to isolate & review the actual "apply" code in toolchain-common repo, and also isolate the changes in separate PRs while propagating the new logic to the respective places in our operators.

In other words, can you please move the "apply" logic to toolchain-common?

PR #638, opened 2025-03-10
force idler reconcile

MatousJobanek:
if the predicate allows reconciliation based on the set startTime, but the client cache is not updated with the change, then the pod won't be tracked and the next reconciliation won't be scheduled based on the age of the pod. 
To avoid this problem, let's force the next reconciliation:
* to the idler timeout if the pod that lacks the starttime is non-VM pod
* to the 1/12th of the idler timeout if the pod that lacks the starttime is a VM pod

[SANDBOX-1023](https://issues.redhat.com/browse/SANDBOX-1023)

https://redhat-internal.slack.com/archives/CHK0J6HT6/p1741610661596739

### DIFF
--- a/controllers/idler/idler_controller.go
+++ b/controllers/idler/idler_controller.go
@@ -116,19 +116,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 		return reconcile.Result{}, r.wrapErrorWithStatusUpdate(ctx, idler, r.setStatusFailed, err,
 			"failed to ensure idling '%s'", idler.Name)
 	}
-	result := reconcile.Result{}
-	if len(idler.Status.Pods) > 0 { // schedule the next requeue only if something is being tracked
-		logger.Info("requeueing for next pod to check", "after_seconds", requeueAfter.Seconds())
-		result = reconcile.Result{
-			Requeue:      true,
-			RequeueAfter: requeueAfter,
-		}
-	} else {
-		// if nothing is being tracked then only wait for the next reconcile triggered
-		// by a pod starting
-		logger.Info("no pods being tracked")
+	logger.Info("requeueing for next pod to check", "after_seconds", requeueAfter.Seconds())
+	result := reconcile.Result{
+		Requeue:      true,
+		RequeueAfter: requeueAfter,
 	}
-
 	return result, r.setStatusReady(ctx, idler)
 }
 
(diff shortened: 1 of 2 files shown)


xcoulon (approved):

rajivnathan (approved):
Nice 👍

controllers/idler/idler_controller.go
```diff
@@ -181,9 +173,14 @@ func (r *Reconciler) ensureIdling(ctx context.Context, idler *toolchainv1alpha1.
 				StartTime: *pod.Status.StartTime,
 			})
 		}
+		// calculate the next reconcile
```
ranakan19, line 176: can we combine this if-else block with the above if-elseif block since the condition `pod.Status.StartTime != nil` is repeating.

So maybe something like, into a if-elseif-else block: 
```
if trackedPod: = findPodByName(idler, pod.Name);
trackedPod != nil {
    ....
... no..change
} else if pod.Status.StartTime != nil {
    podLogger.Info("New pod detected. Start tracking.")
    newStatusPods = append(newStatusPods, toolchainv1alpha1.Pod {
        Name: pod.Name,
        StartTime: * pod.Status.StartTime,
    })
    killAfter: = time.Until(pod.Status.StartTime.Add(time.Duration(timeoutSeconds + 1) * time.Second))
    requeueAfter = shorterDuration(requeueAfter, killAfter)
}
else {
    // if the pod doesn't contain startTime, then schedule the next reconcile to the timeout
    // if not already scheduled to an earlier time
    requeueAfter = shorterDuration(requeueAfter, time.Duration(timeoutSeconds) * time.Second)
}
```

ranakan19: can we also remove the comment on line 177/169 `// Ignore pods without StartTime`

MatousJobanek: I was thinking about it too, but it's not  the same.
we need to calculate the "killAfter" for all pods, not only those that are no being tracked:
```
    killAfter: = time.Until(pod.Status.StartTime.Add(time.Duration(timeoutSeconds + 1) * time.Second))
    requeueAfter = shorterDuration(requeueAfter, killAfter)
```
so it needs to stay outside

MatousJobanek: There is a way to unify it a bit, but let's keep it as a separate PR

ranakan19: aha, makes sense, thanks for the explanation! now the comment above this separate block makes so much more sense! I wonder if the comment can be more verbose - but can't think of it anything anymore after your explanation.

ranakan19 (approved):

alexeykazakov (approved):

PR #642, opened 2025-03-18
Update VM webhooks for 4.18

rajivnathan:
Looks like the `spec.runStrategy` field in the VM yaml is now required and is causing VM creation failure due to the current validating webhook.

This change updates the following:
- VM mutating webhook always applies a patch on CREATE to set the runStrategy to `Manual`
- VM validating webhook no longer checks runStrategy on CREATE but will block UPDATE requests where runStrategy is set to anything other than `Manual`


xcoulon (approved):

rajivnathan:
Closing in favour of https://github.com/codeready-toolchain/member-operator/pull/643

PR #647, opened 2025-03-19
Update vm idler

rajivnathan:


controllers/idler/idler_controller.go
```diff
@@ -609,9 +616,14 @@ func (r *Reconciler) stopVirtualMachine(ctx context.Context, namespace string, o
 		return "", "", false, err
 	}
 
-	// patch the virtualmachine resource by setting spec.running to false in order to stop the VM
-	patch := []byte(`{"spec":{"running":false}}`)
-	_, err = r.DynamicClient.Resource(vmGVR).Namespace(namespace).Patch(ctx, vm.GetName(), types.MergePatchType, patch, metav1.PatchOptions{})
+	err = r.RestClient.Put().
```
metlos, line 619: Wouldn't `r.DynamciClient.Resource(vmGVR).Namespace(namespace).Patch(ctx, vm.GetName(), types.MergePatchType, patch, metav1.PatchOptions{}, "stop")` work somehow, too, instead of using the lowest level client there is (possibly with some changes to `patch` or some such)? :)

main.go
```diff
@@ -207,6 +207,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	restClient, err := newRestClient(cfg)
+	if err != nil {
+		setupLog.Error(err, "unable to create scales client")
```
mfrancisc, line 212:
```suggestion
		setupLog.Error(err, "unable to create rest client")
```

mfrancisc (approved):
Great job 👍 , thanks for fixing it!

PR #648, opened 2025-03-21
feat: update nstemplateset revisions status

mfrancisc:
This PR introduces changes to the nstemplateset controller so that it sets/updates the `NSTemplateSet.Status` field with the `clusterResources` , `namespaces` , `spaceRoles` and `featureToggles` from the `NSTemplateSet.Spec`. As of now the `NSTemplateSet` still uses `TierTemplate` refs , but in the future it will use `TierTemplateRevision` refs.

The status field is not yet used anywhere , but it will be used to drive updates across  the namspaces. Next changes will introduce logic such that when there is a drift between the templates `NSTemplateSet.Spec` and `NSTemplateSet.Status` than the controller will start updating the namespaces to bring them in synch with the templates in the Spec.

Jira: https://issues.redhat.com/browse/KUBESAW-142

e2e: https://github.com/codeready-toolchain/toolchain-e2e/pull/1137


metlos (commented):
A small nitpick about the test placement but lgtm otherwise...

controllers/nstemplateset/features_test.go
```diff
@@ -99,9 +101,9 @@ func TestShouldCreate(t *testing.T) {
 }
 
 func TestReallySplit(t *testing.T) {
-	assert.Empty(t, splitCommaSeparatedList(""))
-	assert.Equal(t, strings.Split("1,2,3", ","), splitCommaSeparatedList("1,2,3"))
-	assert.Equal(t, strings.Split("1", ","), splitCommaSeparatedList("1"))
+	assert.Empty(t, utils.SplitCommaSeparatedList(""))
```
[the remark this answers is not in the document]
mfrancisc: yup, I've moved the function to common as it is being used by the e2e tests as well.

See: https://github.com/codeready-toolchain/member-operator/pull/648/commits/66866e4730d696c49e0b0389c878b76a996ebeb4

metlos (approved):

controllers/nstemplateset/nstemplateset_controller.go
```diff
@@ -134,14 +134,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
...
+		// set the provisioned templates in the status as refs
+		return reconcile.Result{}, r.status.updateStatusClusterResourcesRevisions(ctx, nsTmplSet) // wait for cluster resources to be created
 	}
 
 	if createdOrUpdated, err := r.namespaces.ensure(ctx, nsTmplSet); err != nil {
 		logger.Error(err, "failed to either provision or update user namespaces")
 		return reconcile.Result{}, err
 	} else if createdOrUpdated {
-		return reconcile.Result{}, nil // something in the watched resources has changed - wait for another reconcile
+		return reconcile.Result{}, r.status.updateStatusNamespacesRevisions(ctx, nsTmplSet) // something in the watched resources has changed - wait for another reconcile
```
MatousJobanek, line 145: IIRC, the current logic updates the label at the namespace only when it's done with the deletion, update, and creation of all inner resources.
We should follow the same approach. We should not update the revisions if we are still not completely done.
I understand that before refactoring the controller and merging the logic together, it can be bit weird/tricky, but I would move this out of this if statement, so it's updated when it's done

controllers/nstemplateset/status.go
```diff
@@ -88,6 +91,37 @@ func (r *statusManager) setStatusReady(ctx context.Context, nsTmplSet *toolchain
...
+	nsTmplSet.Status.SpaceRoles = nsTmplSet.Spec.SpaceRoles
+}
+
+func (r *statusManager) updateStatusClusterResourcesRevisions(ctx context.Context, nsTmplSet *toolchainv1alpha1.NSTemplateSet) error {
+	nsTmplSet.Status.ClusterResources = nsTmplSet.Spec.ClusterResources
+	featureAnnotation, featureAnnotationFound := nsTmplSet.Annotations[toolchainv1alpha1.FeatureToggleNameAnnotationKey]
+	if featureAnnotationFound {
+		// save the feature toggles into the status
+		nsTmplSet.Status.FeatureToggles = utils.SplitCommaSeparatedList(featureAnnotation)
+	}
```
MatousJobanek, line 111: it's currently used only for cluster-scoped resources, but could be technically used for namespace-scoped resources as well. 
it's worth adding a comment that it should be more generic and not tied to cluster-scoped template

controllers/nstemplateset/nstemplateset_controller.go
mfrancisc: yes, if I'm understanding this correctly it happens here: https://github.com/mfrancisc/member-operator/blob/c616b3f116561952151101c704da19173ebce3f0/controllers/nstemplateset/namespaces.go#L225-L229

and it's when actually it returns `createdOrUpdated` set to `true` here: https://github.com/mfrancisc/member-operator/blob/c616b3f116561952151101c704da19173ebce3f0/controllers/nstemplateset/namespaces.go#L73

We can move it outside anyways but wanted to update the revision only if there is a change.

controllers/nstemplateset/status.go
mfrancisc: good point. done pls see: https://github.com/codeready-toolchain/member-operator/pull/648/commits/538ffec22b4abdaaf58db4964168e758504aae45

controllers/nstemplateset/nstemplateset_controller.go
mfrancisc: I've moved the status updated outside the if statements as discussed in the synch call.

controllers/nstemplateset/status.go
```diff
@@ -88,6 +90,56 @@ func (r *statusManager) setStatusReady(ctx context.Context, nsTmplSet *toolchain
...
+}
+
+// featureAnnotationNeedsUpdate checks if the feature annotation has changed on the nstemlpateset compared to what was last time saved in the status
+func featureAnnotationNeedsUpdate(nsTmplSet *toolchainv1alpha1.NSTemplateSet) (bool, []string) {
+	featureAnnotation := nsTmplSet.Annotations[toolchainv1alpha1.FeatureToggleNameAnnotationKey]
+	featureAnnotationList := utils.SplitCommaSeparatedList(featureAnnotation)
+	if !reflect.DeepEqual(featureAnnotationList, nsTmplSet.Status.FeatureToggles) {
+		return true, featureAnnotationList
+	}
+	return false, featureAnnotationList
```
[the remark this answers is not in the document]
mfrancisc: yeah, it is shorter. Thanks!

I've wrote the longer version because it requires me to read more but think less 😄 . But this looks good 👍

controllers/nstemplateset/status.go
```diff
@@ -88,6 +90,56 @@ func (r *statusManager) setStatusReady(ctx context.Context, nsTmplSet *toolchain
...
+}
+
+// clusterResourcesNeedsUpdate checks if there is a drift between the cluster resources set in the spec and the status of the nstemplateset
+func clusterResourcesNeedsUpdate(nsTmplSet *toolchainv1alpha1.NSTemplateSet) bool {
+	if (nsTmplSet.Status.ClusterResources == nil && nsTmplSet.Spec.ClusterResources != nil) ||
+		(nsTmplSet.Status.ClusterResources != nil && nsTmplSet.Spec.ClusterResources == nil) ||
+		nsTmplSet.Status.ClusterResources.TemplateRef != nsTmplSet.Spec.ClusterResources.TemplateRef {
+		return true
+	}
+	return false
```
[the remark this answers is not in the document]
mfrancisc: thank you!

mfrancisc:
/retest 

updated e2e

controllers/nstemplateset/nstemplateset_controller.go
```diff
@@ -136,12 +136,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.
 	} else if createdOrUpdated {
 		return reconcile.Result{}, nil // wait for cluster resources to be created
 	}
+	if err := r.status.updateStatusClusterResourcesRevisions(ctx, nsTmplSet); err != nil {
```
metlos, line 139: Do we need to update the revisions 3 times during the reconcile? I get that we do it after we apply the 3 different kinds of resources from the template set but apart from maybe easier state management in case of errors, is there any benefit to doing it this way instead of just once at (each) end of the reconcile method beyond this point?

controllers/nstemplateset/status.go
```diff
@@ -88,6 +90,50 @@ func (r *statusManager) setStatusReady(ctx context.Context, nsTmplSet *toolchain
...
+		return r.Client.Status().Update(ctx, nsTmplSet)
+	}
+	return nil
+}
+
+// featureAnnotationNeedsUpdate checks if the feature annotation has changed on the nstemlpateset compared to what was last time saved in the status
+func featureAnnotationNeedsUpdate(nsTmplSet *toolchainv1alpha1.NSTemplateSet) (bool, []string) {
+	featureAnnotation := nsTmplSet.Annotations[toolchainv1alpha1.FeatureToggleNameAnnotationKey]
+	featureAnnotationList := utils.SplitCommaSeparatedList(featureAnnotation)
+	return !reflect.DeepEqual(featureAnnotationList, nsTmplSet.Status.FeatureToggles), featureAnnotationList
```
metlos, line 111: This assumes the same order of the features in the list. Do we want that?

If we do, `slices.Equal()` would IMHO be preferable to `reflect.DeepEqual()`.
If we don't we could use the `sets` package in k8s API and do `sets.New(..).Equal(sets.New(..))`.

controllers/nstemplateset/status.go
```diff
@@ -88,6 +90,50 @@ func (r *statusManager) setStatusReady(ctx context.Context, nsTmplSet *toolchain
...
+
+// clusterResourcesNeedsUpdate checks if there is a drift between the cluster resources set in the spec and the status of the nstemplateset
+func clusterResourcesNeedsUpdate(nsTmplSet *toolchainv1alpha1.NSTemplateSet) bool {
+	return (nsTmplSet.Status.ClusterResources == nil && nsTmplSet.Spec.ClusterResources != nil) ||
+		(nsTmplSet.Status.ClusterResources != nil && nsTmplSet.Spec.ClusterResources == nil) ||
+		nsTmplSet.Status.ClusterResources.TemplateRef != nsTmplSet.Spec.ClusterResources.TemplateRef
+}
+
+func (r *statusManager) updateStatusNamespacesRevisions(ctx context.Context, nsTmplSet *toolchainv1alpha1.NSTemplateSet) error {
+	if !reflect.DeepEqual(nsTmplSet.Spec.Namespaces, nsTmplSet.Status.Namespaces) {
```
metlos, line 122: Is it safe to assume the same order of elements? As above, regardless of the answer, there are better alternatives to `reflect.DeepEqual()`.

controllers/nstemplateset/status.go
```diff
@@ -88,6 +90,50 @@ func (r *statusManager) setStatusReady(ctx context.Context, nsTmplSet *toolchain
...
+		nsTmplSet.Status.Namespaces = nsTmplSet.Spec.Namespaces
+		return r.Client.Status().Update(ctx, nsTmplSet)
+	}
+	return nil
+}
+
+func (r *statusManager) updateStatusSpaceRolesRevisions(ctx context.Context, nsTmplSet *toolchainv1alpha1.NSTemplateSet) error {
+	if !reflect.DeepEqual(nsTmplSet.Spec.SpaceRoles, nsTmplSet.Status.SpaceRoles) {
```
metlos, line 130: Is it safe to assume the same order of elements? As above, regardless of the answer, there are better alternatives to `reflect.DeepEqual()` (complicated by the fact that `NSTemplateSetSpaceRole` contains an array).

metlos (approved):
Sorry for the late review. Generally, LGTM but I think there are places that would benefit from added comments explaining the implementation choice.

controllers/nstemplateset/nstemplateset_controller.go
mfrancisc: As Matous said this seemed to be the most straightforward way to store the information. Also, IMO it is more accurate to update the status as soon as we are done with one type/kind of template, since things might go wrong later in the code and we don't have to start again from scratch , especially when the update of the namespace will rely on this comparison logic that I'm introducing.

controllers/nstemplateset/status.go
mfrancisc: yeah we don't care about ordering, I though about that but I was too lazy to sort those since they will always be in the same order. Anyways is a good practice to not rely on order if it doesn't matter , so I've used a different approach see https://github.com/codeready-toolchain/member-operator/pull/648/commits/40b6fe52be26a984146b98a5b6ab43d36e85f9b0

controllers/nstemplateset/status.go
mfrancisc: good point, see https://github.com/codeready-toolchain/member-operator/pull/648/commits/40b6fe52be26a984146b98a5b6ab43d36e85f9b0

mfrancisc:
/retest

looks like some potential flakiness https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/codeready-toolchain_member-operator/648/pull-ci-codeready-toolchain-member-operator-master-e2e/1909182636639653888/build-log.txt:
```
=== NAME  TestCreateVirtualMachine/vm-domain-memory-guest
    clean.go:175: 
        	
        	Test:       	TestCreateVirtualMachine/vm-domain-memory-guest
        	Messages:   	The proper cleanup of the UserSignup 'test-vm-3' and related resources wasn't finished within the given timeout
```

controllers/nstemplateset/status.go
```diff
@@ -88,6 +90,76 @@ func (r *statusManager) setStatusReady(ctx context.Context, nsTmplSet *toolchain
 		})
 }
 
+// updateStatusClusterResourcesRevisions updates the cluster resources and features list in the status of the nstemplateset
+func (r *statusManager) updateStatusClusterResourcesRevisions(ctx context.Context, nsTmplSet *toolchainv1alpha1.NSTemplateSet) error {
+	updateFeatureAnnotation, featureAnnotation := featureAnnotationNeedsUpdate(nsTmplSet)
+	if updateFeatureAnnotation || clusterResourcesNeedsUpdate(nsTmplSet) {
+		// save the feature toggles into the status
+		// we updated also the featureToggles since the current the logic handles only cluster scoped resources.
```
fbm3307, line 98:
```suggestion
		// we updated also the featureToggles since the current logic handles only cluster scoped resources.
```

fbm3307 (approved):
Great Work !!

mfrancisc:
/retest

cleanup issue again https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/codeready-toolchain_member-operator/648/pull-ci-codeready-toolchain-member-operator-master-e2e/1909524417713541120/build-log.txt

mfrancisc:
@MatousJobanek @metlos @fbm3307 could you PTAL when you have some time, I've pushed some new changes after you approved. Apologies  for the additional work 🙏

controllers/nstemplateset/status.go
```diff
@@ -145,12 +145,16 @@ func (r *statusManager) updateStatusSpaceRolesRevisions(ctx context.Context, nsT
...
-			})
 			return out[i].TemplateRef < out[j].TemplateRef
 		})
+		// sort usernames within the space role
+		for i := range out {
+			sortedUsernames := append([]string{}, out[i].Usernames...) // Copy input to avoid mutating it
+			sort.Slice(sortedUsernames, func(x, y int) bool {
+				return sortedUsernames[x] < sortedUsernames[y]
+			})
+			out[i].Usernames = sortedUsernames
```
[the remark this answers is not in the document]
mfrancisc: yup that works too, thanks for the suggestion!

fbm3307 (approved):

MatousJobanek (approved):
nice :+1:

PR #649, opened 2025-03-24
idler: patch instead of update

MatousJobanek:
use patch instead of update when scaling down controllers to avoid conflicts https://redhat-internal.slack.com/archives/C064SKK3C2E/p1742656300541629?thread_ts=1742645702.696489&cid=C064SKK3C2E

### DIFF
--- a/controllers/idler/idler_controller.go
+++ b/controllers/idler/idler_controller.go
@@ -9,7 +9,6 @@ import (
 	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
 	"github.com/codeready-toolchain/toolchain-common/pkg/condition"
 	notify "github.com/codeready-toolchain/toolchain-common/pkg/notification"
-	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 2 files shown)


alexeykazakov (approved):
Thanks!

controllers/idler/idler_controller.go
```diff
@@ -416,46 +415,24 @@ func (r *Reconciler) scaleDeploymentToZero(ctx context.Context, namespace string
...
-				scale := autoscalingv1.Scale{
-					ObjectMeta: ctrl.ObjectMeta{
-						Name:      deploymentOwner.Name,
-						Namespace: d.Namespace,
-					},
-					Spec: autoscalingv1.ScaleSpec{
-						Replicas: zero,
-					},
-				}
-				_, err = r.ScalesClient.Scales(d.Namespace).Update(ctx, supportedScaleResource.GroupResource(), &scale, metav1.UpdateOptions{})
```
mfrancisc, line 443: out of curiosity - why is this case not needed anymore ?

mfrancisc (approved):
Nice job 👍 

I have only one minor question/comment.

metlos (approved):
Approved but I am also curious as @mfrancisc in https://github.com/codeready-toolchain/member-operator/pull/649/files?diff=split&w=0#r2011721511 why we needed to handle an internal server error and why we don't have to anymore with a patch...

MatousJobanek:
@metlos @mfrancisc for the context, here is the PR that introduced the check for the error https://github.com/codeready-toolchain/member-operator/pull/388

PR #657, opened 2025-03-27
Ignore the expected error when stopping VM

rajivnathan:
See https://redhat-internal.slack.com/archives/CHK0J6HT6/p1743092974483949


alexeykazakov (approved):
I'm approving it but I think it's better to go back to the REST client. It looks like dynamic clients do not work properly with VMs OOTB and it's getting a bit risky to keep "customizing" it just to make it work for VMs. Future VM changes could require even more tweaks. And the dynamic client is used by the AAP idler as well, so further customization potentially could affect it too.

rajivnathan:
Closing this, we'll revert the original PR instead and just use the REST client since the VM stop API doesn't seem to be made for the dynamic client

PR #659, opened 2025-03-28
idler: fix flaky unit tests

MatousJobanek:
https://redhat-internal.slack.com/archives/CHK0J6HT6/p1743169699845159?thread_ts=1743166514.789429&cid=CHK0J6HT6
* fix the VM start times to avoid flakiness
* increase the idler timeout to 3 hours (to have something more real, but the current logic should work with any value)
* use common assertion logic for verifying the delta in which the returned requeue time should be
* increase the delta to 10 seconds for cases when it takes more time to execute the test

### DIFF
--- a/controllers/idler/idler_controller_test.go
+++ b/controllers/idler/idler_controller_test.go
@@ -49,7 +49,7 @@ const (
 	RestartCountWithinThresholdContainer1 = 30
 	RestartCountWithinThresholdContainer2 = 24
 	RestartCountOverThreshold             = 52
-	TestIdlerTimeOutSeconds               = 540
+	TestIdlerTimeOutSeconds               = 60 * 60 * 3 // 3 hours
 	apiEndpoint                           = "https://api.openshift.com:6443"
 )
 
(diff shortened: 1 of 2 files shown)


controllers/idler/idler_controller_test.go
```diff
@@ -1861,21 +1858,28 @@ func newMUR(name string) *toolchainv1alpha1.MasterUserRecord {
...
+
 func freshStartTimes(timeoutSeconds int32) payloadStartTimes {
 	halfOfIdlerTimeoutAgo := time.Now().Add(-time.Duration(timeoutSeconds/2) * time.Second)
-	twelfthOfIdlerTimeoutMinusOneSecondAgo := time.Now().Add(-time.Duration(timeoutSeconds/12-1) * time.Second) // needs to be smaller than 1/12 which is the vm idler time
+	// needs to be smaller than 1/12 which is the vm idler time, let's set ot to 1/24
+	twentyFourthOfIdlerTimeoutMinusOneSecondAgo := time.Now().Add(-time.Duration(timeoutSeconds/24) * time.Second)
 	return payloadStartTimes{
 		defaultStartTime: halfOfIdlerTimeoutAgo,
-		vmStartTime:      twelfthOfIdlerTimeoutMinusOneSecondAgo, // vms are killed in 1/12 of the idler time
+		vmStartTime:      twentyFourthOfIdlerTimeoutMinusOneSecondAgo, // vms are killed in 1/12 of the idler time
```
MatousJobanek, line 1873: this was too close to the idler timeout, so it could have causes flakiness

alexeykazakov (approved):

PR #661, opened 2025-04-04
KUBESAW-273: Replace go-bindata with go:embed

fbm3307:
We want to replace go-bindata with go:embed so that we don't rely on 3rd party libraries and use functionality provided as part of Go itself.

This PR is is related to -https://github.com/codeready-toolchain/host-operator/pull/1165 
This is for converting nstemplatetiers to use goembed instead of gobindata


make/go.mk
```diff
@@ -48,6 +39,9 @@ tidy:
 vet:
 	go vet ./...
 
+.PHONY: generate-assets
+generate-assets:
```
rsoaresd, line 43:
```suggestion

```

rsoaresd (commented):
Great work! Just minor suggestion:

make/go.mk
fbm3307: I added that to make the CI happy, since we have test-with-coverage action which expectes this , so i will remove this in a spearate PR

rsoaresd: Sorry! Thanks for the explanation!

rsoaresd:
@fbm3307 is there any reason why we do not delete `tools.go` in this PR? (just asking for learning)

fbm3307:
> @fbm3307 is there any reason why we do not delete `tools.go` in this PR? (just asking for learning)

@rsoaresd great catch, i just forgot to delete the file, done now 
Thanks

fbm3307:
/test e2e

looks like a flaky test failure https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/codeready-toolchain_member-operator/661/pull-ci-codeready-toolchain-member-operator-master-e2e/1909205624713908224

deploy/resources.go
```diff
@@ -4,3 +4,9 @@ import "embed"
 
 //go:embed templates/toolchaincluster/*
 var ToolchainClusterTemplateFS embed.FS
+
+//go:embed autoscaler/*
```
mfrancisc, line 8: minor - should we move these to the templates folder as the ones above ?

deploy/resources.go
```diff
@@ -4,3 +4,9 @@ import "embed"
...
+var AutoScalerFS embed.FS
+
+//go:embed webhook/*
```
mfrancisc, line 11: same as above

make/go.mk
```diff
@@ -48,6 +39,9 @@ tidy:
...
+
 .PHONY: pre-verify
-pre-verify: generate-assets
+pre-verify:
```
mfrancisc, line 46: could we delete this target or will this cause issues with the verify-dependencies script ?

mfrancisc (commented):
Nice job !  I have only few minor comments.

make/go.mk
fbm3307: This needs to be deleted from all the other repos, so i will do it as a separate PR, as it will require change to script too ..

deploy/resources.go
fbm3307: good point done

make/go.mk
mfrancisc: I see you removed the dependency from test-with-coverage . Is this still needed ?

fbm3307: The github action test-with-coverage is run from the master branch, for running this, so for this PR i removed it from test-with-coverage, but just to make the CI happy, i kept the empty target, so that the CI doesn't fail and will remove this target once this PR merges

mfrancisc: Ok it's weird 😕 , even if the GH action is run from master, what matters is the make file target which doesn't rely anymore on `generate-assets`. But I'm sure I'm missing something. Also removing it in the next PR is fine 👍

fbm3307: its like a chicken and eg problem, since it runs from the master branch, it would expect the make target to be present. and the action would fail. so even in this PR i removed it from the action but kept an empty target so the ci doesn't complain. and just remove the target in next PR

metlos (approved):

make/go.mk
mfrancisc: yeah maybe I'm just a bit confused . The GH action runs from master but the code ( makefile targets in this case ) are from your PR. But maybe I'm making some confusion. 

Anyways feel free to remove it later if you saw it was failing.

mfrancisc (approved):
Looks good 👍 

Thanks for addressing my comments and the additional details 🙏

MatousJobanek (approved):

PR #665, opened 2025-04-23
SANDBOX-955: update kube & openshift dependencies to 4.18

rsoaresd:
## Description
Update dependencies 

Tool/Library | Current Version | Updates to Version
-- | -- | --
Operator SDK | 1.38 | 1.39.2
k8s.io/* | v0.30.1 | v0.31.2
controller-runtime | v0.18.4 | v0.19.7
controller-tools  | v0.15.0 | v0.16.5

## Issue ticket number and link
[SANDBOX-955](https://issues.redhat.com/browse/SANDBOX-955)


cmd/main.go
```diff
@@ -178,6 +180,9 @@ func main() {
 		// disable caching of Node metrics in the client to avoid getting the following error every second
 		// "failed to watch *v1beta1.NodeMetrics: the server does not allow this method on the requested resource (get nodes.metrics.k8s.io)"
 		Client: client.Options{Cache: &client.CacheOptions{DisableFor: []client.Object{&kmetrics.NodeMetrics{}}}},
+		Controller: ctrlconfig.Controller{
```
rsoaresd, line 183: Due to [this change](https://github.com/kubernetes-sigs/controller-runtime/pull/2918) in `sigs.k8s.io/controller-runtime` (introduced in v0.19.0), we need to skip the name validation to avoid this error:

`{"level":"error","ts":"2025-04-22T10:46:41.577Z","logger":"setup","msg":"unable to create controller","controller":"ToolchainCluster","error":"controller with name toolchaincluster already exists. Controller names must be unique to avoid multiple controllers reporting to the same metric","stacktrace":"main.main\n\t/Users/rsoaresd/member-operator/cmd/main.go:245\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:271"}`

make/run-cicd-script.mk
```diff
@@ -1,4 +1,4 @@
-OWNER_AND_BRANCH_LOCATION=codeready-toolchain/toolchain-cicd/master
+OWNER_AND_BRANCH_LOCATION=rsoaresd/toolchain-cicd/openshift_4_18_cicd
```
rsoaresd, line 1: temporary, i will remove before merge

MatousJobanek: correct me if I'm wrong, you didn't modify anything in the scripts, but only in the composite GH Actions. So there shouldn't be any need for this explicit verification.
What we need to verify (redirect to your branch) is the GH Actions step - see my comment in toolchain-cicd PR

cmd/main.go
MatousJobanek: out of curiosity, is there a way to configure the controller name?

MatousJobanek: I found it - there is `Named` function in the controller builder. We could do something like this instead
```go
func (r *Reconciler) SetupWithManager(mgr ctrl.Manager) error {
	return ctrl.NewControllerManagedBy(mgr).
		Named("ToolchainClusterCache").
          ...
```
and the same for other duplicated controllers too - that should be pretty simple change, right?

make/run-cicd-script.mk
rsoaresd: My bad!! You are right! Sorry! Going to address that

.github/workflows/publish-operators-for-e2e-tests.yml
```diff
@@ -63,7 +63,7 @@ jobs:
         python-version: '3.x'
 
     - name: Prepare tools
-      uses: codeready-toolchain/toolchain-cicd/prepare-tools-action@master
+      uses: rsoaresd/toolchain-cicd/prepare-tools-action@openshift_4_18_cicd
```
rsoaresd, line 66: temporary, i will remove before merge

cmd/main.go
rsoaresd: Going to test!

MatousJobanek: But you should update the `toolchaincluster_cache_controller.go` in toolchain-common.  The idler controller doesn't watch `ToolchainCluster` CRs.

rsoaresd: Thank you so much, addressed it! https://github.com/codeready-toolchain/toolchain-common/pull/478/commits/9d0638b33fb1de2c8006c22670232d3e0a60e4d9

alexeykazakov (approved):

MatousJobanek (approved):

PR #666, opened 2025-04-28
KUBESAW-143: Add Support for Go Templates

fbm3307:
This PR is the 1st PR in the series. 
This just contains the logic on fetching the TTR and fallback to the current logic if not found.

other  steps will follow this PR


MatousJobanek (commented):
Thanks for the changes, could you please update the unit tests accordingly?

fbm3307:
> Thanks for the changes, could you please update the unit tests accordingly?

updated it

controllers/nstemplateset/tier_template_revision.go
```diff
@@ -0,0 +1,34 @@
...
+	"fmt"
+
+	toolchainv1alpha1 "github.com/codeready-toolchain/api/api/v1alpha1"
+	"github.com/codeready-toolchain/toolchain-common/pkg/cluster"
+
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// getTierTemplateRevision gets the TierTemplateRevision resource from the host cluster.
```
metlos, line 14: Nitpick: the name doesn't match the real function name (and I personally like the name without the "Toolchain" better) :)

controllers/nstemplateset/nstemplatetier.go
```diff
@@ -11,37 +11,48 @@ import (
...
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			return nil, err
+		}
 	}
 	tierTmpl := &tierTemplate{
 		templateRef: templateRef,
 		tierName:    tmpl.Spec.TierName,
```
metlos, line 49: This will fail if the `ttr` is found, because `tmpl` will be `nil` in that case.

controllers/nstemplateset/tier_template_revision_test.go
```diff
@@ -0,0 +1,79 @@
...
+			Namespace: test.HostOperatorNs,
+		},
+	}
+}
+
+func TestGetTTR(t *testing.T) {
+	// given
+	ttRev := createTierTemplateRevision("test-ttr")
+	ttRevTest := createTierTemplateRevision("test-ttr-abc")
+	ttRevExtra := createTierTemplateRevision("test-ttr-def")
```
mfrancisc, line 31: minor - those do not seem to be used , unless I'm missing something.

controllers/nstemplateset/nstemplatetier.go
mfrancisc: +1

we'll also need to add a unit test verifying that `getTierTemplate` works when ttr is present.

controllers/nstemplateset/tier_template_revision_test.go
```diff
@@ -0,0 +1,79 @@
...
+			if _, ok := obj.(*toolchainv1alpha1.TierTemplateRevision); ok {
+				return fmt.Errorf("mock error")
+			}
+			return cl.Client.Get(ctx, key, obj, opts...)
+		}
+		//when
+		_, err := getToolchainTierTemplateRevision(ctx, hostCluster, "test-ttr")
+
+		//then
+		require.ErrorContains(t, err, "unable to retrieve the TierTemplateRevision")
```
mfrancisc, line 75: we should check that "mock error" is present, to be sure the error returned is the expected one.

mfrancisc (commented):
Looks Good 👍 

In addition to Lukas review, I have only few minor comments.

controllers/nstemplateset/tier_template_revision_test.go
fbm3307: I just kept them to check that out of all the TTR present it matches and gets the required TTR only

metlos (approved):

controllers/nstemplateset/nstemplatetier.go
```diff
@@ -11,37 +11,59 @@ import (
...
 
 var tierTemplatesCache = newTierTemplateCache()
 
-// getTierTemplate retrieves the TierTemplate resource with the given name from the host cluster
+// getTierTemplate retrieves the TierTemplateRevision resource with the given name from the host cluster,
+// if not found then falls back to the current logic of retrieving the TierTemplate
 // and returns an instance of the tierTemplate type for it whose template content can be parsable.
 // The returned tierTemplate contains all data from TierTemplate including its name.
 func getTierTemplate(ctx context.Context, hostClusterFunc cluster.GetHostClusterFunc, templateRef string) (*tierTemplate, error) {
+	var tmpl *toolchainv1alpha1.TierTemplate
```
MatousJobanek, line 28: only a detail, but this var doesn't have to be defined at this level

MatousJobanek (approved):
:+1:

controllers/nstemplateset/tier_template_revision_test.go
mfrancisc: I see, thanks for explaining.

controllers/nstemplateset/nstemplatetier.go
fbm3307: removed it now , Thanks :)

PR #669, opened 2025-05-07
collect all owners for the pod to be idled

MatousJobanek:
This is the first refactoring step in idler controller. The final goal is to merge the aap idler with the main idler logic & make the controller logic easily extendable so we can add new kinds/types via simple code change in the future.

The idea is:
1. collect the whole tree of all owners for the given pod (that is to be idled)
2. going from the top through the owners find the first known (supported) type - aap, vm, deployment, etc...
3. idle the owner resource based on the given logic

This PR is for the first step - collecting all owners for the given object. 
To make the PR changes smaller, I only extracted the owner-related logic from aap controller. In the next PR I'll modify the main controller logic so it uses the list of owners.

Note: There can be some code-duplication in the intermediate steps, but I'll address them in the following PRs.

[SANDBOX-1101](https://issues.redhat.com/browse/SANDBOX-1101)


xcoulon (approved):

rsoaresd (approved):
Great work 🚀

controllers/idler/owners.go
```diff
@@ -0,0 +1,106 @@
...
+		dynamicClient:   dynamicClient,
+	}
+}
+
+// getOwners returns the whole tree of all controller owners going recursively to the top owner for the given object
+func (o *ownerFetcher) getOwners(ctx context.Context, obj metav1.Object) ([]client.Object, error) {
+	if o.resourceLists == nil {
+		// Get all API resources from the cluster using the discovery client. We need it for constructing GVRs for unstructured objects.
+		// Do it here once, so we do not have to list it multiple times before listing/getting every unstructured resource.
+		resourceLists, err := o.discoveryClient.ServerPreferredResources()
```
alexeykazakov, line 32: So you will discover APIs for every owner fetcher (== every pod in every user namespace per idler run) instead of the current (once per idler run)? I guess it's not a big deal but was curious if you planning to optimize it in the future PRs or leave it as is.

alexeykazakov (approved):

controllers/idler/owners.go
MatousJobanek: The resourceList is retrieved only if it's not already present (and stored) in the  "owners fetcher" object (there is also a unit-test for it). My plan is to create an instance of the owner fetcher per idler run (once in a reconcile), so there will be only one API discovery per idler run - in the same way as it is done now.

PR #670, opened 2025-05-12
use ownerFetcher in main idler logic

MatousJobanek:
follow-up of https://github.com/codeready-toolchain/member-operator/pull/669
It introduces the ownerFetcher in the main idler logic.
For every pod that is about to be idled, it fetches the whole tree of owners and going from the top it picks the first "known" owner and scales it down.

* This PR removes a lot of code because the list of owners already contain the whole object, so there is no need to fetch it again.
* It also changes the VM case since the root top owner is VM (and not VMI)

The final code contains some duplication - mainly in those cases when it simply patch the resource with
`{"spec":{"replicas":0}}`
This duplication will be removed in the upcoming PR(s)

[SANDBOX-1101](https://issues.redhat.com/browse/SANDBOX-1101)


controllers/idler/idler_controller.go
```diff
@@ -460,196 +464,90 @@ func getSupportedScaleResource(ownerReference metav1.OwnerReference) *schema.Gro
...
-		return "", "", false, err
+func (r *Reconciler) scaleDeploymentConfigToZero(ctx context.Context, object client.Object) error {
+	patch := []byte(`{"spec":{"replicas":0,"paused":false}}`)
+	if err := r.AllNamespacesClient.Patch(ctx, object, client.RawPatch(types.MergePatchType, patch)); err != nil {
+		return err
 	}
-	log.FromContext(ctx).Info("DeploymentConfig scaled to zero", "name", dc.Name)
-	return owner.Kind, owner.Name, true, nil
+	log.FromContext(ctx).Info("DeploymentConfig scaled to zero", "name", object.GetName())
+	return nil
```
alexeykazakov, line 506: I guess we can replace most of those `scaleXtoZero()` methods by a generic `scaleControllerReplicasToZero()`  or something like this, right? The only difference between them is just logging if I'm not missing anything.

controllers/idler/idler_controller.go
```diff
@@ -460,196 +464,90 @@ func getSupportedScaleResource(ownerReference metav1.OwnerReference) *schema.Gro
...
-		if apierrors.IsNotFound(err) { // Ignore not found errors. Can happen if the parent controller has been deleted. The Garbage Collector should delete the pods shortly.
-			return owner.Kind, owner.Name, true, nil
-		}
-		return "", "", false, err
+		return err
 	}
-	logger.Info("Job deleted", "name", j.Name)
-	return owner.Kind, owner.Name, true, nil
+	logger.Info("Job deleted", "name", object.GetName())
+	return nil
```
alexeykazakov, line 531: And same for `deleteX()`, could be something like `deleteObject()`

alexeykazakov (approved):
Looks good. However I think we could make some of the methods generic.

controllers/idler/idler_controller.go
MatousJobanek: Yes, this is exactly the duplication I was talking about. Let me do the change in smaller PRs. For this case, in order to minimize the potential damage, I wanted to "only" introduce the owner fetcher, but still keep the logic similar to the previous format.

controllers/idler/idler_controller.go
```diff
@@ -381,71 +384,73 @@ func (r *Reconciler) getUserEmailsFromMURs(ctx context.Context, hostCluster *clu
...
+func (r *Reconciler) scaleControllerToZero(ctx context.Context, meta metav1.Object, ownerFetcher *ownerFetcher) (kind string, name string, err error) {
+	log.FromContext(ctx).Info("Scaling controller to zero", "name", meta.GetName())
+	var owners []*objectWithGVR
+	owners, err = ownerFetcher.getOwners(ctx, meta)
+	if err != nil {
+		return
+	}
+	for _, ownerWithGVR := range owners {
+		owner := ownerWithGVR.object
+		kind = owner.GetObjectKind().GroupVersionKind().Kind
```
metlos, line 399: Does the dynamic client fill in the `TypeMeta` of an object? Because generally, one cannot assume it is filled in in a `client.Object` instance. E.g. the dance in https://github.com/codeready-toolchain/toolchain-common/blob/master/pkg/client/ssa_client.go#L223

controllers/idler/owners.go
```diff
@@ -24,8 +24,13 @@ func newOwnerFetcher(discoveryClient discovery.ServerResourcesInterface, dynamic
 	}
 }
 
+type objectWithGVR struct {
+	object client.Object
+	gvr    *schema.GroupVersionResource
```
metlos, line 29: This doesn't seem to be used.

controllers/idler/idler_controller.go
MatousJobanek: it's a resource retrieved from a cluster (the type is unstructured). I expect that any resource that is retrieved from the cluster/etcd with any type of client should contain the TypeMeta.
In your case (the line that you referenced) the code can read objects that were created in the code (using the specific type of the resource) without setting the TypeMeta explicitly

controllers/idler/owners.go
MatousJobanek: good catch, this is gonna be used in the upcoming PR :-) stay tuned :wink:

controllers/idler/idler_controller.go
metlos: > apart from that, the code will work with resources whose API is not in Scheme (eg. AAP or VMs)

That's a very good point.

Should we think about modifying our apply clients to work explicitly with `Unstructured` and dynamic client, too, so that we're not bound to a small subset of resources that we "know about"?

MatousJobanek: I think that the SSA code is written in a way it can support both:
* `Unstructured` (or some generic `runtime.Object`) when it's retrieved from a template/cluster - in this case it contains the `TypeMeta`. In this case, the API doesn't have to be in the Schema
* typed objects that don't have the `TypeMeta` but it's API is in the Schema, so the logic can automatically fill it in

this is exactly what we want

PR #673, opened 2025-05-16
use the SSA apply client for installing the autoscaler

metlos:
`$TITLE`

### DIFF
--- a/pkg/autoscaler/autoscaler.go
+++ b/pkg/autoscaler/autoscaler.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/codeready-toolchain/member-operator/deploy"
+	"github.com/codeready-toolchain/member-operator/pkg/constants"
 	applycl "github.com/codeready-toolchain/toolchain-common/pkg/client"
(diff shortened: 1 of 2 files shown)


pkg/autoscaler/autoscaler.go
```diff
@@ -31,18 +32,13 @@ func Deploy(ctx context.Context, cl client.Client, s *runtime.Scheme, namespace
...
-	applyClient := applycl.NewApplyClient(cl)
+	applyClient := applycl.NewSSAApplyClient(cl, constants.MemberOperatorFieldManager)
 	// create all objects that are within the template, and update only when the object has changed.
 	for _, obj := range objs {
-		applied, err := applyClient.ApplyObject(ctx, obj)
-		if err != nil {
+		if err := applyClient.ApplyObject(ctx, obj); err != nil {
 			return errs.Wrap(err, "cannot deploy autoscaling buffer template")
 		}
-		if applied {
```
fbm3307, line 41: Just a thought , should we keep the check of it was applied or not?

fbm3307 (approved):
looks good

PR #677, opened 2025-05-27
drop tracked pods from idler CR

MatousJobanek:
As part of the refactoring process of the idler controller, I decided to drop the list of pods from the Idler CRDs to significantly simplify the flow of the controller. The tracked pods in the CRs:
* require extra reconciles
* add additional complexity to the tests to instruct the controller to start tracking the pods
* with more generic approach in the controller, it's a huge challenge to calculate which pods should be still tracked and which not in the tests

For debugging potential problem in the idler, I rather use the logs (I haven't ever used the list of the tracked pods in the CRs). Apart from that, all needed information to idle pods is already present in the in the cluster (pod resources)

[KUBESAW-269](https://issues.redhat.com/browse/KUBESAW-269)

related PR: https://github.com/codeready-toolchain/api/pull/478

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/codeready-toolchain/member-operator
 
 require (
-	github.com/codeready-toolchain/api v0.0.0-20250506092100-39b4862e1271
+	github.com/codeready-toolchain/api v0.0.0-20250603081710-8a56a236e3a4
 	github.com/codeready-toolchain/toolchain-common v0.0.0-20250506093954-2b65ad3a2e12
 	github.com/go-logr/logr v1.4.2
 	github.com/google/go-cmp v0.6.0
@@ -19,8 +19,6 @@ require (
 	k8s.io/metrics v0.31.2
 	sigs.k8s.io/controller-runtime v0.19.7
 )
-
-replace github.com/codeready-toolchain/api => github.com/matousjobanek/api v0.0.0-20250527073940-9271ed30af92
 
 require (
 	github.com/google/uuid v1.6.0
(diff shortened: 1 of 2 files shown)


controllers/idler/idler_controller.go
```diff
@@ -297,7 +285,7 @@ func (r *Reconciler) notify(ctx context.Context, idler *toolchainv1alpha1.Idler,
 		logger.Error(err, "failed to create Notification")
 		if err = r.setStatusIdlerNotificationCreationFailed(ctx, idler, err.Error()); err != nil {
 			logger.Error(err, "failed to set status IdlerNotificationCreationFailed")
-		} // not returning error to continue tracking remaining pods
+		} // not returning error to continue idling remaining pods
```
alexeykazakov, line 288:
```suggestion
		} // not returning error to continue processing remaining pods
```

alexeykazakov (approved):
:+1:

controllers/idler/idler_controller.go
```diff
@@ -297,7 +285,7 @@ func (r *Reconciler) notify(ctx context.Context, idler *toolchainv1alpha1.Idler,
 		logger.Error(err, "failed to create Notification")
 		if err = r.setStatusIdlerNotificationCreationFailed(ctx, idler, err.Error()); err != nil {
 			logger.Error(err, "failed to set status IdlerNotificationCreationFailed")
-		} // not returning error to continue tracking remaining pods
+		} // not returning error to continue processing remaining pods
```
fbm3307, line 288: would it make sense to return error/list of error at the end?

fbm3307 (approved):

controllers/idler/idler_controller.go
MatousJobanek: in this particular case, not really. Reconciling again wouldn't help with sending the notification again

PR #684, opened 2025-06-11
treat Camel CRs as standalone scalable resources

MatousJobanek:
till now, we've treated the Camel resources as a way to idle deployment. In such cases, we informed users about the deployment that was scaled down, not about the Camel resource. If the user wanted to scale the deployment again up, it failed because the deployment was scaled immediately down by the Camel resource that had replica set to `0`. 
This PR modifies the logic so it treats the Camel CRs as standalone "top-level" scalable resources. The user will get notified about the Camel resource that was scaled down, not the deployment.

As always, to make the things happen I had to refactor/modify a few more things:
* refactored the unit test that verifies the content of the notifications so it focuses purely on the notification content only
* introduced FakeClientSet that encapsulates all the fake clients used in unit tests (newly also the FakeScaleClient) to minimize the number of parameters passed/returned in functions
    * I know that there is already one `clientSet` - this is gonna be removed in the next PR where I'll merge the aap idler into the main one
* moved common logic of creating deployment to a separate function

### DIFF
--- a/controllers/idler/owners_test.go
+++ b/controllers/idler/owners_test.go
@@ -86,8 +86,8 @@ var testConfigs = map[string]createTestConfigFunc{
 	},
 	"KameletBinding": func(plds payloads) payloadTestConfig {
 		return payloadTestConfig{
-			// We are testing the case with a nested controllers (KameletBinding -> Deployment -> ReplicaSet -> Pod) here,
-			// so we the pod's owner is ReplicaSet but the expected scaled app is the top-parent KameletBinding CR.
+			// We are testing the case with nested controllers (KameletBinding -> Deployment -> ReplicaSet -> Pod) here,
+			// so the pod's owner is ReplicaSet but the expected scaled app is the top-parent KameletBinding CR.
 			podOwnerName:    fmt.Sprintf("%s-deployment-replicaset", plds.kameletBinding.GetName()),
 			expectedAppName: plds.kameletBinding.GetName(),
 			ownerScaledUp: func(assertion *test.IdleablePayloadAssertion) {


controllers/idler/owners_test.go
```diff
@@ -70,6 +70,34 @@ var testConfigs = map[string]createTestConfigFunc{
 			},
 		}
 	},
+	"Integration": func(plds payloads) payloadTestConfig {
+		return payloadTestConfig{
+			// We are testing the case with a nested controllers (Integration -> Deployment -> ReplicaSet -> Pod) here,
+			// so we the pod's owner is ReplicaSet but the expected scaled app is the top-parent Integration CR.
```
alexeykazakov, line 76:
```suggestion
			// We are testing the case with nested controllers (Integration -> Deployment -> ReplicaSet -> Pod) here,
			// so the pod's owner is ReplicaSet but the expected scaled app is the top-parent Integration CR.
```

controllers/idler/owners_test.go
```diff
@@ -70,6 +70,34 @@ var testConfigs = map[string]createTestConfigFunc{
...
+			},
+			ownerScaledDown: func(assertion *test.IdleablePayloadAssertion) {
+				assertion.ScaleSubresourceScaledDown(plds.integration)
+			},
+		}
+	},
+	"KameletBinding": func(plds payloads) payloadTestConfig {
+		return payloadTestConfig{
+			// We are testing the case with a nested controllers (KameletBinding -> Deployment -> ReplicaSet -> Pod) here,
+			// so we the pod's owner is ReplicaSet but the expected scaled app is the top-parent KameletBinding CR.
```
alexeykazakov, line 90:
```suggestion
			// We are testing the case with nested controllers (KameletBinding -> Deployment -> ReplicaSet -> Pod) here,
			// so the pod's owner is ReplicaSet but the expected scaled app is the top-parent KameletBinding CR.
```

alexeykazakov (approved):
Nice!

PR #688, opened 2025-07-03
Idle Notebooks in controller

MatousJobanek:
Moves the CronJob part of idling Notebooks into the idler controller https://github.com/codeready-toolchain/sandbox-sre/blob/b4705f3b43b16d3cdb9f0614a2ff024ce6635b3f/components/rhoai/base/nb-culler.yaml#L72-L82

paired with https://github.com/codeready-toolchain/toolchain-e2e/pull/1170

Assisted-by: Cursor

### DIFF
--- a/controllers/idler/idler_controller_test.go
+++ b/controllers/idler/idler_controller_test.go
@@ -228,7 +228,10 @@ func TestEnsureIdling(t *testing.T) {
 				VMRunning(noise.vmStopCallCounter).
 				AAPIdled(podsRunningForTooLong.aap).
 				AAPRunning(podsTooEarlyToKill.aap).
-				AAPRunning(noise.aap)
+				AAPRunning(noise.aap).
+				NotebookStopped(podsRunningForTooLong.notebook).
+				NotebookRunning(podsTooEarlyToKill.notebook).
+				NotebookRunning(noise.notebook)
 
 			memberoperatortest.AssertThatIdler(t, idler.Name, fakeClients).
 				HasConditions(memberoperatortest.Running(), memberoperatortest.IdlerNotificationCreated())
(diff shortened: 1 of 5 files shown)


xcoulon (approved):

MatousJobanek:
> Nice! It's cool to see how easy it is now to extend the idler now (besides writing tests :)

Thanks. I was thinking about bringing some higher abstraction that could technically also help with the tests, but it would require a lot of additional code and I don't think that it's worth it. I think that it's fine to keep it as it is - we can use this PR as an example for the next types of workloads to know what to change.

> Are you planning to remove the notebook stop functionality from nb-culler.yaml next? 

yes, definitely

> We will still need to keep the cleanup of the resources left after the user was deleted there though :(

are you talking about this part? https://github.com/codeready-toolchain/sandbox-sre/blob/5c6f8df9bcc010eafd830f063f2423cf41f58753/components/rhoai/base/nb-culler.yaml#L40-L59
TBH, I think that we can delete it too - the notebooks should be created only in users namespaces, but it's worth checking it with AI folks.
see
https://github.com/codeready-toolchain/sandbox-sre/blob/5c6f8df9bcc010eafd830f063f2423cf41f58753/components/rhoai/base/odhdashboardconfig.yaml#L32-L35
https://redhat-internal.slack.com/archives/CHK0J6HT6/p1743433202765729
if it's not true, then we would need to keep also the idler CronJob since our idler controller supports only users namespaces

alexeykazakov:
> if it's not true, then we would need to keep also the idler CronJob since our idler controller supports only users namespaces

Right! I forgot about that.

controllers/idler/idler_controller_test.go
```diff
@@ -228,7 +228,10 @@ func TestEnsureIdling(t *testing.T) {
 				VMRunning(noise.vmStopCallCounter).
 				AAPIdled(podsRunningForTooLong.aap).
 				AAPRunning(podsTooEarlyToKill.aap).
-				AAPRunning(noise.aap)
+				AAPRunning(noise.aap).
+				NotebookStopped(podsRunningForTooLong.notebook).
```
fbm3307, line 232: Just out of curiosity, there will not be any idle pods?

fbm3307 (approved):

MatousJobanek:
Closing in favor of the internal idler in RHOAI operator: https://redhat-internal.slack.com/archives/CHK0J6HT6/p1751986950016269?thread_ts=1751877903.895249&cid=CHK0J6HT6

PR #742, opened 2026-04-15
idler: catch the case when other owners are being deleted

MatousJobanek:
This is to cover the case when idling of the top-level owner didn't have any effect and all other owners are already being deleted. When this persists for more than 110% of the timeout, then it will try to delete the pod.
https://redhat-internal.slack.com/archives/C0AEFNZ5KHB/p1776136788506989

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved idle pod deletion fallback logic to better handle scenarios where pod owners are already being deleted and pods exceed timeout thresholds.

* **Tests**
  * Added test coverage for pod deletion when owners are in deletion state and pod exceeds idle timeout duration.

### DIFF
--- a/controllers/idler/owners.go
+++ b/controllers/idler/owners.go
@@ -54,11 +54,11 @@ func (i *ownerIdler) scaleOwnerToZero(ctx context.Context, pod *corev1.Pod) (str
 
 	timeoutSeconds := getTimeout(i.idler, *pod)
 	var topOwnerKind, topOwnerName string
-	attemptSuccessful := false
+	attempted := false
 	var errToReturn error
 	for _, ownerWithGVR := range owners {
 		if util.IsBeingDeleted(ownerWithGVR.Object) {
-			attemptSuccessful = false
+			attempted = false
 			continue
 		}
 		owner := ownerWithGVR.Object
@@ -81,7 +81,7 @@ func (i *ownerIdler) scaleOwnerToZero(ctx context.Context, pod *corev1.Pod) (str
 			continue // Skip unknown owner types
 		}
 
-		attemptSuccessful = true
+		attempted = true
 		// Store the first processed owner's info and preserve its error
 		if topOwnerKind == "" {
 			topOwnerKind = ownerKind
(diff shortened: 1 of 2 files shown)


xcoulon (approved):

controllers/idler/owners.go
```diff
@@ -78,6 +81,7 @@ func (i *ownerIdler) scaleOwnerToZero(ctx context.Context, pod *corev1.Pod) (str
 			continue // Skip unknown owner types
 		}
 
+		attemptSuccessful = true
```
rajivnathan: `attemptSuccessful` will be true even if there was an error right? That sounds a bit confusing. Maybe just  `attempted` would be better?

rajivnathan (approved):
👍

controllers/idler/owners.go
MatousJobanek: In this particular case, the name is always a bit confusing, because it always tries to idle the top-level owner. Naming is hard :-)

MatousJobanek: renamed 253e1891

controllers/idler/owners.go
```diff
@@ -52,10 +52,13 @@ func (i *ownerIdler) scaleOwnerToZero(ctx context.Context, pod *corev1.Pod) (str
 		logger.Error(err, "failed to find all owners, try to idle the workload with information that is available")
 	}
 
+	timeoutSeconds := getTimeout(i.idler, *pod)
 	var topOwnerKind, topOwnerName string
+	attempted := false
 	var errToReturn error
 	for _, ownerWithGVR := range owners {
 		if util.IsBeingDeleted(ownerWithGVR.Object) {
+			attempted = false
```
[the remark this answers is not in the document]
MatousJobanek: let's keep it as it is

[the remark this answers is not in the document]
metlos: I'll let @MatousJobanek confirm, but I believe the behaviour is actually correct. The important factor is the time test. If the pod is still running after 110% of the timeout, we delete it directly even if all its parents are marked for deletion. This prefers the idling over waiting for eventual consistency.

metlos (approved):

PR #761, opened 2026-07-31
Use toolchain-common/pkg/idler for Idler owner actions (core idling functionality moves to common)

alexeykazakov:
- Refactor the member Idler to call `toolchain-common/pkg/idler` for owner idle actions instead of owning that logic locally.
- Keep timeout-driven orchestration in member-operator (105%/110% second-owner policy, pod delete fallback, notifications, predicates/status).

Depends on: https://github.com/codeready-toolchain/toolchain-common/pull/540

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of resource ownership during idling operations.
  * Unsupported resource types are now skipped gracefully instead of interrupting processing.
  * Timeout settings and secondary-owner policies are applied consistently during idling.

* **Refactor**
  * Consolidated idling and ownership behavior for more consistent results across supported resources.
  * Improved owner-chain processing and resource scaling checks for more reliable idling decisions.
  * Simplified maintenance by centralizing shared idling behavior.


controllers/idler/owners.go
```diff
@@ -69,22 +58,8 @@ func (i *ownerIdler) scaleOwnerToZero(ctx context.Context, pod *corev1.Pod) (str
...
-		case "VirtualMachine":
-			err = i.stopVirtualMachine(ctx, ownerWithGVR) // Nothing to scale down. Stop instead.
-		case "AnsibleAutomationPlatform":
-			err = i.idleAAP(ctx, ownerWithGVR) // Nothing to scale down. Stop instead.
-		case "Claw":
-			err = i.idleClaw(ctx, ownerWithGVR)
-		case "ServingRuntime":
-			err = i.idleServingRuntime(ctx, ownerWithGVR) // Idle by deleting old InferenceService objects.
-		default:
+		err = i.actions.IdleOwner(ctx, ownerWithGVR, opts)
```
MatousJobanek: how about moving also this part that walks over the owners - isn't it what the mcp server would need to do as well anyway?

MatousJobanek (approved):
looks good, but I guess that you could move more code, or is there some significant difference that should be executed in the mcp server compared to what we have here?

controllers/idler/owners.go
MatousJobanek: GH doesn't show properly the part of the code that I selected, I meant most of (if not the whole) `scaleOwnerToZero` function. This means including fetching the owners and walking through them. I know that you don't need the additional thresholds, but that can be only a configuration option.

alexeykazakov: `scaleOwnerToZero()` does more than we need in the MCP tool for now. It also deletes the kinds it can't just scale down. And it has the timeouts in it which we don't need in the MCP tool (we just need to scale down everything no matter how long the worklaods run from the mcp tool). But let me see if we can make it configurable (we might want to start deleting unsalable kinds from MCP tools in the future too) so we can move more code to common.

alexeykazakov: OK. More code moved to common now with extra config options. Please take a look.

alexeykazakov:
/retest

(infra issue)

